PR #3, opened 2018-06-16
Rewrote ore generation

alcatrazEscapee:
Some data on cascading lag. I did ten trials for each generation type in which I loaded a new world and counted the number of chunks that were generated via cascading lag. This is the results:

OLD ORES: mean = 32.8, standard dev. = 12.8
NO ORES: mean = 8.1, standard dev. = 6.5
NEW ORES: mean = 8.5, standard dev = 4.6

I think it is fair to say that this eliminates cascading lag that is caused by ore generation.

Other changes:
 - Ore generation is now controlled via a json file
 - Default file is at assets/tfc/config, which will be copied into .minecraft/config/tfc at first load
 - Can support non-tfc blocks, by putting a block name in the ore field (i.e. "ore": "minecraft:gold_ore" instead of "ore": "native_gold")
 - Size, type, etc. are all controlled via enums. Can add and change as necessary.

### DIFF
--- a/src/main/java/net/dries007/tfc/TerraFirmaCraft.java
+++ b/src/main/java/net/dries007/tfc/TerraFirmaCraft.java
@@ -5,6 +5,8 @@
 
 package net.dries007.tfc;
 
+import net.dries007.tfc.objects.Ore;
+import net.dries007.tfc.util.OreSpawnData;
 import org.apache.logging.log4j.Logger;
 import net.minecraftforge.fluids.FluidRegistry;
 import net.minecraftforge.fml.common.FMLCommonHandler;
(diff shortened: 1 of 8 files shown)


dries007:
Nice work, a few points:

- Please use our code styles, if you need help with getting that working ping me on Discord.
- Why `base-rocks`? everything else is done by underscore, this should be `base_rocks` to fit in.
- Related to that: I immensely prefer using GSON over typesafe's config lib, any reason why you are using that?

alcatrazEscapee:
Should be all fixed now. Let me know if there is anything else. I also fixed a few issues with the ore generation behaving weirdly.

PR #7, opened 2018-07-25
Calendar climate

Bunsan:
Fixes for Calendar. This is still WIP, but provides hooks to use for getting dates. This also adjusts climate calculation so that it is Hot south and cold north.


Bunsan:
IndexOutOfBounds exception in this. Alcatraz is doing fancy stuff with temperatures now.

PR #8, opened 2018-07-28
Trees

alcatrazEscapee:
Adds 90% of everything to do with trees:
 - pre-block registry Tree object has replaced wood enum.
 - Tree generation, using a bunch of different structure blocks / tree generation methods
 - TFC leaf decay, and tree chopping
 - world gen for trees. Still WIP, waiting for @Bunsan calendar PR to properly get temperature based spawning working. Also generation is pretty basic right now. More plant life + jungles and such need to come later.
 - Lots of fixes for random wood / tree related things

Still todo:
 - improve world gen.
 - properly implement jungles / maybe rethink the whole tree layer system to allow for more customized areas.


alcatrazEscapee:
Added more to worldgen:
* New per-chunk temperature / rainfall / Flora density / Flora diversity layer
* Removed EVT / Rain / Tree GenLayers
* new layers are per-chunk. Make note when using them; no need to sample per block
* new layers are also continuous.  Also make note: use random variation to achieve smooth transition zones

Tree spawning / creation / growing is all working.

PR #12, opened 2018-09-07
Tweaks and Tool Racks!

LightningSh0ck:
Can I be a contributor on Discord now?


src/main/java/net/dries007/tfc/client/ClientRegisterEvents.java
```diff
@@ -188,6 +195,9 @@ public ModelResourceLocation getModelLocation(@Nonnull ItemStack stack)
         ModelLoader.setCustomStateMapper(BlocksTFC.INGOT_PILE, blockIn -> ImmutableMap.of(BlocksTFC.INGOT_PILE.getDefaultState(), new ModelResourceLocation("tfc:empty")));
         ClientRegistry.bindTileEntitySpecialRenderer(TEIngotPile.class, new TESRIngotPile());
 
+        ModelLoader.setCustomStateMapper(BlocksTFC.INGOT_PILE, blockIn -> ImmutableMap.of(BlocksTFC.INGOT_PILE.getDefaultState(), new ModelResourceLocation("tfc:empty")));
```
alcatrazEscapee: Why have these been duplicated?

src/main/java/net/dries007/tfc/objects/te/TEWorldItem.java
```diff
@@ -82,11 +82,12 @@ public SPacketUpdateTileEntity getUpdatePacket()
     @Override
     public NBTTagCompound getUpdateTag()
     {
-        // The tag from this method is used for the initial chunk packet, and it needs to have the TE position!
```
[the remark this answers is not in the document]
LightningSh0ck: I did it like so because I was afraid I might have broken something by doing that even though it souldn't and did not look like so. WorldItems left me with a scar.

src/main/java/net/dries007/tfc/client/ClientRegisterEvents.java
LightningSh0ck: Ohhh, snap. I was copy-pasting code for the tool racks and I left this one here by accident. To be honest I don't know what these methods do under the hood.

alcatrazEscapee: State Mappers map a block's `IBlockState` properties to a variant string ("type=north,color=blue"). The variant string is what is used in the json blockstate file. If you look at `IStateMapper`, it is just a wrapper around a map from a `IBlockState` to a `ModelResourceLocation`. Certain blocks have `IProperty<>`s that aren't used in model loading, so they have custom state maps which ignore these properties. (This simplifies blockstate jsons).

There is no point binding a custom statemap `new StateMap.Builder().build()`, because if not specified, default behavior is that every `IProperty` gets added to the variant string. It is a redundant default declaration.

alcatrazEscapee:
Yeah, that's fine. It needs some rework anyway.

PR #14, opened 2018-09-19
Added Bellows.

LightningSh0ck:
And fixed a mistake I previously made while IJ made other passive-aggressive code format changes...


src/main/java/net/dries007/tfc/objects/blocks/BlocksTFC.java
```diff
@@ -396,7 +396,8 @@ public static void registerBlocks(RegistryEvent.Register<Block> event)
         // todo: sluice
         // todo: quern
         // todo: loom
-        // todo: bellows
+        inventoryItemBlocks.put(register(r, "bellows", new BlockBellows(), CT_MISC), ItemBlockTFC.class);
+        ;
```
LightningSh0ck: I really don't get why IJ did place that semicolon...

alcatrazEscapee: Don't review your own PR... there is no point. Just fix what needs fixing.

LightningSh0ck: Do you really want me to do an entire commit just to erase that semicolon?

alcatrazEscapee:
This has been wrapped up into the torches / fire pit PR as per request.

PR #18, opened 2018-12-17
Delete /run/.gitignore

Cadiboo:
This also removes the /run/ folder from the repository

### DIFF
--- a/run/.gitignore
+++ b/run/.gitignore
@@ -1,2 +1,0 @@
-*
-!.gitignore


dries007:
This is intentional to stop people from committing the contents of the `run` folder. The run folder can't be removed because if it doesn't exist, you can't run the server or client.
Is there any good reason to remove it?

Cadiboo:
It doesn't belong in the repository and every other MC mod repository (and the default forge MDK) have it excluded in the main .gitignore

Cadiboo:
I'll open another PR to add the stuff from the default Forge MDK

PR #19, opened 2018-12-18
Added ignore config from the default Forge MDK

Cadiboo:


dries007:
**Again, why are you making these changes?**

The only actual changes are:

1. Add `.metadata` and `eclipse` (useful, but not really required unless you are using Eclipse, in which case I feel sorry for you.)
2. Moved lines around for no reason.
3. Made the folders less specific so if we need a `bin`, `out`, ... folder anywhere else than the root of the repo it will now not be committed.
4. Added some gitignore exceptions for files that don't need them.
5. Removed the run folder, without which you can't start the client/server, thus requiring an extra manual step after cloning a fresh repository.

The Forge MDK is a great starting point for a beginner, but it's not some holy grale or reference/standard to uphold at cost of convenience.

If you need these changes for any reason other than to make changes, feel free to let me know, but I don't see any propose at all.

PR #20, opened 2019-02-03
Create zh_CN.lang

233Isaac:
add Chinese translation 
:D


alcatrazEscapee:
File name should be all lowercase since we are on pack version 3

PR #33, opened 2019-03-07
Barrels

domi1819:
I added barrels that can hold liquids and items.

There are still a lot of things to do, but i wanted to merge what i have so far.
The barrels should be fairly stable and already have some basic functionality.


src/main/java/net/dries007/tfc/objects/container/ContainerBarrel.java
```diff
@@ -0,0 +1,44 @@
...
+
+        if (inventory != null)
+        {
+            this.addSlotToContainer(new SlotTEInput(inventory, SLOT_FLUID_CONTAINER_IN, 35, 20, tile));
+            this.addSlotToContainer(new SlotOutput(inventory, SLOT_FLUID_CONTAINER_OUT, 35, 54));
+            this.addSlotToContainer(new SlotTEInput(inventory, SLOT_ITEM, 89, 37, tile));
+        }
+    }
+
+    public IFluidHandler getFluidTank()
```
alcatrazEscapee: This should be null checked on usage - the capability can return null if used by another implementation

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -223,10 +227,86 @@ public static EnumFacing getASolidFacing(World worldIn, BlockPos pos, @Nullable
         return null;
     }
 
-    @Nonnull
+    @Nullable
```
alcatrazEscapee: Don't change this `@Nonnull` - if it is null there is no point having it here, it just is unnecessary complication. The nonnull is intentional to prevent injected fields from giving null / constant conditions warnings everywhere. See discussion in discord.
```suggestion
    @Nonnull
```

src/main/java/net/dries007/tfc/client/ClientRegisterEvents.java
```diff
@@ -145,6 +145,12 @@ public ModelResourceLocation getModelLocation(@Nonnull ItemStack stack)
         for (ItemBlock item : BlocksTFC.getAllInventoryItemBlocks())
             ModelLoader.setCustomModelResourceLocation(item, 0, new ModelResourceLocation(item.getRegistryName(), "inventory"));
 
+        for (ItemBlock item : BlocksTFC.getAllBarrelItemBlocks())
+        {
+            ModelLoader.setCustomModelResourceLocation(item, 0, new ModelResourceLocation(item.getRegistryName(), "sealed=false"));
```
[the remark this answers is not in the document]
alcatrazEscapee: Nevermind, I don't know what I'm talking about

src/main/java/net/dries007/tfc/objects/container/ContainerBarrel.java
domi1819: Added @Nullable to the getters and null checks where they are used

src/main/java/net/dries007/tfc/util/Helpers.java
domi1819: Reverted. I excluded that change from being committed now

PR #41, opened 2019-04-19
Implemented prospector pickaxe.

moller21:
Here's the propick i talked about in the discord. Keep in mind there might be issues, I'm not a confident programmer:P

Notes:
- Uses the BlockOreTFC class to identify blocks, and from what I can understand thats the only block the world generator can generate in the world.
- Might needs some balancing (durability/speed/messages)


src/main/java/net/dries007/tfc/objects/items/metal/ItemProPick.java
```diff
@@ -0,0 +1,264 @@
...
+import static net.dries007.tfc.objects.blocks.stone.BlockOreTFC.GRADE;
+
+public class ItemProPick extends ItemMetalTool
+{
+    private BlockPos activeBlockPos = null;
+    private int timesUsedOnBlock = 0;
+    private final static int cooldown = 40;
+    private Random random = new Random();
+    private float efficiency;
+    private Map<String, ProspectResult> results = new HashMap<>();
```
alcatrazEscapee: This should be a return variable - its only used once and then discarded.
```java
Collection<ProspectResult> scanSurroundingBlocks()
...
return results.values();
```

src/main/java/net/dries007/tfc/objects/items/metal/ItemProPick.java
```diff
@@ -0,0 +1,264 @@
...
+                    {
+                        //Found nothing, tell nothing
+                        player.sendStatusMessage(new TextComponentTranslation("tfc.propick.found_nothing"),true);
+                        setCooldown(player);
+                    }
+                    else
+                    {
+                        ProspectResult[] list = results.values().toArray(new ProspectResult[0]);
+                        // todo: made show grades/multiple results if skilled prospector
+                        int index = results.size()== 1 ? 0 : random.nextInt(results.size()-1);
```
alcatrazEscapee: Another why? This should display all results, not pick randomly.

Bunsan:
Initial comments. I’ll need to look at it in much more detail

First we will stick to the wording used in TFC-Classic. So “Found traces of”, “Found a small sample of”,  “Found a medium sample of”, “Found a large sample of”, “Found a very large sample of” and “Found nothing of interest.”

Secondly the prospectors pick should not use BlockOreTFC. It should look the blocks delineated in the ore_spawn_data.json.

moller21:
Thanks for the feedback!

I'll change the messages and search function so it's more like the classic version.

The reason the I added the 5 times thing was because i felt it was abit silly to get an instant result, but 5 is probably too much. Maybe 3-2 times, but it's up to you.

Also, I'm using BlockOreTFC because it looks like thats what the ore gen uses too, see https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/dfc767b935440f145452515b0c75dfa88ad49954/src/main/java/net/dries007/tfc/world/classic/worldgen/WorldGenOre.java#L157

I wanted to use the Ore class, but couldnt figure out how to get the block from that. Gonna need help on that.

alcatrazEscapee:
So after some looking around, it seems `OreSpawnData` isn't really built for reverse lookups on ore spawned yet - so for now if you could just add a `//todo` it's fine to leave the ore lookup as is.

Bunsan:
The basic guidance is to go for a 1:1 port. Changes are certainly happening, but they are discussed first. So if you’d like to change any propick mechanics please discuss/explain reasons why these are good changes. The reason isn’t that the Dev Team think we are right and everyone else is wrong, but maintains a consistency in gameplay design and vision.

src/main/java/net/dries007/tfc/objects/items/metal/ItemProPick.java
alcatrazEscapee: Nevermind, I'm forgetting how TFC works. But it should probably seed the random with the location. Try `random.setSeed(pos.toLong())`. That should allow us to have well defined false negatives that aren't bypassed by repeated clicking.

moller21:
Sorry for late answer, holidays :)

Like I said above, I'll change the searching, messages and all the other logic behind the scanning to be like the classic version. Don't know how my changes would affect it so it's probably best to keep it at the default machanics.

What I would like to change is that messages should be shown above the actionbar instead of chat and that you'd have to hold down the right mouse button for a second or so untill you get the result. I feel mods shouldn't output gameplay messages in chat as it clutters, and also doesn't give the player a free log. The holding down part adds some realism i guess, that you can run past a block a block and rightclick it but gotta stop. When i think about it this could maybe be reduced by quality of the metal.

alcatrazEscapee:
As mentioned in the commit, I've added a proper way to check if a block is an ore block:
```java
VeinRegistry.INSTANCE.getVeins().values(); // returns a Collection<VeinType>
vein#isOreState(IBlockState state); // checks if state is valid for a specific vein
```

moller21:
Sorry for the spam, as you can see I've no idea how this works. I also threw in a little fix there so falling blocks doesn't destroy blocks in any of the horizonatal directions if there's a block there and not below it. Wasn't intended to be included here, but please don't make me remove it, I've no idea how to 😃 

I did the changes to the propick as I mentioned earlier, made the search etc. work like vanilla TFC. But kept the 3 times per block. Also added your other suggestions.

PR #44, opened 2019-04-28
Plants

Konlii:
Last update to plants before temperature-based rewrite


alcatrazEscapee:
There is a bug with the plant temperature tooltip - it sometimes will flicker constantly between two contrasting temperatures. This should be thought out better.

Konlii:
Looks like server and client have _slighty_ different temp values which is causing the flickering right at the edge of temp range.  Added `worldIn.isRemote` check.

alcatrazEscapee:
So I did some digging, because temperature shouldn't be different from client to server, and I was seeing a difference of ~5 C, which is pretty high. Calendar time is synced, base temperature, month temperature are all fine. The issue is actually that on client, the world seed is not synced, which is used to calculate the daily temperature modifier. So I'm sacrificing different daily temperature per world (which should not affect game play in any significant manner to keep the temperatures consistent.

PR #67, opened 2019-05-07
Item holder block.

Claycorp:
Added Item holder block that works like the pit kiln 'display'.
Placed by keybinding. Packet sent from client to server is sanity checked and shouldn't allow exploits of arbitrary block placement.
Also includes misc fixes like, adding a TFC key category and fixing breaking particles on technical blocks to be nothing instead of missing texture.

See it in action!


alcatrazEscapee:
Closing as per discord discussion, this will be merged into a much larger feature PR

PR #91, opened 2019-06-20
[Ready for review] Now forge can only be lit if requirements are met.

CtrlAltDavid01:
I noticed that charcoal pile only drops 1 charcoal if the block under it gets destroyed, should I drop its item LAYERS times, so the proper amount is dropped?

(Its my first ever interaction with java and minecraft modding, hopefully I don't mess up anything)

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/blocks/BlockCharcoalPile.java
+++ b/src/main/java/net/dries007/tfc/objects/blocks/BlockCharcoalPile.java
@@ -178,11 +178,13 @@ public class BlockCharcoalPile extends Block implements ILightableBlock
             {
                 if (!world.isRemote)
                 {
-                    world.setBlockState(pos, BlocksTFC.CHARCOAL_FORGE.getDefaultState().withProperty(LIT, true));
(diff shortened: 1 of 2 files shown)


CtrlAltDavid01:
I didn't see the issue you created I'll update accordingly

CtrlAltDavid01:
And made charcoal piles drop the correct amount when the lock is broken below it.

Bunsan:
The charcoal pile dropping all charcoal pieces when block below is broken makes it so player can just dig out bottom block to harvest the pile no matter how high it is. This is not behaviour we want. I don’t disagree only dropping one piece is mean, but this isn’t a good solution.

CtrlAltDavid01:
I see the problem, should I undo it or something else?

alcatrazEscapee:
@CtrlAltDavid01, look at the `removedByPlayer` and `onNeighourChange` methods in `BlockCharcoalPile` - it doesn't break normally. If you try and run this you'll end up having huge duplication bugs when you try and break charcoal

CtrlAltDavid01:
I have tried every way I could think of and there is no dupe.

PR #93, opened 2019-06-21
Player Tweaks

DisasterMoo:
Health, thirst and hunger bar like TFC-Classic.
Healthbar drawing can go beyond max health(for mods that adds that). Tested with /effect @p 21 200 20 (double vanilla health for 200 secs)
Thirstbar depletes faster while running/jumping. Apply bad effects when low, hurt when 0.
Normal living depletes thirstbar at one full bar per minecraft day(in Config 1.0F = one day)
No hunger changes, just changed it's HUD to be TFC style.
Health works under damage reescalation. PlayerBaseHealth == Vanilla(20) HP. Max and Min health(in config) determines how damage reescalate to hurt more/less. If player is at PlayerBaseHealth no damage reescalation is done.

A function implementation missing(for now) is damage reducing by armor defence, need some discussion on how to do that.


src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -147,6 +194,33 @@ public static void onUseHoe(UseHoeEvent event)
...
+    public static void onLivingHurt(LivingHurtEvent event)
+    {
+        //Starvation should not be scaled, you receive damage based on % anyway.
+        if (event.getEntityLiving() instanceof EntityPlayer && !(event.getSource() == DamageSource.STARVE))
+        {
+            EntityPlayer player = (EntityPlayer) event.getEntityLiving();
+            IPlayerData cap = event.getEntityLiving().getCapability(CapabilityPlayer.CAPABILITY_PLAYER_DATA, null);
+            if (cap != null)
+            {
+                float damage = DamageManager.rescaleDamage(event.getAmount(), 20, ConfigTFC.GENERAL.playerBaseHealth);
```
alcatrazEscapee: As I said in discord - there is **no reason** to scale damage up to 1000 here. In the same vein, having a config value for this is confusing and counterproductive - It should only be used for display purposes.

This still needs to rescale damage, yes, but instead of doing `finalDamage = damage * (1000/20) * modifiers * (20/actualMaxHp)`, it should just become `finalDamage = damage * modifiers * maxHpModifier`

src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -192,12 +266,12 @@ public static void onPlayerLoggedIn(PlayerEvent.PlayerLoggedInEvent event)
             // World Data (Calendar) Sync Handler
             TerraFirmaCraft.getNetwork().sendTo(new PacketCalendarUpdate(), player);
 
-            // Player nutrients
-            IPlayerNutrients cap = player.getCapability(CapabilityFood.CAPABILITY_PLAYER_NUTRIENTS, null);
+            // Player data
+            IPlayerData cap = player.getCapability(CapabilityPlayer.CAPABILITY_PLAYER_DATA, null);
             if (cap != null)
             {
-                cap.updateNutrientsFastForward();
```
alcatrazEscapee: Why was this call removed? It was specifically there to avoid players who had been logged out of a server for a while from getting a massive decay drop

src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -108,6 +108,26 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
         @Config.RangeDouble(min = 0, max = 10)
         @Config.LangKey("config." + MOD_ID + ".general.playerNutritionDecayModifier")
         public double playerNutritionDecayModifier = 0.2;
+
+        @Config.Comment("Starting base health for the player")
```
alcatrazEscapee: These should be modifiers, since 1000 health is just a display value.

src/main/java/net/dries007/tfc/TerraFirmaCraft.java
```diff
@@ -155,6 +159,15 @@ public void init(FMLInitializationEvent event)
             log.warn("You are not running an official build. Please do not use this and then report bugs or issues.");
         }
 
+        if (event.getSide() == Side.CLIENT)
```
alcatrazEscapee: I'm not convinced this won't just crash on a dedicated server. That's the whole point of having a proxy to do client / server specific event handling

src/main/java/net/dries007/tfc/api/capability/player/CapabilityPlayer.java
```diff
@@ -30,28 +30,28 @@
...
     private static List<Supplier<PotionEffect>> rottenFoodEffects = null;
 
     public static void preInit()
     {
-        // Item nutrient capability
+        // Item player capability
         CapabilityManager.INSTANCE.register(IFood.class, new DumbStorage<>(), FoodHandler::new);
-        // Player nutrient capability
-        CapabilityManager.INSTANCE.register(IPlayerNutrients.class, new DumbStorage<>(), PlayerNutrientsHandler::new);
+        // Player player capability
```
alcatrazEscapee: You need to fix a bunch of these errors all over - take care when doing an refactor in Intellij to not grab comments that aren't part of what you are trying to rename.

src/main/java/net/dries007/tfc/api/capability/player/IPlayerData.java
```diff
@@ -0,0 +1,54 @@
...
+     * Drinks fluid(ie: water) to fill or drain(salt water) thirst bar.
+     * @param value value to fill/drain(negative)
+     */
+
+    /**
+     * Drinks fluid(ie: water) to fill or drain(salt water) thirst bar.
+     *
+     * @param value value to fill/drain(negative)
+     * @return true if the fluid was drank(ie: cooldown)
+     */
```
alcatrazEscapee: Duplicate comments, can remove

src/main/java/net/dries007/tfc/api/capability/player/PlayerDataHandler.java
```diff
@@ -0,0 +1,269 @@
...
+    public float getMaxHealth()
+    {
+        float nutrientPercent = getNutrientAverage() / MAX_PLAYER_NUTRIENTS;
+        float skillPercent = getSkillAverage() / 100; //MAX_PLAYER_SKILLS; This should return a value between 0 and 1.0F
+        float averagePercent = (nutrientPercent + skillPercent) / 2.0f;
+        float maxHealth;
+        if (averagePercent < BASE_HEALTH_THRESHOULD)
+        {
+            //Less than base health
+            if (averagePercent <= MIN_HEALTH_THRESHOULD) return ConfigTFC.GENERAL.playerMinHealth;
```
alcatrazEscapee: This should be modifier based - no config references, no health amount references.

src/main/java/net/dries007/tfc/client/gui/overlay/PlayerDataOverlay.java
```diff
@@ -0,0 +1,178 @@
...
+    private static final ResourceLocation ICONS = new ResourceLocation(MOD_ID, "textures/gui/overlay/icons.png");
+    private static final PlayerDataOverlay INSTANCE = new PlayerDataOverlay();
+
+    public static PlayerDataOverlay getInstance() { return INSTANCE; }
+    private FontRenderer fontrenderer = null;
+    private float maxHealth = 1000, curThirst = 100;
+
+    private PlayerDataOverlay() {}
+
+    public void setMaxHealth(float value) { this.maxHealth = value; }
```
alcatrazEscapee: These should just read directly from the capability. This is potential for errors since there's no guarantee that these will be called when the capability gets updated

src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -68,6 +77,44 @@ public static void onBlockHarvestDrops(BlockEvent.HarvestDropsEvent event)
...
 
+    @SubscribeEvent
+    public static void onRightClickWater(PlayerInteractEvent event)
+    {
+        if (event.getSide() == Side.CLIENT) return;
+        if (!(event instanceof PlayerInteractEvent.RightClickBlock || event instanceof PlayerInteractEvent.RightClickEmpty))
+            return;
+        EntityPlayer player = event.getEntityPlayer();
+        if (!(player.getHeldItemMainhand().isEmpty() && player.getHeldItemOffhand().isEmpty()))
+            return; //Need both hands to drink water
```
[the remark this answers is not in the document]
DisasterMoo: well, it's already harder to drink water using both your hands. Using only one should be even harder, that's why i made impossible, but can be tweaked to drink half the amount or ignore offhand entirely.

src/main/java/net/dries007/tfc/CommonEventHandler.java
DisasterMoo: it wasn't, I refactored the method to onUpdate()

src/main/java/net/dries007/tfc/CommonEventHandler.java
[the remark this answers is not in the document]
alcatrazEscapee: I see your point, but I'm trying to see this from a gameplay perspective. There's no real "two handed" actions in minecraft (I think the only one is the crossbow, which has a real visual indicator that your offhand is not in use. Even so, you can still have something in the offhand). I think this would just be a point of player frustration - I for one basically always have torches in my offhand and having to switch them out every time I want to drink would be a pain

src/main/java/net/dries007/tfc/client/gui/overlay/PlayerDataOverlay.java
DisasterMoo: Looks like I missed changing this back after I fixed the problem I had at the early implementation of the overlay(it wasn't reading capabilities). Fixed in the new commit.

src/main/java/net/dries007/tfc/api/capability/player/PlayerDataHandler.java
DisasterMoo: Changed to use modifiers instead.

src/main/java/net/dries007/tfc/api/capability/player/CapabilityPlayer.java
DisasterMoo: Slap me in the face, I missed another one :X

src/main/java/net/dries007/tfc/TerraFirmaCraft.java
DisasterMoo: Missed the lines below which already check for client side.

src/main/java/net/dries007/tfc/CommonEventHandler.java
DisasterMoo: since i detected: instead of `finalDamage = damage * modifiers * maxHpModifier` => `finalDamage = damage * modifiers / maxHpModifier`
As HP is the inverse of damage.

src/main/java/net/dries007/tfc/api/capability/player/IFood.java
```diff
@@ -125,7 +125,7 @@ default void addNutrientInfo(@Nonnull ItemStack stack, @Nonnull List<String> tex
         else
         {
             text.add(I18n.format("tfc.tooltip.food_expiry_date", CalendarTFC.getTimeAndDate(getRottenDate())));
-            // Show nutrient values if not rotten
+            // Show player values if not rotten
```
alcatrazEscapee: *slapped*

DisasterMoo: :+1:

src/main/java/net/dries007/tfc/CommonEventHandler.java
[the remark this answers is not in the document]
DisasterMoo: Well, to further increase this discussion, this is only for drinking directly from water sources. You can, and should, use containers for that. Glass bottles, Water jugs and such should be used instead of drinking using your hands. It's waaaay easier irl, and at least IMO for TFC perspective, should be a mechanic.

PR #144, opened 2019-06-30
Cooling items in barrel of fresh water (and small vessel showing its items) + some loom fixes

CtrlAltDavid01:


CtrlAltDavid01:
1. It was gold in 1.7.10
2. I'll fix and learn

src/main/java/net/dries007/tfc/api/recipes/BarrelRecipe.java
```diff
@@ -122,4 +124,14 @@ private int getMultiplier(FluidStack inputFluid, ItemStack inputStack)
         }
         return 0;
     }
+
+    public boolean shouldRepeate()
```
alcatrazEscapee: Typo
```suggestion
    public boolean shouldRepeat()
```

src/main/java/net/dries007/tfc/api/recipes/BarrelRecipe.java
```diff
@@ -122,4 +124,14 @@ private int getMultiplier(FluidStack inputFluid, ItemStack inputStack)
...
+    {
+        return false;
+    }
+
+    public void playSound(World world, BlockPos pos, int tickCounter)
```
alcatrazEscapee: This seems like really specific design that will be useful in only one case. I don't like the current implementation - if there's a way to make this actually expandable and useful, i.e. a generic `onRecipeComplete` function then I'll be okay with it, but if it's just an exception for one recipe, I'd rather remove it.

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -220,7 +220,7 @@ public static void registerItems(RegistryEvent.Register<Item> event)
 
         simpleItems.add(register(r, "crop/product/jute", new ItemMisc(Size.TINY, Weight.LIGHT), CT_MISC));
         simpleItems.add(register(r, "crop/product/jute_fiber", new ItemMisc(Size.TINY, Weight.LIGHT), CT_MISC));
-        simpleItems.add(register(r, "crop/product/burlap_cloth", new ItemMisc(Size.TINY, Weight.LIGHT, "cloth"), CT_MISC));
+        simpleItems.add(register(r, "crop/product/burlap_cloth", new ItemMisc(Size.TINY, Weight.LIGHT, "cloth", "cloth_low_quality"), CT_MISC));
```
alcatrazEscapee: Does TFC use these ore dict names? If not, then remove them.

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
```diff
@@ -5,182 +5,172 @@
...
 import static net.dries007.tfc.api.capability.heat.CapabilityItemHeat.ITEM_HEAT_CAPABILITY;
-import static net.dries007.tfc.api.util.TFCConstants.MOD_ID;
 import static net.minecraftforge.fluids.capability.CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY;
 
 @ParametersAreNonnullByDefault
-public class UnmoldRecipe extends IForgeRegistryEntry.Impl<IRecipe> implements IRecipe
+public class UnmoldRecipe extends ShapelessOreRecipe
 {
-    private final ItemMold mold;
+    int moldSlotID;
```
alcatrazEscapee: Creating global variables to hold state between function calls is awful practice here. You have no guarantee that the `IRecipe` methods are called in a specific order, and this is prone to all sorts of horrible errors. If you need to check state in multiple methods, then check the state again - do not rely on whatever was passed in in a different method call.

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
```diff
@@ -5,182 +5,172 @@
...
+            final Metal.ItemType type = Metal.ItemType.valueOf(result.toUpperCase());
+            final String group = JsonUtils.getString(json, "group", "");
 
-    @Override
-    @Nonnull
-    public String getGroup()
-    {
-        return MOD_ID + ":unmold_" + mold.type.name().toLowerCase();
+            //Chance of getting the mold back
+            float chance = 0;
```
alcatrazEscapee: Unnecessary duplication of `Metal.ItemType.moldReturnRate`

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
```diff
@@ -5,182 +5,172 @@
...
+
+    @Override
+    @Nonnull
+    public ItemStack getRecipeOutput() { return ItemStack.EMPTY; }
+
+    @Override
+    public boolean matches(@Nonnull InventoryCrafting inv, @Nonnull World world)
+    {
+        int ingredientCount = 0;
+        RecipeItemHelper recipeItemHelper = new RecipeItemHelper();
```
alcatrazEscapee: Why did you completely rewrite the logic here? This looks much more complex than it needs to be, and for no particular reason. It's a hardcoded recipe that takes one input, of a particular item, with a particular capability.

src/main/java/net/dries007/tfc/objects/te/TEBarrel.java
```diff
@@ -184,9 +186,9 @@ public void update()
         {
             tickCounter++;
 
-            if (tickCounter == 10)
+            if (tickCounter % 10 == 0)
```
alcatrazEscapee: Why?

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
CtrlAltDavid01: I was just thinking it shouldn't be hardcoded like that, but should I return it to it's original state?

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
CtrlAltDavid01: Yeah, some of them are unnecessary, will be removed

src/main/java/net/dries007/tfc/api/recipes/BarrelRecipe.java
CtrlAltDavid01: Yeah, I agree.

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
CtrlAltDavid01: Nither do I know why I did this, will revert

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
CtrlAltDavid01: Okay, I see

src/main/java/net/dries007/tfc/objects/te/TEBarrel.java
CtrlAltDavid01: I liked the sound playing every 4 ticks more then every 5 ticks, so I incresed the tick counter to 20 so it's divisible with 4 and that's to keep that thing at a 10 tick repeat time

CtrlAltDavid01:
I'm sorry for pushing everything in one place, I know that's bad practice, I need to learn more about git, I haven't used it much

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
alcatrazEscapee: Fine with me, but one of them should be removed.

CtrlAltDavid01: Okay, will do!

src/main/java/net/dries007/tfc/api/recipes/BarrelRecipeCoolItems.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public boolean shouldRepeat()
+    {
+        return true;
+    }
+
+    @Override
+    public void onRecipeComplete(World world, BlockPos pos, int tickCounter)
```
alcatrazEscapee: Remove `tickCounter`, just use `world.getTotalTime()`, it's more intuitive and requires less changes to `TEBarrel`

src/main/java/net/dries007/tfc/api/recipes/BarrelRecipeCoolItems.java
```diff
@@ -0,0 +1,66 @@
...
+        int retainAmount = inputFluid.amount - multiplier;
+        if (retainAmount > 0)
+        {
+            return new FluidStack(inputFluid.getFluid(), retainAmount);
+        }
+        return null;
+    }
+
+    @Override
+    public boolean shouldRepeat()
```
alcatrazEscapee: I think this can be removed, as there is no use case for this ever returning false.

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
```diff
@@ -26,21 +32,77 @@
...
+            }
+        }
+        if (moldStack != null)
+        {
+            if (Constants.RNG.nextFloat() <= chance)
+            {
+                EntityPlayer player = ForgeHooks.getCraftingPlayer();
+                if (player != null)
+                {
+                    player.addItemStackToInventory(new ItemStack(moldStack.getItem()));
```
alcatrazEscapee: As discussed in discord - if we make molds unstackable while filled, this be switched to just add to the return value i.e. how normal crafting recipes deal with container items?

src/main/java/net/dries007/tfc/api/recipes/BarrelRecipeCoolItems.java
CtrlAltDavid01: I think it's essential, it returns false in eveery other recipe, but only returns true in this one, so it can repeat, but it may be used in later recipes

src/main/java/net/dries007/tfc/objects/recipes/UnmoldRecipe.java
CtrlAltDavid01: Because of the chance that it gets deleted this doesn't seem possible. There is no way I could find to update the crafting grid from the server, so I don't know how to get around the desync that rand causes

alcatrazEscapee: You shouldn't be updating the crafting grid from anything *except* the server. The client shouldn't perform the crafting, the server should and sync any changes to the client

CtrlAltDavid01: Okay, I know, but when the server performs the craft the item doesn't appear in the grid, it stays invisible, how can I sync it to the client

alcatrazEscapee: The same way containers sync their inventory normally - look into what `WorkbenchContainer` is doing and figure out why you have a desync bug.

PR #161, opened 2019-07-11
Animals

### LINKED ISSUE #159: Animals:
* More animals! (later)
* Spawning of vanilla animals (or TFC variants, see 1.7)
* Fish population - save to chunk data?

DisasterMoo:
Added new capability: Egg, which handles oviparous animal birthing
Added new block: NestBox, which stores and handles eggs
Animals have two categories, mammals and oviparous, which should be obvious for what reasons i made this way.
For now, the Age.OLD isn't used, but can be impl for stopping old animals from mating, birthing children etc.

The only thing missing on this PR is drops, which should be easy.

Closes #159


src/main/java/net/dries007/tfc/objects/entity/animal/EntityAnimalTFC.java
```diff
@@ -0,0 +1,352 @@
...
+        this.matingTime = nbt.getLong("mating");
+        this.fertilized = nbt.getBoolean("fertilized");
+        this.setFamiliarity(nbt.getFloat("familiarity"));
+
+    }
+
+    @Override
+    public boolean isBreedingItem(ItemStack stack)
+    {
+        return DEFAULT_BREEDING_ITEMS.contains(stack.getItem());
```
alcatrazEscapee:
```suggestion
        return OreDictionaryHelper.doesStackMatchOre(stack, "grain")
```
Other things that need to change:
 - since `grain` is already used as an ore dict name for all grain category items, that should probably be changed to be more specific, i.e. `foodCategoryGrain`. Then the ore dict name for grain can be added. However, proper ore dict names for food are already a part of #158, so for now just leave it as is, and it can be fixed once that is merged.

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -103,6 +105,22 @@
     @GameRegistry.ObjectHolder("bloom/refined")
     public static final ItemBloom REFINED_BLOOM = getNull();
 
+    @GameRegistry.ObjectHolder("powder/salt")
+    public static final ItemPowder SALT = getNull();
+
+    @GameRegistry.ObjectHolder("food/barley_grain")
```
alcatrazEscapee: Not necessary if switching to ore dict, see comment in `EntityAnimalTFC`

Issue #157, opened 2019-07-11
Pit Kiln / Placed Item rendering is incomplete / missing for Huge size items (i.e. quern base)


alcatrazEscapee:
- It should render as a single larger item in the center. Mechanics seem good, if invisible.

Issue #137, opened 2019-06-29
Z-Fighting When placed items are covered by straw


muzikbike:
![2019-06-29_17 40 47](https://user-images.githubusercontent.com/52297970/60387013-31557700-9a95-11e9-9be1-58809aa3d5f6.png)
![2019-06-29_17 40 49](https://user-images.githubusercontent.com/52297970/60387014-31557700-9a95-11e9-8d5a-0d144e469b0e.png)

1.6.4 does not have this issue as placed items have their tops somewhat above the top layer of straw.

Issue #63, opened 2019-05-05
Crafting Recipe Mega-Issue


alcatrazEscapee:
Mega-issue for all missing or broken crafting recipes.
 - [x] Concrete powder recipes should be shapeless
 - [x] Flux
 - [x] Saw recipes don't accept damaged saws

alcatrazEscapee:
More info on saw recipes. Using ore dictionary ingredients doesn't respect wildcards, see here:
<https://github.com/MinecraftForge/MinecraftForge/issues/4557>. That says it is fixed, I am not seeing it. If this can't be fixed, it may be worth moving it to a container item implementation instead of `IRecipeFactory`

Red3Tango:
Missing: Stone Knife variants

alcatrazEscapee:
- [ ] Replace `vanilla/iron_bars` with anvil recipe.
 - [ ] `vanilla/redstone/comparator` outputs a lapis block
 - [ ] `vanilla/redstone/piston` should use ore dict for wrought iron tuyere
 - [ ] `vanilla/redstone/tripwire_hook` should use ore dict for wrought iron ingot

Verify / Discuss
 - [ ] Recipes using nether quartz are using the ore dict `gem` instead, is this intended?
 - [ ] Is a tuyere a good item to use a crafting ingredient in the place of a metal "rod / stick" item as per most other mods? A tuyere is supposed to be hollow...

alcatrazEscapee:
- [ ] Support beams are missing a recipe - they need a new `ShapedDamageRecipe` implementation

alcatrazEscapee:
Everything here is done - anything else can receive a separate issue as I believe most of the issues from the mega recipe PR have been solved.

Jummit:
Saw recipes don't accept damaged saws is still an issue in 7657f0bec.

PR #164, opened 2019-07-12
Pit kiln and placed item rendering fix + quern rendering fix

CtrlAltDavid01:
Fixed issue #157 
Fixed issue #137 
Fixed issue #63 
- Stone tools can use bone as tool rod

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/recipes/ShapedDamageRecipe.java
+++ b/src/main/java/net/dries007/tfc/objects/recipes/ShapedDamageRecipe.java
@@ -5,20 +5,12 @@
 
 package net.dries007.tfc.objects.recipes;
 
-import java.util.Map;
-import java.util.Set;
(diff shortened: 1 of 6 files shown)


src/main/java/net/dries007/tfc/objects/recipes/ShapedDamageRecipe.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @SuppressWarnings("unused")
+    public static class Factory implements IRecipeFactory
+    {
+        @Override
+        public IRecipe parse(final JsonContext context, final JsonObject json)
+        {
+            String group = JsonUtils.getString(json, "group", "");
+
+            Map<Character, Ingredient> ingMap = Maps.newHashMap();
```
alcatrazEscapee: Can this be extracted to a helper method in `RecipeUtils` a. la. `parseShapeless()`?

PR #167, opened 2019-07-13
Calendar update + Temperature fixes

alcatrazEscapee:
- Complete refactor of the calendar system in order to be better consistent, compatible, and just generally work better.
 - Adds the capability to track time that players are online, so certain resources don't disappear if a server runs without any players online.

Still todo:
 - [x] Player time is sometimes displaying negative values, this shouldn't be happening.
 - [x] Test plant spawning based on temperature regions (it seems to be too harsh at the moment?)
 - [x] Test sleeping in a bed
 - [x] Test for servers


alcatrazEscapee:
Merged in bdf4b15073501b21378e39f41ea1bf54fc137f6e

PR #172, opened 2019-07-15
Fruit Trees

### LINKED ISSUE #131: Fruit trees
Fruit trees exist in prior versions of TerraFirmaCraft, and their fruits exist in 1.12.2, but their trees and saplings do not seem to.

- Banana
- Cherry
- Green Apple
- Lemon
- Olive
- Orange
- Peach
- Plum
- Red Apple

DisasterMoo:
Added all fruit trees. 

This still needs to be added to world gen.
Closes #131


src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockFruitTreeTrunk.java
```diff
@@ -0,0 +1,339 @@
...
+                                    if (worldIn.getBlockState(missingLeaf).getMaterial().isReplaceable())
+                                    {
+                                        worldIn.setBlockState(missingLeaf, BlockFruitTreeLeaves.get(tree).getDefaultState().withProperty(BlockFruitTreeLeaves.HARVESTABLE, true));
+                                    }
+                                }
+                            }
+                            //Starting with top
+                            for (EnumFacing branchFacing : EnumFacing.VALUES)
+                            {
+                                if (flag) break;
```
alcatrazEscapee: Is there any reason that this loop needs all the `if (flag) break;` instead of just using `break` where you set the flag to true?

DisasterMoo: You're right, i could just change `te.resetCounter()` and use return in the inner loop.

alcatrazEscapee:
I'm still seeing a few issues / problems

 - The first five leaves grow in a very specific order, they should grow randomly. (In `getMissingLeaf`, shuffle the iteration order)
 - Once a branch has grown, leaves try and grow in a diagonal position, and immediately disappear after as there is no connection to the trunk at that point.
 - Trees should break when the bottom block (or any trunk block for that matter) is broken. Should be simple enough to iterate upwards when broken and break any above or to the sides.

PR #199, opened 2019-07-23
Berry Bushes

alcatrazEscapee:
Replaces #174

### DIFF
--- a/generateResources.py
+++ b/generateResources.py
@@ -150,6 +150,19 @@ WOOD_TYPES = [
     'blackwood',
     'palm',
 ]
+BERRY_TYPES = [
+    'blackberry',
+    'blueberry',
+    'bunch_berry',
+    'cloud_berry',
+    'cranberry',
+    'elderberry',
+    'gooseberry',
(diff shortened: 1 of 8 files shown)


alcatrazEscapee:
@DisasterMoo 
This has some major differences from 1.7.10, most of which are unnecessary changes. Remember - this should be a direct port within reason.

1) Bushes should function like 1.7.10 - growth stages are unnecessary. Size is dependent on bush type (some are small, some are medium, some are full block)
2) Bushes are missing world gen
3) Bush items in inventory are incorrectly sized - they render as large blocks when held instead of items
4) Missing feature - bushes should damage the player when they walk through them (certain types, see 1.7.10). All bushes should slow movement similar to grass / plants
5) Bush block textures are scaled wrong with the small bushes - pixels should generally stay the same size. Take note when switching to the standard bush models / sizes.

DisasterMoo:
Done.

I noted in tfc's wiki that it's possible to place another bush on top of a full block one, so i added that too.

PR #201, opened 2019-07-23
Added Rabbits

dotpixel:
Added Rabbits, using vanilla textures and model. No Killer-rabbit (skin pick chance), all rabbits types the same, differ only in texture.

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/entity/animal/EntityRabbitTFC.java
+++ b/src/main/java/net/dries007/tfc/objects/entity/animal/EntityRabbitTFC.java
@@ -8,22 +8,13 @@ package net.dries007.tfc.objects.entity.animal;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
-import net.dries007.tfc.Constants;
-import net.dries007.tfc.util.LootTableListTFC;
-import net.dries007.tfc.util.calendar.CalendarTFC;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockCarrot;
 import net.minecraft.block.state.IBlockState;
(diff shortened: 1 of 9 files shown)


src/main/java/net/dries007/tfc/objects/entity/animal/EntityRabbitTFC.java
```diff
@@ -0,0 +1,456 @@
...
+		this.setRabbitType(compound.getInteger("RabbitType"));
+	}
+
+	public SoundCategory getSoundCategory() {
+		return SoundCategory.NEUTRAL;
+	}
+
+	@Override
+	public boolean isBreedingItem(ItemStack stack) {
+		return stack.getItem() == Items.CARROT; // TFC carrot?
```
alcatrazEscapee: Instead, use an ore dict check for "carrot" - add the ore dict values if nessecary

dotpixel: Do you mean like OreDictionaryHelper.doesStackMatchOre(stack, "carrot")? Why? Carrot is exact one item, no other 'carrot' items.

alcatrazEscapee: There are two - vanilla and TFC variants. Potentially more if other mods want to add them. But the simplest reason is this way it becomes modifiable for pack makers instead of hardcoding it.

dotpixel: Ok

dotpixel:
Сan check. About Deserts, as far as I know TFC have no deserts, so i skip that.

PR #209, opened 2019-07-25
Added Wolfs

dotpixel:
Vanilla like tameable Wolfs, extends TFC EntityAnimalMammal, can be tamed if Familiarity >= some value (0.3 now).


dotpixel:
Done.

PR #212, opened 2019-07-29
Added Fired & Unfired Ceramic Jug with drinking capabilities

xxMrPHDxx:
Added features:
- [x] Support for all fluids (Universal Forge Fluids)
- [x] Store filled fluid into NBT Tags to store it states
- [x] Hold up to 1000 milli buckets (Once only - The value is just for display)
- [x] Show temperature (in Kelvin)
- [x] Can transfer fluids into and out from block with fluid handler capabilities
- [x] Filled fluid shown in the tool tip beside the name
- [x] Use Fluid Handler Capability (Re-Added)
- [x] Able to place temporary flowing liquids


alcatrazEscapee:
This PR needs to be fixed so it isn't modifying *literally every line of code in the entire project*
 - Fix whatever whitespace issues are causing that.
 - Use our code style please. You've hard wrapped things all over the project that we do not want.

And then I'll let you know that our plans for the ceramic jug / wooden bucket was to make them forge fluid containers, with a whitelist of fluids, with the special interaction that they don't actually carry volume - i.e. they can pick up a fluid but can't place it back down (so they just end up placing flowing fluid). As such, based on your description, (because I can't even find your actual changes), this PR needs rework anyways.

src/main/java/net/dries007/tfc/world/classic/fluids/FluidThirstConfig.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Work under Copyright. Licensed under the EUPL.
+ * See the project README.md and LICENSE.txt for more information.
+ */
+
+package net.dries007.tfc.world.classic.fluids;
+
+public class FluidThirstConfig
```
Claycorp: Why isn't this using the already existing config? There isn't a need for a file just for what you can drink.
Oregen is an exception due to the complexity.

xxMrPHDxx: This will enable support for other fluid from other mods too and would make configuration for thirst amount and potion effects later.

src/main/java/net/dries007/tfc/world/classic/fluids/FluidThirstRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+            fluidThirstData = FileUtils.readFileToString(fluidThirstFile, Charset.defaultCharset());
+        }
+        catch (IOException e)
+        {
+            throw new Error("Error reading fluid thirst config file.", e);
+        }
+
+        if (Strings.isNullOrEmpty(fluidThirstData))
+        {
+            TerraFirmaCraft.getLog().warn("The fluid thirst config file is empty! All water will produce thirst value of 0!");
```
Claycorp: Empty files should fail-safe with default values and regenerate. This should never be allowed to be a valid option.

xxMrPHDxx: I just followed the style of `net.dries007.tfc.world.classic.worldgen.vein.VeinRegistry`.
Maybe it's a bad idea

src/main/java/net/dries007/tfc/world/classic/fluids/FluidThirstConfig.java
Claycorp: The config can support that just fine...

Potion effects should be coming from the fluid, not the drinking of a jug.

src/main/java/net/dries007/tfc/world/classic/fluids/FluidThirstRegistry.java
Claycorp: Oregen is special.
If you want to disable thirst it should be done via config not an empty file.

src/main/java/net/dries007/tfc/world/classic/fluids/FluidThirstConfig.java
xxMrPHDxx: Okay, I'll change it later I guess

src/main/java/net/dries007/tfc/world/classic/fluids/FluidThirstRegistry.java
xxMrPHDxx: Great! I'll try that

xxMrPHDxx:
This pull request was broken. Will now be closed

xxMrPHDxx:
More features added to the ceramic jug listed above basically finalized all the features available.

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemJug.java
```diff
@@ -1,101 +1,129 @@
...
+				}
+			}
+			else if(block.hasTileEntity(worldIn.getBlockState(pos)))
+			{
+				TileEntity te = worldIn.getTileEntity(pos);
+				if(te != null && te.hasCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, facing))
+				{
+					IFluidHandler cap = (IFluidHandler) te.getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, facing);
+					FluidStack fluidStack = new FluidStack(jug.getFluid(stack), MAX_FLUID_AMOUNT);
+					if(cap.fill(fluidStack, false) == MAX_FLUID_AMOUNT)
```
Claycorp: Jugs are not buckets.
They can be filled only. Emptying the fluid within should destroy it or place a finite block.

xxMrPHDxx: They can only be filled once and I haven't add the ability to empty the bucket into the world as a source block or something. Should I really do that though? I mean I can destroy it on place of course. Also, there's a small chance of the jug destroyed when you drink though like 2% chance every time you drink.

Claycorp: Unless jugs are going to be the "Wooden bucket" of the rewrite they shouldn't transfer fluids either. I guess that's up to Alcatraz to say then.

xxMrPHDxx: Yeah, probably. You can pretty much transfer liquid into barrels and etc. easily without even a bucket. I guess I need to remove it then if "Wooden bucket" have that feature or else it would be useless

Claycorp: Let Alcatraz look it over first. Strictly speaking 1.7.10 if the jug did the same thing as the wooden bucket and the bucket didn't exist, nothing would have changed really. If they held the chance to break on empty perhaps a better option could be made further up the tech line.

xxMrPHDxx: Okay then. I'll leave it up to the master to decide

src/main/java/net/dries007/tfc/world/classic/fluids/FluidThirstRegistry.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.commons.io.FileUtils;
+
+import com.google.common.base.Strings;
+import com.google.gson.JsonParseException;
+import com.google.gson.reflect.TypeToken;
+
+import net.dries007.tfc.TerraFirmaCraft;
+import net.dries007.tfc.world.classic.worldgen.vein.VeinRegistry;
+
+public enum FluidThirstRegistry
```
alcatrazEscapee: Claycorp already said this - this entire config is way way way too complicated and unnecessary. At most, what needs to be configurable is the possible fluids, which can be done via `String[]` in `ConfigTFC`, i.e. "ceramicJugDrinkableFluids". There's no need for a per-fluid config setting.
TL;DR. Remove all of this.

src/main/java/net/dries007/tfc/world/classic/worldgen/WorldGenOreVeins.java
```diff
@@ -58,6 +58,11 @@
     {
         Random rand = new Random(worldSeed + chunkX * 341873128712L + chunkZ * 132897987541L);
         List<Vein> veins = new ArrayList<>();
+
```
alcatrazEscapee: I said this before - this is a dirty hack that doesn't actually solve the issue here. Remove.

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemJug.java
xxMrPHDxx: The jug can now place temporary flowing fluid for a few seconds

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemFluidStoragePottery.java
```diff
@@ -0,0 +1,231 @@
...
+
+	public ItemFluidStoragePottery(int itemUseDuration)
+	{
+		this.itemUseDuration = itemUseDuration;
+	}
+	
+	@Override
+	public ICapabilityProvider initCapabilities(ItemStack stack, NBTTagCompound nbt)
+	{
+		if(nbt == null) stack.setTagCompound(nbt = new NBTTagCompound());
```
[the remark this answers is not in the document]
xxMrPHDxx: `CapabilityFluidHandler` is too much for the "fixed" capacity of the jug. I implement it anyway just to make it consistent with other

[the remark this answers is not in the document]
alcatrazEscapee: It is absolutely not. Any fluid container should use a fluid capability. No exceptions.

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemJug.java
```diff
@@ -0,0 +1,329 @@
...
+import net.minecraftforge.common.capabilities.Capability;
+import net.minecraftforge.common.capabilities.ICapabilityProvider;
+import net.minecraftforge.fluids.BlockFluidBase;
+import net.minecraftforge.fluids.Fluid;
+import net.minecraftforge.fluids.FluidRegistry;
+import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.fluids.capability.CapabilityFluidHandler;
+import net.minecraftforge.fluids.capability.IFluidHandler;
+
+public class ItemJug extends ItemPottery implements ICapabilityProvider
```
alcatrazEscapee: Don't make the item instance the capability provider - the provider is specific to the `ItemStack` instance, not the `Item`. As such instance fields should be on a separate provider / capability implementation. See how molds have a joint heat/fluid/item capability implementation

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemJug.java
```diff
@@ -0,0 +1,329 @@
...
+{
+
+	public static final int MAX_FLUID_AMOUNT = 1000;
+    private static final int MAX_USE_DURATION = 32;
+	private static final float BREAK_CHANCE = 0.02f;
+	
+	private NBTTagCompound getFluidTank(ItemStack stack)
```
alcatrazEscapee: This is all still using NBT hacks. Use capabilities, they are much cleaner and much better for this kind of usage.

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemUnfiredJug.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the project README.md and LICENSE.txt for more information.
+ */
+
+package net.dries007.tfc.objects.items.ceramics;
+
+import net.dries007.tfc.api.capability.size.Size;
+import net.dries007.tfc.api.capability.size.Weight;
+import net.minecraft.item.ItemStack;
+
+public class ItemUnfiredJug extends ItemPottery
```
alcatrazEscapee: This class isn't necessary due to changes in pottery item structure

alcatrazEscapee:
Superseded by #254 as it was closer to what we are looking for.

PR #213, opened 2019-07-29
Rework on fissure generation

### LINKED ISSUE #51: Fissures generation problems
* Current causing cascading lag (see #40)
* Water / lava doesn't seem to rise to surface.

DisasterMoo:
Closes #51 

I'm certain that this won't cause cascading lag, but could use more tweaking on rock placement logic. I'm leaving the old method there disabled for anything.


DisasterMoo:
The new loooks:

scr1
scr2
scr3
scr4

PR #226, opened 2019-08-03
Added CraftTweaker and JEI Integration

### LINKED ISSUE #228: Recipe System Refactoring / Craft Tweaker / JEI
See https://gist.github.com/alcatrazEscapee/c078386d7a41ddb48d0010b25fdec109 for an idea of how to properly do recipes in TFC

 - [x] Move stuff to `IIngredient<ItemStack>` from `ItemStack`
 - [ ] Add JEI Categories for all forms of TFC recipes.
 - [ ] Add Craft Tweaker hooks for pack makers (low priority)

DisasterMoo:
Replaces #194 
Also removed redundancy in `LoomRecipe`

Closes #228


src/main/java/net/dries007/tfc/compat/crafttweaker/CTItemRegistry.java
```diff
@@ -0,0 +1,235 @@
...
+                public String describe()
+                {
+                    return "Registered heat capacity for " + input.toCommandString();
+                }
+            });
+        }
+    }
+
+    @ZenMethod
+    public static void registerHeatRecipe(String registryName, IItemStack input, IItemStack output, float transformTemp, float maxTemp)
```
alcatrazEscapee: My suggestion for this:
 - Move it to it's own class, i.e. `CTHeatRecipe`
 - Add remove variants for `HeatRecipe`, targeting `HeatRecipeSimple`, as the others shouldn't really be modifiable via scripts as they are too dynamic.

src/main/java/net/dries007/tfc/compat/jei/categories/HeatMetalCategory.java
```diff
@@ -0,0 +1,61 @@
+/*
```
alcatrazEscapee: Can `HeatMetalCategory` and `HeatSimpleCategory` be merged into a generic heat recipe category? It should show the input, and both possible outputs (items and fluids), along with the temperature (in red, yellow, brilliant white, etc.)

alcatrazEscapee:
After testing a bit, I have a few change requests:
 - The anvil should show step icons, not just green + red squares, as well as the required steps for the item. i.e. what you see on the anvil GUI if you're about to make that item.
 - Welding should not show the full anvil GUI + buttons, as everything there is unused. Just show the input + ouputs and an arrow, since the operation is actually an in world craft
 - Anvil should show all hammers (ore dict "hammer") and all flux (ore dict "flux") instead of wrought iron + TFC flux item
 - Fire clay should use a different background texture (this is a more general bug with the knapping GUI as well, not a specific JEI bug, could it be fixed? the missing texture is [this one](https://github.com/Deadrik/TFCraft/blob/master/src/Resources/assets/terrafirmacraft/textures/items/pottery/clay%20flat%20dark%20fire.png)
 - Alloy %s are very small - maybe make them stacked with the %s displayed on the side?

PR #229, opened 2019-08-04
Wooden Bucket

DisasterMoo:


src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -114,6 +115,9 @@
     @GameRegistry.ObjectHolder("powder/salt")
     public static final ItemPowder SALT = getNull();
 
+    @GameRegistry.ObjectHolder("wooden_bucket")
```
alcatrazEscapee: Object holder annotations not necessary if the field name matches the registry name (case insensitive)

src/main/java/net/dries007/tfc/objects/items/wood/ItemBucketTFC.java
```diff
@@ -0,0 +1,159 @@
...
+{
+    public static final Set<Fluid> ALLOWED_BUCKET_FLUIDS;
+    private static final int CAPACITY = 1000;
+
+    static
+    {
+        ALLOWED_BUCKET_FLUIDS = new HashSet<>();
+        for (String fluidName : ConfigTFC.GENERAL.woodenBucketWhitelist)
+        {
+            ALLOWED_BUCKET_FLUIDS.add(FluidRegistry.getFluid(fluidName));
```
alcatrazEscapee: I have one concern here: this could potentially run too early, before fluids have been registered. TFC will register Fluids -> Blocks -> Items, so it shouldn't be a problem there, but other mods may wait until init or after items.

My suggestion would be to move this to an on-demand / lazy loaded calculation in `initCapabilities`, as I doubt it's that resource intensive.

src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -69,6 +69,10 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
         @Config.LangKey("config." + MOD_ID + ".general.leafStickDropChanceBonusClasses")
         public String[] leafStickDropChanceBonusClasses = new String[] {"knife", "scythe"};
 
+        @Config.Comment("List of fluids allowed to be picked up by wooden bucket")
+        @Config.LangKey("config." + MOD_ID + ".general.woodenBucketWhitelist")
+        public String[] woodenBucketWhitelist = new String[] {"fresh_water", "hot_water", "salt_water", "water", "limewater", "tannin", "olive_oil", "vinegar", "rum", "beer", "whiskey", "rye_whiskey", "corn_whiskey", "sake", "vodka", "cider", "brine", "milk_curdled", "milk_vinegar"};
```
alcatrazEscapee: Oh, and as of b7c9c2e6a105d91e56d6c359925ea2099113538c I'm going to be picky about this: Please add an `en_us.lang` entry with the lang key and tooltip of any config values you add. See the existing ones for style - the name should be a readable version of the variable name and the tooltip should match the comment.

PR #251, opened 2019-08-05
Improve support for additional health mods/Absorption

GreyMario:
Cosmetic changes to the health bar renderer to support additional colors when health reaches high enough.

Config option to switch to "single bar" mode (if the above is not desired by the user) in which all health capacity is crammed into the red health bar.

Adds subtle shadow to avoid white text on lighter parts of health bar blending together.

Requires change to TFC-Assets (https://github.com/TerraFirmaCraft/TFC-Assets/pull/59).


src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -186,6 +186,10 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
         @Config.Comment("Should the prospectors pick output to the actionbar? (the space just above the hotbar)")
         @Config.LangKey("config." + MOD_ID + ".client.propickOutputToActionBar")
         public boolean propickOutputToActionBar = true;
+
+        @Config.Comment("Show extra player health using additional health bar overlays? (for Absorption or increased max health)")
+        @Config.LangKey("config." + MOD_ID + ".client.showExtraHealthWithOverlays")
+        public boolean showExtraHealthWithOverlays = true;
```
alcatrazEscapee: As of b7c9c2e6a105d91e56d6c359925ea2099113538c I'm going to be picky about this: Please add an `en_us.lang` entry with the lang key and tooltip of any config values you add. See the existing ones for style - the name should be a readable version of the variable name and the tooltip should match the comment.

GreyMario:
>code style

Huh. I definitely imported the code style, and it hasn't actually complained about anything at all. Let me double check, I guess.

src/main/java/net/dries007/tfc/ConfigTFC.java
GreyMario: Lang file updated to match.

GreyMario:
Closing on the grounds of Claycorp's "I have a banhammer, so the fact that I'm _provably wrong_ doesn't matter" attitude makes me not want to contribute to the project at all.

PR #254, opened 2019-08-05
Added jug with drinkable ability.

DisasterMoo:


src/main/java/net/dries007/tfc/objects/items/wood/ItemWoodenBucket.java
```diff
@@ -173,6 +173,7 @@ public void getSubItems(CreativeTabs tab, NonNullList<ItemStack> items)
     {
         if (isInCreativeTab(tab))
         {
+            items.add(new ItemStack(this));
```
DisasterMoo: I forgot to comment why i'm adding this: It adds an empty bucket to the creative tab before adding the filled ones.

alcatrazEscapee: I must've forgotten that, thanks.

PR #264, opened 2019-08-08
Bugfixing

### LINKED ISSUE #233: Thatch should block light + rain per 1.7
In 1.7, thatch would block light fully (which had the original function of creating dark, cool rooms for storage) and would block rainfall.

The original light opacity of thatch was `255`. This is up for debate if light level no longer affects decay etc.

Thatch doesn't block rainfall currently because of its `Material.PLANTS` definition; it will need a material for which `blocksMovement` returns `true`, such as what it used in 1.7, `Material.GRASS`. However, `Material.GRASS` is also `isSolid`, killing grass underneath it; if this is not intended, a new material will have to be created.

### LINKED ISSUE #235: Zombie mobs drop vanilla minecraft potatoes
Zombie mobs can occasionally drop vanilla Minecraft potatoes. These potatoes can be planted anywhere on the ground to let grow. Breaking the planted potatoes resets their decay time so they never go bad. Letting the crop mature just gets you more vanilla potatoes like default behavior.

### LINKED ISSUE #257: Bones
Animals should drop bones when killed for consistency with 1.7.10 mechanics.

### LINKED ISSUE #260: Mortar + Bricks crafting recipe
- [x] Add mortar item
 - [x] Add barrel recipes for mortar
 - [x] Updated stone crafting recipes to use mortar + brick items (brick blocks, brick walls, etc.)

### LINKED ISSUE #263: Bloomery recipe missing
The Bismuth Bronze and Black Bronze recipes for the bloomery are missing, only the regular bronze recipe works.

### LINKED ISSUE #266: Placing items on side of block replaces block above it
When using the "Place Item Holder Block" control on the side of a block, the block above is replaced by the placed item. This erases the need to have any tools in order to dig upwards, and deletes the block itself.

### LINKED ISSUE #268: Alternate ladder recipe
In 1.7.10 you could craft 16 ladders with 6 lumber

### LINKED ISSUE #270: config missing & confusing naming issue
You can see these should follow a same naming rule but they act differently. I think latter 2 shouldn't have space in their name.

I bet this can be found in lang files, but I don't know why it doesn't show.

### LINKED ISSUE #271: JEI-Related barrel recipe issues
1.Barrel-related recipes can only be shown using acacia barrel.

2.Incorrect recipe: water becomes rum with nothing

### LINKED ISSUE #272: Wool recipe issue
Using loom:

Craft directly:

### LINKED ISSUE #276: Bloom item melting using unknown temperature instead of wrought iron's

### LINKED ISSUE #278: JEI doesn't show name of forgerule, instead shows keyname

### LINKED ISSUE #279: Oak boat recipe not working
Nothing happens in JEI when you click on the recipe, trying it on the workbench doesn't work.

Seems to be working with other types of wood, was able to build a hickory boat.

Tested on build #44.

### LINKED ISSUE #280: Putting refined bloom in a crucible creates "unknown alloy"
Putting a refined bloom of any size into an empty crucible (or one with wrought iron) creates an "unknown alloy" instead of wrought iron like it should.

For now, you can work around this by refining the bloom using an anvil.

DisasterMoo:
Closes #233 
Closes #235 
Closes #257 (in Assets PR)
Closes #260 
Closes #263 
Closes #266 
Closes #268 (in Assets PR)
Closes #270 
Closes #271 
Closes #272 
Closes #276 
Closes #278 
Closes #279 
Closes #280

### DIFF
--- a/src/main/java/net/dries007/tfc/network/PacketPlaceBlockSpecial.java
+++ b/src/main/java/net/dries007/tfc/network/PacketPlaceBlockSpecial.java
@@ -46,7 +46,7 @@ public class PacketPlaceBlockSpecial implements IMessageEmpty
                         double placeReach = player.getEntityAttribute(EntityPlayer.REACH_DISTANCE).getAttributeValue();
                         if (player.getDistanceSq(pos) <= placeReach * placeReach && hitFace != null)
                         {
-                            if (!stack.isEmpty() && world.getBlockState(pos).isNormalCube() && world.getBlockState(pos.offset(hitFace)).getBlock().isReplaceable(world, pos))
+                            if (!stack.isEmpty() && world.getBlockState(pos).isNormalCube() && hitFace == EnumFacing.UP && world.getBlockState(pos.up()).getBlock().isReplaceable(world, pos))
                             {
                                 world.setBlockState(pos.up(), BlocksTFC.PLACED_ITEM.getDefaultState());
                                 TEPlacedItem tile = Helpers.getTE(world, pos.up(), TEPlacedItem.class);
--- a/src/main/java/net/dries007/tfc/network/PacketPlaceBlockSpecial.java
+++ b/src/main/java/net/dries007/tfc/network/PacketPlaceBlockSpecial.java
@@ -46,16 +46,7 @@ public class PacketPlaceBlockSpecial implements IMessageEmpty
                         double placeReach = player.getEntityAttribute(EntityPlayer.REACH_DISTANCE).getAttributeValue();
                         if (player.getDistanceSq(pos) <= placeReach * placeReach && hitFace != null)
                         {
-                            if (!stack.isEmpty() && world.getBlockState(pos).isNormalCube() && hitFace == EnumFacing.UP && world.getBlockState(pos.up()).getBlock().isReplaceable(world, pos))
-                            {
-                                world.setBlockState(pos.up(), BlocksTFC.PLACED_ITEM.getDefaultState());
-                                TEPlacedItem tile = Helpers.getTE(world, pos.up(), TEPlacedItem.class);
-                                if (tile != null)
-                                {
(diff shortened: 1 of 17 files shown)


src/main/java/net/dries007/tfc/network/PacketPlaceBlockSpecial.java
```diff
@@ -46,7 +46,7 @@ public IMessage onMessage(PacketPlaceBlockSpecial message, MessageContext ctx)
                         double placeReach = player.getEntityAttribute(EntityPlayer.REACH_DISTANCE).getAttributeValue();
                         if (player.getDistanceSq(pos) <= placeReach * placeReach && hitFace != null)
                         {
-                            if (!stack.isEmpty() && world.getBlockState(pos).isNormalCube() && world.getBlockState(pos.offset(hitFace)).getBlock().isReplaceable(world, pos))
+                            if (!stack.isEmpty() && world.getBlockState(pos).isNormalCube() && hitFace == EnumFacing.UP && world.getBlockState(pos.up()).getBlock().isReplaceable(world, pos))
```
alcatrazEscapee: Instead of restricting this to targeting the top block, change this to use `pos.offset(hitFace)` where necessary.

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -321,6 +325,15 @@ public static void registerItems(RegistryEvent.Register<Item> event)
         // todo: foods & plants & seeds & fruits & fruit tree saplings & berries & berry bushes
 
         allSimpleItems = simpleItems.build();
+        // Register oredict bronze items as any bronze
```
alcatrazEscapee: This should either be in the `ItemMetal` constructor, or if it really needs to be separate, in `OreDictionaryHelper#init` imo. Preferably the former.

alcatrazEscapee:
Merged.

PR #273, opened 2019-08-08
Added thatch bed.

### LINKED ISSUE #227: Straw + Thatch Bed
- Right click to set spawn
 - Allows player to sleep through thunder storms, not through night. (same as 1.7.10)

DisasterMoo:
Closes #227


DisasterMoo:
Had to change extending `BlockBed` (it expects a `TileEntity` as it implements `ITileEntityProvider`, which in turn force to use TESR instead of models) and copy paste most of it's code, but now it's working with a static model.

alcatrazEscapee:
iirc can still extend `BlockBed` - just override `createTileEntity`, `hasTileEntity` and `createNewTileEntity` to all return null / false.

Also implementing `ITileEntityProvider` has nothing to do with expecting a `TESR` - it just indicates it has a tile entity. The renderer is a totally separate part.

DisasterMoo:
> iirc can still extend `BlockBed` - just override `createTileEntity`, `hasTileEntity` and `createNewTileEntity` to all return null / false.

I did try that, but still couldn't make it use the model file.

alcatrazEscapee:
Just override the methods in `BlockBed` that delegate to using a TESR for rendering. See `getRenderType`, `getRenderLayer`, `hasCustomBreakingProgress`, etc. Basically anything in `BlockBed` that is overridden from it's default in `Block`

TESR's are bound to Tile Entities - so if you don't return a TE, it won't ask you for a TESR.

PR #274, opened 2019-08-08
Fluid Rewrite

alcatrazEscapee:
- All TFC fluids now are stored as wrappers, which hold the fluid instance (either the TFC created one, or a default one), and records if it is default or not
 - Blocks are registered for non-default fluids, using the wrapper.
 - Moved Drinkables and Metal fluids into their own wrapper classes

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
+++ b/src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
@@ -112,13 +112,13 @@ public final class FluidsTFC
     {
         allInfiniteFluids = ImmutableSet.<FluidWrapper>builder()
             .add(
-                HOT_WATER = registerFluid(new Fluid("hot_water", STILL, FLOW, 0xFF345FDA).setTemperature(350), (fluid, isDefault) -> new DrinkableFluidWrapper(fluid, isDefault, player -> {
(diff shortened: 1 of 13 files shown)


src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
```diff
@@ -5,147 +5,194 @@
...
 
-            for (Metal metal : TFCRegistries.METALS.getValuesCollection())
+    public static void preInit()
+    {
+        allInfiniteFluids = ImmutableSet.<FluidWrapper>builder()
+            .add(
+                HOT_WATER = registerFluid(new Fluid("hot_water", STILL, FLOW, 0xFF345FDA).setTemperature(350), (fluid, isDefault) -> new DrinkableFluidWrapper(fluid, isDefault, player -> {
+                    if (player.getFoodStats() instanceof FoodStatsTFC)
+                    {
+                        ((FoodStatsTFC) player.getFoodStats()).addThirst(40);
```
DisasterMoo: You probably missed, this is making hot_water behaves like fresh_water.

src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
```diff
@@ -5,147 +5,194 @@
...
+                    }
+                })),
+                FRESH_WATER = registerFluid(new Fluid("fresh_water", STILL, FLOW, 0xFF1F32DA)),
```
DisasterMoo: And fresh water not being drinkable

PR #277, opened 2019-08-09
Fix #270 different language requires different format

### LINKED ISSUE #270: config missing & confusing naming issue

HeartyYF:
NOTE: This is **not** the best way to solve the problem(maybe the worst way). The best way, as I've claimed, is to use "%s %s" in the language file.


alcatrazEscapee:
Yes, you are right, this is **not** the way to solve this. We are not going to hard code a fix to one specific translation issue.

PR #281, opened 2019-08-10
More bugfixing

### LINKED ISSUE #43: Morning glories can spawn on leaves
https://i.imgur.com/w70e78v.png

### LINKED ISSUE #133: Nitpick with grass slowing players
Certain tall plants, namely fescue grass and switchgrass, slow the player down as they try to walk through them. While this makes sense, the slowing effect is applied by the plant regardless of its age. This can become slightly irritating when the plant is very young, and barely reaches up to the player's knees, less so than other non-slowing species of grass.

I suggest that the slowing effect is only imparted by grasses which have "part" set to "lower", as this implies that the grass is full-sized and would believably impede movement.

### LINKED ISSUE #235: Zombie mobs drop vanilla minecraft potatoes

### LINKED ISSUE #247: Drinking may cause thirst bar to flicker slightly
HArd to capture with a screenshot, but when holding right click on the water, sometimes the thirst fills up for a frame, then flashes back down, before going back up again.

### LINKED ISSUE #255: Adding logs to a pit kiln does not play a sound
I swear it did before.

DisasterMoo:
Closes #43 
Closes #133 
Closes #235 
Closes #247 
Closes #255

### DIFF
--- a/src/main/java/net/dries007/tfc/TerraFirmaCraft.java
+++ b/src/main/java/net/dries007/tfc/TerraFirmaCraft.java
@@ -32,6 +32,7 @@ import net.dries007.tfc.client.gui.overlay.PlayerDataOverlay;
 import net.dries007.tfc.client.render.animal.RenderAnimalTFCFamiliarity;
 import net.dries007.tfc.command.*;
 import net.dries007.tfc.network.*;
+import net.dries007.tfc.objects.LootTablesTFC;
 import net.dries007.tfc.objects.entity.EntitiesTFC;
 import net.dries007.tfc.objects.items.ItemsTFC;
 import net.dries007.tfc.proxy.IProxy;
@@ -181,6 +182,7 @@ public final class TerraFirmaCraft
             GuiIngameForge.renderArmor = false;
             GuiIngameForge.renderExperiance = false;
         }
+        MinecraftForge.EVENT_BUS.register(LootTablesTFC.getInstance()); //Register removal loot entries in init so it won't crash
 
         worldTypeTFC = new WorldTypeTFC();
         GameRegistry.registerWorldGenerator(new RarityBasedWorldGen(x -> x.lavaFissureRarity, new WorldGenFissure(true)), 0);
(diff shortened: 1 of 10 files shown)


src/main/java/net/dries007/tfc/TerraFirmaCraft.java
```diff
@@ -181,6 +182,7 @@ public void init(FMLInitializationEvent event)
             GuiIngameForge.renderArmor = false;
             GuiIngameForge.renderExperiance = false;
         }
+        MinecraftForge.EVENT_BUS.register(LootTablesTFC.getInstance()); //Register removal loot entries in init so it won't crash
```
alcatrazEscapee: This can be reverted to static, as per discord discussion.

src/main/java/net/dries007/tfc/world/classic/worldgen/WorldGenPlantTFC.java
```diff
@@ -113,7 +115,12 @@ public boolean generate(World worldIn, Random rand, BlockPos position)
                 for (int i = 0; i < ChunkDataTFC.getRainfall(worldIn, position) / 16; ++i)
                 {
                     BlockPos blockpos = position.add(rand.nextInt(4) - rand.nextInt(4), rand.nextInt(4) - rand.nextInt(4), rand.nextInt(4) - rand.nextInt(4));
-
+                    //Make sure this is a good position to place (eg: not on top of leaf blocks, replaceable blocks)
+                    while (!worldIn.isAirBlock(blockpos) || worldIn.getBlockState(blockpos.down()).getBlock() instanceof BlockLeaves || worldIn.getBlockState(blockpos.down()).getMaterial().isReplaceable())
```
alcatrazEscapee: This shouldn't be duct-taped on here - it should become part of `BlockPlantTFC` or whichever is used for moss type plants. Ideally there's a flag for if it requires a solid block to support it. May require an additional plant type.

PR #300, opened 2019-08-13
Make Water Sane Again

### LINKED ISSUE #146: Fluids are goddammed broken
word.

DavidTriphon:
Attempts to fix #146

Makes liquids interact as if they were all the same liquid.


src/main/java/net/dries007/tfc/objects/blocks/BlockFluidTFC.java
```diff
@@ -0,0 +1,424 @@
...
+        if (adjState.getMaterial().isSolid())
+            return -1;
+
+        // adjacent air or non-mergeable liquids
+        return 0;
+    }
+
+    protected boolean isMergeableFluid(@Nonnull IBlockState blockstate)
+    {
+        return (blockstate.getMaterial() == getDefaultState().getMaterial()) && (blockstate.getMaterial().isLiquid());
```
alcatrazEscapee: I think this should do a `blockstate.getBlock() instanceof BlockFluidTFC` instead of comparing by material.
 - I don't see a particular reason why water can't merge with alchohol - It's certainly not obvious to the player why not.
 - We don't want to override fluid blocks that aren't TFC ones.
If there's something I'm missing here let me know.

PR #328, opened 2019-08-24
Bugfixing

DisasterMoo:


src/main/java/net/dries007/tfc/objects/te/TEPitKiln.java
```diff
@@ -291,6 +291,22 @@ public boolean tryLight()
                 burnTicksToGo = ConfigTFC.GENERAL.pitKilnTime;
                 updateBlock();
                 world.setBlockState(above, Blocks.FIRE.getDefaultState());
+                //Light other adjacent pit kilns
+                for (EnumFacing solidBlock : EnumFacing.HORIZONTALS)
```
alcatrazEscapee: This is more than adjacent pit kilns - just store a list of positions (the four diagonals) and iterate through those. i.e.
```java
Set<BlockPos> ADJACENT_POS = Set.of(new BlockPos(-1, 0, -1), ...
```

PR #340, opened 2019-08-27
Skills

alcatrazEscapee:
- Prospecting updated to be less grindy and more based on finding ore
 - Smithing still WIP
 - Agriculture and Butchering ported directly
 - GUI still missing a few compatibility options for multiple skills

Closes #163

PR #460, opened 2019-09-24
Skills

### LINKED ISSUE #163: Skills
- These are already scaffolded, talk to Alcatraz about how to continue with this.
 - Prospecting - effect is easy to implement, change leveling method to be ore based?
 - Smithing (all kinds) - need to figure out what kind of bonuses this will give to the tool. (Enchantments?)
 - Agriculture / Butchery - drops increace based on skill level, possible requirement for unique crops / animals?

DisasterMoo:
Supersedes #340 

Everything from #340 plus:
- Added `/skills` command to set skill level.
- Gui progress bar are colored based on skill tier.

Closes #163


src/main/java/net/dries007/tfc/api/capability/food/FoodStatsTFC.java
```diff
@@ -30,7 +30,8 @@
 @ParametersAreNonnullByDefault
 public class FoodStatsTFC extends FoodStats implements IFoodStatsTFC
 {
-    public static final float PASSIVE_HEAL_AMOUNT = 20 * 0.0002f; // On the display: 1 HP / 5 second
```
alcatrazEscapee: These should all be reverted - they were part of a fix for nutrition decay that got fixed on the main branch instead

src/main/java/net/dries007/tfc/network/PacketSkillsUpdate.java
```diff
@@ -0,0 +1,64 @@
...
+import net.minecraftforge.fml.common.network.simpleimpl.IMessage;
+import net.minecraftforge.fml.common.network.simpleimpl.IMessageHandler;
+import net.minecraftforge.fml.common.network.simpleimpl.MessageContext;
+
+import io.netty.buffer.ByteBuf;
+import net.dries007.tfc.TerraFirmaCraft;
+import net.dries007.tfc.api.capability.player.CapabilityPlayerData;
+import net.dries007.tfc.api.capability.player.IPlayerData;
+
+public class PacketSkillsUpdate implements IMessage
```
alcatrazEscapee:
```suggestion
public class PacketPlayerDataUpdate implements IMessage
```
Since skills got renamed to more generic player data capability

src/main/java/net/dries007/tfc/objects/items/metal/ItemProspectorPick.java
```diff
@@ -56,6 +59,13 @@ public EnumActionResult onItemUse(EntityPlayer player, World worldIn, BlockPos p
 
             if (!worldIn.isRemote)
             {
+                float falseNegativeChance = 0.4f; //Classic value was random(100) >= (60 + rank)
```
alcatrazEscapee: I think we should keep the same calculation. It lowers the bar for "no false negatives", which is the only important milestone to the player (continuous incremental gains aren't rewarding, better stats on a new tier is). Thus false negative chance is `0.4 - skill.getTier().ordinal() * 0.1`

src/main/java/net/dries007/tfc/util/skills/SimpleSkill.java
```diff
@@ -0,0 +1,71 @@
...
+    public float getLevel()
+    {
+        // checks >=4f for full progress bar in MASTER tier.
+        return amount >= 4f ? 1.0F : amount % 1.0F;
+    }
+
+    @Override
+    public void setTotalLevel(double value)
+    {
+        amount = (float) value * 4f;
```
alcatrazEscapee: This should do the same verification check that `add` does - don't allow a value outside [0, 1]

src/main/java/net/dries007/tfc/util/skills/SmithingSkill.java
```diff
@@ -0,0 +1,165 @@
...
+        return totalSkill >= 160 ? 1.0F : (totalSkill % 40) / 40.0f;
+    }
+
+    @Override
+    public void setTotalLevel(double value)
+    {
+        //Evenly distribute value accordingly
+        for (Type smithType : Type.values())
+        {
+            skillLevels[smithType.ordinal()] = (int) (value * smithType.getMax());
```
alcatrazEscapee: This should do a sanity check on value, don't allow a value outside [0, 1]

src/main/java/net/dries007/tfc/objects/items/metal/ItemProspectorPick.java
alcatrazEscapee: Sorry, I made a mistake in the above calculation. The intent is for each tier to reduce false negatives by 10%, with all of master (not just 100% skill in master) giving no false negatives. You don't need any complicated logic here. The only change from above is `0.3 - skill.getTier().ordinal() * 0.1`

Issue #436, opened 2019-09-19
Config option to disable the ability to place ingot stacks


UnlimatedStone9:
I think it would be a nice improvement to be able to, optionally, disable the ability to make ingot piles, given other mods, such as charset immersion, do this mechanic better, and allow creating stacks from ingots of any mod, I think it would be nice if people were able to disable TFC's ingot stack mechanics, for consistency.

It could also be done if people just want ingots to only be store-able in containers because they don't like ingot stacks for whatever reason.

Merely a quality of life enhancement for customization, nothing more.

alcatrazEscapee:
> do this mechanic better, and allow creating stacks from ingots of any mod
So the real suggestion here is to improve TFC's ingot piles?

I am opposed to just adding in disable config options for various mechanics or features unless there's a very specific need for it. *Especially* with the current state of the mod, "X doesn't work well or isn't fully implemented" is a really bad reason for asking to disable it.

UnlimatedStone9:
Well, given that TFC ingot piles only handle ingots, and aren't gamebreaking to be disabled (i.e not needed for a large content adding feature, merely cosmetic), and the only downside to disabling them is not having easy storage (which likely isn't an issue if you're disabling them anyways), I think it would be a useful addition for players to be able to disable them.
Unless you're willing to allow ingot piles to handle sheets, plates, gears, rods, gems/crystals, double ingots, etc alongside regular ingots, TFC's ingot piles are just bloat when using better alternatives.
While yes TFC isn't meant to be torn apart for better alternatives to replace, it is nice to have the option to disable small parts of the mod.
A player might just not want ingot piles, period. That's up to them to disable it then.

Just my thoughts.

PR #462, opened 2019-09-25
Moar bugfixing and enhancing the game!

DisasterMoo:
Commits should describe each one.


src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -144,15 +167,32 @@ public static void onRightClickBlock(PlayerInteractEvent.RightClickBlock event)
         IPlaceableItem placeable = IPlaceableItem.Impl.getPlaceable(stack);
         if (placeable != null)
         {
-            if (placeable.placeItemInWorld(world, pos, stack, player, event.getFace(), event.getHitVec()))
+            // Dirty trick to fire onBlockActivated first
```
alcatrazEscapee: I don't like the fact that this is half-implemented. It should follow the exact same logic chain that vanilla does in `PlayerInteractionManager#handleRightClick`. That is, always cancel the event if something matches, then duplicate the rest of the function (all of the relevant context is available). When we get to `onItemUse`, handle our placeable instead of `onItemUse`

Minecraft's right click handling is very detailed and handles a lot of cases. I do not think it's a good idea to just blindly fire `onBlockActivated` without properly considering the effects that may have

src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -203,6 +243,16 @@ public static void onRightClickItem(PlayerInteractEvent.RightClickItem event)
         IPlaceableItem placeable = IPlaceableItem.Impl.getUsable(stack);
         if (placeable != null)
         {
+            if (IPlaceableItem.Impl.getPlaceable(stack) != null)
```
alcatrazEscapee: This should not be necessary. `onRightClickItem` is fired through a totally separate call path, when you right click without targeting a block (that's within your reach distance). This also (again) ignores the standard placement logic.

src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -257,6 +257,18 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
...
+        @Config.LangKey("config." + MOD_ID + ".general.oreDictIron")
+        public boolean oreDictIron = false;
+
+        @Config.Comment("Add plate ore dictionary to sheet items?")
+        @Config.LangKey("config." + MOD_ID + ".general.oreDictPlate")
+        public boolean oreDictPlate = false;
+
+        @Config.Comment("Disable the ability to create ingot piles?")
+        @Config.LangKey("config." + MOD_ID + ".general.disableIngotPile")
+        public boolean disableIngotPile = false;
```
alcatrazEscapee: This is something I am really reluctant to do - it opens the door to a whole range of modularity based problems - i.e. disabling and enabling specific parts of the mod which causes other parts to have to sanity check that they are enabled. I do not think that is a viable approach to take with this mod, mainly due to the size and complexity.

Also, this seems like a bandaid fix to an actual problem - TFC ingot piles don't work that well. So instead of just slapping a disable config on them, can we actually make them work better?

src/main/java/net/dries007/tfc/objects/items/ItemFireClay.java
```diff
@@ -1,48 +0,0 @@
...
-import net.minecraft.util.ActionResult;
-import net.minecraft.util.EnumActionResult;
-import net.minecraft.util.EnumHand;
-import net.minecraft.world.World;
-
-import net.dries007.tfc.api.capability.size.Size;
-import net.dries007.tfc.api.capability.size.Weight;
-import net.dries007.tfc.client.TFCGuiHandler;
-
-public class ItemFireClay extends ItemTFC
```
alcatrazEscapee: Why was this moved from a custom item class to an `IPlaceable`? `IPlaceable` is specifically a hack for when we can't / don't want to use a custom item class because of inheritance or inaccessibility reasons - not a general use case for onItemUse.

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -313,6 +314,9 @@ public static void registerItems(RegistryEvent.Register<Item> event)
 
         allSimpleItems = simpleItems.build();
 
+        //Adding vanilla items ore dict
```
alcatrazEscapee: These can be moved to `OreDictionaryHelper#init`

src/main/java/net/dries007/tfc/objects/items/metal/ItemOreTFC.java
```diff
@@ -66,9 +67,18 @@ public ItemOreTFC(Ore ore)
             OreDictionaryHelper.register(this, "gem", ore);
             //noinspection ConstantConditions
             if (ore.getRegistryName().getPath().equals("lapis_lazuli"))
+            {
                 OreDictionaryHelper.register(this, "gem", "lapis");
+                OreDictionaryHelper.register(this, "ore", "lapis");
+            }
+            if (ore.getRegistryName().getPath().equals("cinnabar"))
+            {
+                OreDictionaryHelper.register(this, "ore", "cinnabar");
```
alcatrazEscapee: Is this actually the only ore that should use `ore` instead of `gem`?

src/main/java/net/dries007/tfc/objects/te/TEFirePit.java
```diff
@@ -322,7 +327,57 @@ private void handleInputMelting(ItemStack stack)
                 outputStack = inventory.insertItem(SLOT_OUTPUT_1, outputStack, false);
                 if (!outputStack.isEmpty())
                 {
-                    inventory.insertItem(SLOT_OUTPUT_2, outputStack, false);
+                    outputStack = inventory.insertItem(SLOT_OUTPUT_2, outputStack, false);
+                }
+                if (!outputStack.isEmpty()) // Couldn't merge directly
+                {
+                    // If both the output and input is the same food, try merging-updating the creation date to the earliest one
```
alcatrazEscapee: Two concerns:
 - the "merge two stacks that have different dates" should become a helper method in `CapabilityFood`. It's already used in the crafting recipe for food merging.
 - Do not try and mix traits / force stacks to stack if their traits don't match. The only thing that should be forcibly stacked is creation date

src/main/java/net/dries007/tfc/objects/te/TEFirePit.java
```diff
@@ -322,7 +327,57 @@ private void handleInputMelting(ItemStack stack)
...
+                                slot2Stack.setCount(merge);
+                            }
+                        }
+                    }
+                    if (!outputStack.isEmpty())
+                    {
+                        // If we got here, we failed merging foods, or this wasn't a food at all
+                        // Spit out the item in a random position. This is gonna avoid the food catching fire in the fire pit.
+                        // Unless this fire pit is placed near walls.
+                        Helpers.spawnItemStack(world, pos.offset(EnumFacing.byHorizontalIndex(Constants.RNG.nextInt(4))), outputStack);
```
alcatrazEscapee: Instead of throwing it in the world, just replace the result in the input slot. The result at the moment is essentially the same as current, just with a random "did you build your fireplace an open location" chance. Not an intuitive or friendly mechanic.

src/main/java/net/dries007/tfc/objects/items/ItemFireClay.java
alcatrazEscapee: Never mind, I realized its to make it ore dict triggered

src/main/java/net/dries007/tfc/CommonEventHandler.java
DisasterMoo: Unfortunately, it is. I've added it there since both are called at the same time if you are looking at a block (ie: looking at a chest while right clicking with a clay in hand tried to open both GUIs).

src/main/java/net/dries007/tfc/ConfigTFC.java
DisasterMoo: I tho this enhancement was because it was asked for a config so other mods can take over if modpacker wants to? see https://github.com/TerraFirmaCraft/TerraFirmaCraft/issues/436

src/main/java/net/dries007/tfc/CommonEventHandler.java
alcatrazEscapee: It's not. There are two seperate call paths - `processRightClickBlock`, and `processRightClick`. One for each event. As in the javadoc:
```
Note that this is NOT fired if the player is targeting a block {@link RightClickBlock} or entity {@link EntityInteract} {@link EntityInteractSpecific}.
```

src/main/java/net/dries007/tfc/ConfigTFC.java
alcatrazEscapee: Yes it was. I disagree with the sentiment, for the reasons I mentioned above. Just because somebody asks for it doesn't mean it should get implemented.

src/main/java/net/dries007/tfc/objects/items/metal/ItemOreTFC.java
DisasterMoo: It's not instead, but additionally. There may be others idk about it, which would need to investigate other mods for this.

src/main/java/net/dries007/tfc/CommonEventHandler.java
DisasterMoo: Well, in my testing it said otherwise. I've added that because it happened. Unless something is wrong in my env, that's why I need you to test it to be sure.

src/main/java/net/dries007/tfc/objects/te/TEFirePit.java
DisasterMoo: Came across a design problem here: While it is possible to just output to the input slot when the recipe is 1 item input -> 1 item output (and both have the same max stack), if, for example, we have 1 item output -> 2 item outputs, how can I put it into the input slot? The way it is implemented currently `handleInputMelting` works by converting the `cachedRecipe` input->output one at a time.

Another suggestion for this i could think of is working exactly like barrels, storing a queue of items if both slots are full.

src/main/java/net/dries007/tfc/ConfigTFC.java
DisasterMoo: Removed commit and changes

DisasterMoo:
Removed conflicting commits

src/main/java/net/dries007/tfc/objects/items/metal/ItemOreTFC.java
DisasterMoo: After a good inspection, i changed default minerals to include `ore` tag by default and the `gem` ones to also have `gemMineral` (ie: `gemKimberlite`)

Petrified wood tho is a confusing one. Is it an ore? The changes will made it to be called `orePetrifiedWood`, if that is right.

PR #486, opened 2019-10-10
Another bug fixing PR

DisasterMoo:
More details in commits.

### DIFF
--- a/generateResources.py
+++ b/generateResources.py
@@ -118,14 +118,14 @@ ORE_TYPES = {
 POWDERS = [
     'flux',
     'coke',
-    'kaolinite_powder',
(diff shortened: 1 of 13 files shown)


alcatrazEscapee:
I've cherry picked two of these fixes. As discussed in discord, the fire starting handling needs to be a bit more exhaustive. A handler for items, ore dict values, somewhere separated (maybe in `Helpers`?) would be good, for the charcoal forge, fire pit, and pit kiln. Also checking for fire above is good, but doesn't cover all cases.

As fr the material check, I have a bit of a different idea I want to try out first, so I'm holding off on that for now.

PR #496, opened 2019-10-15
Small improvements

DisasterMoo:
- Improved performance over small ore generation
- Added a function in `VeinRegistry` so addons can register veins without overriding config file.

### DIFF
--- a/src/main/java/net/dries007/tfc/world/classic/worldgen/WorldGenLooseRocks.java
+++ b/src/main/java/net/dries007/tfc/world/classic/worldgen/WorldGenLooseRocks.java
@@ -25,7 +25,6 @@ import net.dries007.tfc.objects.items.rock.ItemRock;
 import net.dries007.tfc.objects.te.TEPlacedItemFlat;
 import net.dries007.tfc.util.Helpers;
 import net.dries007.tfc.world.classic.ChunkGenTFC;
-import net.dries007.tfc.world.classic.WorldTypeTFC;
 import net.dries007.tfc.world.classic.chunkdata.ChunkDataTFC;
 import net.dries007.tfc.world.classic.worldgen.vein.Vein;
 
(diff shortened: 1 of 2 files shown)


alcatrazEscapee:
No. Addons should not force enable veins via code. That defeats the entire purpose of making it json configurable. They should provide resources the same way, i.e. by manually copying a default json file. The required change here would be to make TFC search for all files in the default folder, not just the single specified one.

DisasterMoo:
My intention is exactly that, not forcing via code but adding ores via json without overriding the base TFC one.

alcatrazEscapee:
This is not a good way to expose that hook to addons. Instead, if this is necessary to avoid overwriting TFC's default ore gen, TFC should just scan for all files in the `/config/tfc/` folder, and the addon should just copy their files there.

There's no need to duplicate json reading from multiple mods. And there's no reason to expose a hook that can be used to force veins to exist. For the same reason that forge doesn't expose direct hooks to register models, blockstates, or recipes.

DisasterMoo:
> TFC should just scan for all files in the /config/tfc/ folder
But that's exactly what i just did in the last commit.

alcatrazEscapee:
Ah, I didn't see that last change before I commented, sorry.

aldo1028:
any chance of this getting added soon Alc?

src/main/java/net/dries007/tfc/world/classic/worldgen/WorldGenLooseRocks.java
```diff
@@ -68,20 +68,9 @@ public void generate(Random random, int chunkX, int chunkZ, World world, IChunkG
                 {
                     veins.removeIf(v -> {
                         if (!v.type.hasLooseRocks()) return true;
-
```
alcatrazEscapee: If we're going to remove the full scan check for ore blocks, we should introduce a better check for vein proximity:
 - In 1.7.10, it was pointed out to me that the maximum depth scanned was much smaller than we have here, it should probably be adjusted to match. We can do a check of the "center" of the vein, and use the vertical size + vein size modifier to get an approximate value for the highest ore block.
 - The horizontal detection is (at the moment) chunk restricted. Ideally, the veins will still use position / distances to check validity, and the chunk data ore only becomes a sanity check that any ores actually generated.

src/main/java/net/dries007/tfc/world/classic/worldgen/vein/VeinRegistry.java
```diff
@@ -61,53 +61,73 @@ public void preInit(File dir)
...
             try
             {
                 weightedVeinTypes.clear();
                 veinTypeRegistry.clear();
-                Map<String, VeinType> values = GSON.fromJson(worldGenData, new TypeToken<Map<String, VeinType>>() {}.getType());
-                values.forEach((name, veinType) -> {
-                    veinType.setRegistryName(name);
-                    veinTypeRegistry.put(name, veinType);
-                    weightedVeinTypes.add(veinType.weight, veinType);
+                worldGenData.forEach(data -> {
```
alcatrazEscapee: The try / catch block here should be nested as close to the GSON call as possible, so if one file has a syntax error, it doesn't block all files from processing.

DisasterMoo:
Since I feel this needs some screenshot samples:

image

image

Small cassiterite generated about 1 chunk after the ones with that vein.

src/main/java/net/dries007/tfc/world/classic/worldgen/WorldGenLooseRocks.java
DisasterMoo: To pin this down, i'm commenting the line i found in classic:
https://github.com/TerraFirmaCraft/TFC-1.7.10/blob/9979960a7c13c19e4e6d018aa774c8a707c85592/src/Common/com/bioxx/tfc/WorldGen/Generators/WorldGenLooseRocks.java#L61

PR #512, opened 2019-10-21
Make anvil and mold recipes consistent for sword blades and mace heads.

### LINKED ISSUE #511: Conflict between sword and mace metal content in anvil and mold recipes
Swords and maces are listed as containing 200 units of metal, yet only take 100 units in a ceramic mold. The sword recipe on the anvil is also only a single ingot. 

Since metal can be melted down in the forge, this creates metal out of nowhere.

This is a inherent conflict in TFC classic, as the recipes are inconsistent between anvil and mold. Since they can't be melted down, this isn't a problem there.

 - TFC Version: Build 86

agilmore2:
Decreases anvil recipe for mace to one ingot.
Changes metal contents of swords and maces to 100 units.
Fixes #511

Signed-off-by: Gaelmare <agilmore2@users.noreply.github.com>


aldo1028:
i dont think this is what they wanted, think they wanted a check when melting to see if the item was worked on the anvil or not, if worked return 200 if not return 100

alcatrazEscapee:
Sword blades and maces should stay as double ingots. There's potential to investigate making the molds also require twice as much molten metal, but for sake of consistency I'd rather keep that the same.

However, in order to properly fix #511, the concept of recycling metal needs to be addressed. As it is something that's not inconceivable that would want to be added back in via addon, I'd like to make this configurable, with the caveat that we fix any potential duplication glitches via situations like this. In which case, I think the appropriate solution is to hardcode the melting amount of these two tools to be 100, similar to how blooms are handled. Then add a configuration option for disabling any recycling of materials that provide smithing bonuses.

agilmore2:
I think I understand hardcoding the melting mount by setting the smelt amount of the sword and mace Metal.ItemType to 100. I'm not sure how that's related to how blooms are handled?

Agreed that a config option to disable tool recycling is a good enhancement. I'll update the pull request. It's small, but good practice for me.

PR #527, opened 2019-11-05
Make anvil recipes consistent with 1.7 for sword blades. Fixes Dupe from #511.

agilmore2:
Reduce metal content of swords and maces to 100 units to avoid dupes.


agilmore2:
Alternative fix to #512, fixes #511 and matches Classic more closely, while avoiding metal dupe.

alcatrazEscapee:
24c746b53a0d87b62bb675390a34b3f4de8c86e6

Issue #414, opened 2019-09-11
Knapping GUI overwrites right click events


UnlimatedStone9:
**Describe the bug**
1. What did you expect to happen?
Be able to place clay (or any other knappable item) into a container using right click (examples being Tinkers' Construct drying racks, PyroTech Compacting Bins or Immersive Craft chests)

2. What actually happened instead (i.e. what was the bug)
The knapping GUI opened when it shouldn't

**To Reproduce**
Find a container that you can place items into using right click.
Attempt to place a knappable object into that container

**Meta Info**
 - TFC Version: 1.12.2-0.23.2.70

 - If necessary, what other mods (including versions) MUST be present to experience the bug?
Any mod that has a container you can place items into using right click rather than opening a GUI, examples being;
Tinkers' Construct's drying racks
PyroTech's compacting bins
Immersive Craft's chests, etc...

UnlimatedStone9:
From a little testing, it appears only clay balls have this issue but I could be wrong.

PR #530, opened 2019-11-09
Fix glass bottle interaction #471

### LINKED ISSUE #471: TFC alcohols and waters can be bottled into vanilla water.
**Describe the bug**
Glass bottle to not fill.
Vanilla glass bottles can be filled with liquid:water when clicking any of the TFC water types (limewater, hotwater, saltwater, freshwater), and any of the TFC alcohol types (rum, beer, whiskey, etc...).

**To Reproduce**

1. Click a fluid from TFC with a glass bottle.
2. Scream in horror!

**Meta Info**

 - TFC Version: 1.12.2-0.24.1.78

Lysergid:
Fixes #471

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/items/ItemOverriddenGlassBottle.java
+++ b/src/main/java/net/dries007/tfc/objects/items/ItemOverriddenGlassBottle.java
@@ -1,0 +1,89 @@
+/*
+ * Work under Copyright. Licensed under the EUPL.
+ * See the project README.md and LICENSE.txt for more information.
+ */
+
+package net.dries007.tfc.objects.items;
+
+import java.util.Optional;
+import javax.annotation.ParametersAreNonnullByDefault;
+
+import net.minecraft.block.material.Material;
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.init.Blocks;
+import net.minecraft.init.Items;
+import net.minecraft.item.ItemGlassBottle;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.ActionResult;
+import net.minecraft.util.EnumActionResult;
+import net.minecraft.util.EnumHand;
+import net.minecraft.util.math.RayTraceResult;
+import net.minecraft.world.World;
+import net.minecraftforge.fluids.FluidActionResult;
+import net.minecraftforge.fluids.FluidUtil;
+import net.minecraftforge.fluids.capability.IFluidHandler;
+import net.minecraftforge.fluids.capability.IFluidTankProperties;
+import net.minecraftforge.items.IItemHandler;
+
+import mcp.MethodsReturnNonnullByDefault;
+import net.dries007.tfc.util.Helpers;
+
+import static net.minecraftforge.items.CapabilityItemHandler.ITEM_HANDLER_CAPABILITY;
+
+@MethodsReturnNonnullByDefault
+@ParametersAreNonnullByDefault
+
+public class ItemOverriddenGlassBottle extends ItemGlassBottle
+{
+    public ItemOverriddenGlassBottle()
+    {
+        setTranslationKey("glassBottle");
+        setRegistryName(Items.GLASS_BOTTLE.getRegistryName());
+    }
+
+    @Override
+    public ActionResult<ItemStack> onItemRightClick(World worldIn, EntityPlayer playerIn, EnumHand handIn)
+    {
+        RayTraceResult rayTraceResult = Helpers.rayTrace(worldIn, playerIn, true);
+
+        if (rayTraceResult != null && isNotVanillaFluid(worldIn.getBlockState(rayTraceResult.getBlockPos())))
+        {
+            ItemStack heldItem = playerIn.getHeldItem(handIn);
+            FluidActionResult fluidActionResult = tryFill(playerIn, heldItem);
+            return fluidActionResult.isSuccess()
+                ? ActionResult.newResult(EnumActionResult.SUCCESS, fluidActionResult.getResult())
+                : ActionResult.newResult(EnumActionResult.FAIL, heldItem);
+        }
+        else
+        {
+            return super.onItemRightClick(worldIn, playerIn, handIn);
+        }
+    }
+
+    private static FluidActionResult tryFill(EntityPlayer playerIn, ItemStack heldItem)
+    {
+        IFluidHandler fluidHandler = FluidUtil.getFluidHandler(heldItem);
+        IItemHandler itemHandler = playerIn.getCapability(ITEM_HANDLER_CAPABILITY, null);
+        if (fluidHandler == null || itemHandler == null)
+        {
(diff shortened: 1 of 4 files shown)


src/main/java/net/dries007/tfc/util/interaction/InteractionManager.java
```diff
@@ -216,8 +233,11 @@ public static void onRightClickItem(PlayerInteractEvent.RightClickItem event)
             {
                 result = ServerInteractionManager.processRightClickItem(event, action);
             }
-            event.setCancellationResult(result);
-            event.setCanceled(true);
+            if (result != EnumActionResult.PASS)
+            {
+                event.setCancellationResult(result);
+                event.setCanceled(true);
+            }
```
alcatrazEscapee: This is intentional. Our right click handling exactly simulates the call / execution path for vanilla's right click, so we can replace the call to `Item#onRightClick` at the appropriate time. (see 93502c5c0cf070c0d874793d891d1afc313c4ac8). As such, we need to cancel the event afterwards, otherwise it would potentially call to the item / block twice. In addition, we register this event handler at the lowest priority, so if other mods intercept the event they will take precedence and not trigger our special logic.

src/main/java/net/dries007/tfc/util/interaction/InteractionManager.java
```diff
@@ -177,6 +181,19 @@
             return EnumActionResult.FAIL;
         });
 
+        RIGHT_CLICK_ACTIONS.put(stack -> Items.GLASS_BOTTLE.equals(stack.getItem()), (worldIn, playerIn, handIn) -> {
```
alcatrazEscapee:
```suggestion
        RIGHT_CLICK_ACTIONS.put(stack -> stack.getItem() == Items.GLASS_BOTTLE, (worldIn, playerIn, handIn) -> {
```

src/main/java/net/dries007/tfc/util/interaction/InteractionManager.java
```diff
@@ -177,6 +181,19 @@
...
+            if (traceResult == null)
+            {
+                return EnumActionResult.PASS;
+            }
+
+            Block rayTracedBlock = worldIn.getBlockState(traceResult.getBlockPos()).getBlock();
+            boolean isTfcFluid = FluidsTFC.getAllWrappers().stream()
+                .anyMatch(fluidWrapper -> fluidWrapper.get().getBlock().equals(rayTracedBlock));
+
+            return isTfcFluid ? EnumActionResult.FAIL : EnumActionResult.PASS;
```
alcatrazEscapee: The condition to ignore the action here should be if the fluid matches the check in `ItemGlassBottle` (which is a `material == Material.WATER` check, and the fluid is not equal to water. In addition, if that is the case, we want to try and fill the bottle using a fluid handler (i.e. call to `FluidUtil#tryFillContainerAnyStow`), in the event that something adds a proper fluid handler to the bottle. Otherwise, the action will fail.

If the fluid isn't `Material.WATER`, we need to manually call to `Item#onRightClick`, as it wouldn't get called otherwise.

Lysergid: I don't think that spreading calls to original logic across multiple components will serve well.
We already have code to call original logic in `ClientInteractionManager` and `ServerInteractionManager` for `RightClickBlock` event. 
I suggest to move `Item.onItemRightClick` to interaction managers. As of now there is no special handling for `Server/ClientInteractionManager.processRightClickItem` it simply delegates execution to IRightClickItemAction it will not call original logic on `EnumActionResult.PASS`.

alcatrazEscapee: The reason we don't just hook `RightClickBlock` event is because then we end up intercepting the right click event before `onBlockActivated` is called, leading to inconsistent placement effects. (For instance, see #414) This intercept at lowest priority, re-call all logic, then cancel the event regardless is done to *only* intercept `Item#onRightClick`

Not calling back to vanilla logic on pass is intentional - it has already been called (or supersceeded by an action) In the case with glass bottles, manually calling `ItemGlassBottle#onRightClick` is simply an event based workaround to overriding the method and calling super.

alcatrazEscapee:
As an alternative to using a event based intercept, we could just registry replace the glass bottle item with one that has a better check than `== Material.WATER`. There's potential for adding proper fluid handler logic to the item, but it is by no means a priority.

Lysergid:
Changed approach to item override since hook becomes too tricky plus it requires logic to replace ItemGlassBottle with ItemPotion in inventory

src/main/java/net/dries007/tfc/objects/items/ItemOverriddenGlassBottle.java
```diff
@@ -0,0 +1,89 @@
...
+
+import mcp.MethodsReturnNonnullByDefault;
+import net.dries007.tfc.util.Helpers;
+
+import static net.minecraftforge.items.CapabilityItemHandler.ITEM_HANDLER_CAPABILITY;
+
+@MethodsReturnNonnullByDefault
+@ParametersAreNonnullByDefault
+
+public class ItemOverriddenGlassBottle extends ItemGlassBottle
```
alcatrazEscapee: Keeping consistent with our other registry overrides, this should be called `ItemGlassBottleTFC`

src/main/java/net/dries007/tfc/objects/items/ItemOverriddenGlassBottle.java
```diff
@@ -0,0 +1,89 @@
...
+{
+    public ItemOverriddenGlassBottle()
+    {
+        setTranslationKey("glassBottle");
+        setRegistryName(Items.GLASS_BOTTLE.getRegistryName());
```
alcatrazEscapee: These should both be part of the registration call, i.e. `new ItemGlassBottleTFC().setRegistryName()...`

src/main/java/net/dries007/tfc/objects/items/ItemOverriddenGlassBottle.java
Lysergid: renamed

src/main/java/net/dries007/tfc/objects/items/ItemOverriddenGlassBottle.java
Lysergid: My 2c is - I think configuring item in ItemTFC harms readability. Having self-sufficient object upon construction is generally good practice. 
On the other hand I see you have established approaches, so I made change as you requested.

alcatrazEscapee: A couple reasons:
 - registry name and translation key need to be unique, so putting it in the constructor forces that class to only be used for exactly one item (not a problem with this case, but for many items this will apply)
 - All of vanilla item constructors do this, which is what this is copied from. This way it's easier to verify that we've not missed something if it looks the same.

PR #639, opened 2019-12-08
Add Lead recipe. Fixes #576

### LINKED ISSUE #576: Add replacement recipe for vanilla lead as replacement for Classic rope.
Jute has no use. As slimes may not spawn in TFC worlds(?), we should add a recipe for vanilla leads using jute fiber. What should we use instead of the slimeball?

Gaelmare:


alcatrazEscapee:
In future, this PR is completely unnecessary if you already made one to TFC-Assets

PR #641, opened 2019-12-09
Prevent calendar from advancing with no players on

rjb3977:
Previously the calendar would continue ticking even when no players were logged on to a server, which appears to have been unintentional.

This is immediately noticeable when starting a server as
`[tfc]: World time and calendar time are out of sync by -2 ticks, correcting!`
is spammed repeatedly.


alcatrazEscapee:
Nuts

PR #644, opened 2019-12-09
Changes in blast furnace and fuel manager

DisasterMoo:
More info on commits

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/items/metal/ItemOreTFC.java
+++ b/src/main/java/net/dries007/tfc/objects/items/metal/ItemOreTFC.java
@@ -77,8 +77,10 @@ public class ItemOreTFC extends ItemTFC implements IMetalItem
                     OreDictionaryHelper.register(this, "gem", "lapis");
                     break;
                 case "bituminous_coal":
-                case "lignite":
                     OreDictionaryHelper.register(this, "gem", "coal");
+                    break;
+                case "lignite":
+                    OreDictionaryHelper.register(this, "gem", "lignite");
                     break;
                 default:
                     OreDictionaryHelper.register(this, "gem", ore);
--- a/src/main/java/net/dries007/tfc/util/fuel/FuelManager.java
+++ b/src/main/java/net/dries007/tfc/util/fuel/FuelManager.java
@@ -9,20 +9,17 @@ import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nonnull;
 
-import net.minecraft.init.Items;
 import net.minecraft.item.ItemStack;
 
 import net.dries007.tfc.api.registries.TFCRegistries;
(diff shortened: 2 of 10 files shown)


src/main/java/net/dries007/tfc/objects/items/metal/ItemOreTFC.java
```diff
@@ -77,9 +77,11 @@ public ItemOreTFC(Ore ore)
                     OreDictionaryHelper.register(this, "gem", "lapis");
                     break;
                 case "bituminous_coal":
-                case "lignite":
                     OreDictionaryHelper.register(this, "gem", "coal");
                     break;
+                case "lignite":
+                    OreDictionaryHelper.register(this, "gem", "lignite");
+                    break;
```
alcatrazEscapee: No, this is intentional. We register ore type items based on what their product is. Using the ore doesn't suit what the purpose of the ore dictionary is.

src/main/java/net/dries007/tfc/util/fuel/FuelManager.java
```diff
@@ -41,34 +38,52 @@ public static boolean isItemForgeFuel(ItemStack stack)
...
         }
 
-        // Coal (Vanilla)
-        FUELS.add(new Fuel(new ItemStack(Items.COAL, 1, 0), 2200, 1350f, true));
-        // Coal (TFC Variants)
-        FUELS.add(new Fuel(ItemOreTFC.get(TFCRegistries.ORES.getValue(DefaultMetals.BITUMINOUS_COAL), 1), 2200, 1415f, true));
-        FUELS.add(new Fuel(ItemOreTFC.get(TFCRegistries.ORES.getValue(DefaultMetals.LIGNITE), 1), 2200, 1415f, true));
+        // Coals
+        FUELS.add(new Fuel(IIngredient.of("gemCoal"), 2200, 1415f, true, true));
+        FUELS.add(new Fuel(IIngredient.of("gemLignite"), 2000, 1350f, true, true));
```
alcatrazEscapee: Instead of using ore dictionary values, use the item instance here to differentiate. (Since they should both really be `gemCoal` for purposes of other mods)

Gaelmare:
Classic blast furnace would consume one charcoal per piece of ore melted on top of those charcoal used to heat furnace. Since the BF melts all ore at once, the delay used by Classic to allow time to add more fuel doesn't exist. This fix to setting fuel consumption much higher may be a good solution, then?

src/main/java/net/dries007/tfc/objects/items/metal/ItemOreTFC.java
alcatrazEscapee: After some more thought, this is probably the better way to handle this case.

PR #698, opened 2019-12-27
Some small bugfixes.

### LINKED ISSUE #663: Dough amounts are less rather than more than flour amounts when mixing.
**Describe the bug**
Mixing flour and water produces about half as much dough (in weight) rather than double. 

**To Reproduce**
Mill flour. Mix with water. Two flour (10 oz of flour) and water will produce 1 dough (5 oz of dough,) rather than the usual which is double (2 flour + water = 4 dough or 10 oz flour + 10 oz water = 20oz of dough.)

Uncertain if this is intended, but it's not listed as a difference between 1.7 and 1.12. 

TFC's wiki states, under Bucket:

"Dough
When combining flour with fresh water, the resulting dough will have double the weight. If the flour weighs more than 80 oz, crafting the dough will leave the flour in the crafting grid, minus the 80 oz used to create a max weight piece of dough."

**Meta Info**
 - TFC Version: MC1.12.2-0.27.6.106

### LINKED ISSUE #673: Swords aren't effective against cobwebs
Expected Behaviour: 
Swords cut cobweb efficiently, dropping string.

Behaviour:
Swords are not any faster than any other tool (or no tool), and cobwebs dont drop string when broken.

Reproduce:
1. Grab sword
2. Attempt to break web

**Meta Info**
 - TFC Version: 0.27.7.107

### LINKED ISSUE #676: cooked eggs recipe does not exist
- TFC Version: build 107

### LINKED ISSUE #678: Brine recipe is missing bucket of vinegar
**Describe the bug**
Can't make Brine, recipe looks to be missing the vinegar bucket.
brine

**Meta Info**
 - TFC Version: 0.X.X.107

### LINKED ISSUE #684: Pheasants take fall damage.
Pheasants take fall damage why they shouldn't..

They fly down like a chicken and instantly explode into items.

They shouldn't do that.

**Meta Info**
TFC: 107

### LINKED ISSUE #687: Fences are missing hardness.
Smacking a fence insta-breaks them.

TFC: 107.

### LINKED ISSUE #701: Can't make cheese
**Describe the bug**
Make sure to include:
Adding bucket of vinegar to barrel of milk to make curdled milk for cheese.
Nothing is made.

**Meta Info**
 - TFC Version: 0.X.X.107

Gaelmare:
Fix #687 by adding hardness and resistance matching vanilla
Fix #684 for TFC chickens and pheasants
Fix #676 by adding heat capacity to vanilla egg and a heating recipe for cooked eggs.
Fix #673 by adding a check for sword vs web to canHarvestBlock
Fix #663 by removing extra recipes, and changing output ratio of flour+water
Fix #678 by fixing brine recipe. Still consumes wooden bucket though.
Fix #701 by adding recipes for vinegar milk, curdled milk, and cheese. Adds empty input stack ability.

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
+++ b/src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
@@ -323,7 +323,6 @@ public final class ItemsTFC
             new ItemBlockTorch(Blocks.TORCH).setRegistryName("minecraft", "torch"),
             new ItemGlassBottleTFC().setRegistryName(Items.GLASS_BOTTLE.getRegistryName()).setTranslationKey("glassBottle")
         );
-        CapabilityItemHeat.CUSTOM_ITEMS.put(IIngredient.of(Items.EGG), () -> new ItemHeatHandler(null, 1, 480));
     }
 
     public static void init()
--- a/src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
+++ b/src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
@@ -7,6 +7,9 @@ package net.dries007.tfc.objects.items;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
+import net.dries007.tfc.api.capability.heat.CapabilityItemHeat;
+import net.dries007.tfc.api.capability.heat.ItemHeatHandler;
+import net.dries007.tfc.objects.inventory.ingredient.IIngredient;
 import net.minecraft.creativetab.CreativeTabs;
 import net.minecraft.init.Blocks;
 import net.minecraft.init.Items;
@@ -320,6 +323,7 @@ public final class ItemsTFC
             new ItemBlockTorch(Blocks.TORCH).setRegistryName("minecraft", "torch"),
             new ItemGlassBottleTFC().setRegistryName(Items.GLASS_BOTTLE.getRegistryName()).setTranslationKey("glassBottle")
         );
+        CapabilityItemHeat.CUSTOM_ITEMS.put(IIngredient.of(Items.EGG), () -> new ItemHeatHandler(null, 1, 480));
     }
 
     public static void init()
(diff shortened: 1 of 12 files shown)


src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -320,6 +323,7 @@ public static void registerVanillaOverrides(RegistryEvent.Register<Item> event)
             new ItemBlockTorch(Blocks.TORCH).setRegistryName("minecraft", "torch"),
             new ItemGlassBottleTFC().setRegistryName(Items.GLASS_BOTTLE.getRegistryName()).setTranslationKey("glassBottle")
         );
+        CapabilityItemHeat.CUSTOM_ITEMS.put(IIngredient.of(Items.EGG), () -> new ItemHeatHandler(null, 1, 480));
```
alcatrazEscapee: This should be in `CapabilityItemHeat#preInit`, not here.

src/main/java/net/dries007/tfc/objects/te/TEBarrel.java
```diff
@@ -259,8 +259,10 @@ public void update()
                         surplus.addAll(output);
                         IBlockState state = world.getBlockState(pos);
                         world.notifyBlockUpdate(pos, state, state, 3);
+                        onSealed(); //run the sealed check again in case we have a new valid recipe.
                     }
-                    recipe = null;
+                    else recipe = null;
```
alcatrazEscapee: Eww
```suggestion
                    else 
                    {
                        recipe = null;
                    }
                        
```

Gaelmare: This one didn't get caught by Code Format. Thanks.

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
Gaelmare: Thanks for providing correct location, wasn't sure at all.

PR #727, opened 2020-01-04
Implement Red/Blue Steel buckets

### LINKED ISSUE #650: Blue and Red Steel buckets are missing
These are iconic TFC items

Gaelmare:
Fixes #650.


Gaelmare:
Comments/criticism welcome

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -154,6 +158,8 @@ public static void registerItems(RegistryEvent.Register<Item> event)
         simpleItems.add(register(r, "wand", new ItemDebug(), CT_MISC));
         simpleItems.add(register(r, "mortar", new ItemMisc(Size.TINY, Weight.LIGHT, "mortar"), CT_MISC));
         register(r, "wooden_bucket", new ItemWoodenBucket(), CT_WOOD); //not a simple item, use a custom model
+        register(r, "blue_steel_bucket", new ItemMetalBucket(Metal.BLUE_STEEL, Metal.ItemType.BUCKET), CT_METAL); //not a simple item, use a custom model
+        register(r, "red_steel_bucket", new ItemMetalBucket(Metal.RED_STEEL, Metal.ItemType.BUCKET), CT_METAL); //not a simple item, use a custom model
```
alcatrazEscapee: Why are these using a completely separate method of registration than the existing metal check? If the bucket is going to be an `ItemType`, it shouldn't exclude itself from the standard pattern of `metal/[type]/[metal]` registry names

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -187,7 +193,7 @@ public static void registerItems(RegistryEvent.Register<Item> event)
         {
             for (Metal metal : TFCRegistries.METALS.getValuesCollection())
             {
-                if (type.hasType(metal))
+                if (type != Metal.ItemType.BUCKET && type.hasType(metal))
```
alcatrazEscapee:
```suggestion
                if (type.hasType(metal))
```
The modification to `Metal.ItemType#hasType` should already cover this

src/main/java/net/dries007/tfc/objects/items/metal/ItemMetalBucket.java
```diff
@@ -0,0 +1,190 @@
...
+                                catch (IllegalArgumentException e) { /* Just skip */ }
+                            }
+                        }
+                        worldIn.playSound(null, pos, SoundEvents.ITEM_BUCKET_EMPTY, SoundCategory.PLAYERS, 1.0F, 1.0F);
+                        return ActionResult.newResult(EnumActionResult.SUCCESS, stack);
+                    }
+                }
+            }
+        }
+        return ActionResult.newResult(EnumActionResult.PASS, stack);
```
alcatrazEscapee: Prefer just calling constructors (there's a couple other places where this can be changed as well)
```suggestion
        return new ActionResult<>(EnumActionResult.PASS, stack);
```

src/main/java/net/dries007/tfc/objects/items/metal/ItemMetalBucket.java
```diff
@@ -0,0 +1,190 @@
...
+        if (metal.equals(BLUE_STEEL)) {
+            Set<Fluid> blueFluids = FluidsTFC.getAllMetalFluids().stream().map(FluidWrapper::get).collect(Collectors.toSet());
+            blueFluids.add(LAVA);
+            return blueFluids;
+        } else if (metal.equals(RED_STEEL)) {
+            return FluidsTFC.getAllWrappers().stream().filter(x -> x.get(DrinkableProperty.DRINKABLE) != null).map(FluidWrapper::get).collect(Collectors.toSet());
+        }
+        else //noinspection unchecked
+        return (Set<Fluid>) Collections.EMPTY_SET; // No other metal buckets implemented
+    }
```
alcatrazEscapee: This by design is hardcoded and not extensible. It should implement a similar condition to how the ceramic jug checks for fluids by using a config-supplied list of fluid names. Two config entries, one for blue steel, one for red steel. The current code is fine for default values.

Gaelmare: The ceramic jug uses exactly this same code as the red steel bucket, all drinkable fluids. The wooden bucket uses a manually configured fluid list. Which makes sense? Classic would have just fresh water for the red steel bucket, and just Lava for blue steel bucket. Do we want to leave these open as currently coded, make them manually configured lists, or mimic the classic behavior of just fresh water/lava?

src/main/java/net/dries007/tfc/objects/items/metal/ItemMetalBucket.java
Gaelmare: This mimics code in ItemWoodenBucket. Happy to change in both places, if that's what we want?

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
Gaelmare: They're registering separate models because they have empty and fluid filled versions, and the simple registry of the rest of the Metal ItemTypes will not provide the model registration. We still want the metal information as well as the rest of the ItemMetal super class stuff, unless we want to reimplement it because the buckets are a special snowflake? Do we want to make a super class for buckets, and implement wooden and metal versions of them?

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
Gaelmare: The modification of hasType ensures that no other metals are valid buckets (wrought_iron, bronzes, etc.). This modification ensures that buckets are not registered via the simple registry to handle their custom filled fluid models.

Thanks for feedback!

alcatrazEscapee: I see the problem. This makes sense then.

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
alcatrazEscapee: I see the reason for separating them from the bulk model registration. It should still use the established naming convention though.

src/main/java/net/dries007/tfc/objects/items/metal/ItemMetalBucket.java
alcatrazEscapee: Yes

src/main/java/net/dries007/tfc/objects/items/metal/ItemMetalBucket.java
alcatrazEscapee: Whoops, I meant to say wooden bucket. Yes, they should both be config based fluid lists. Default to lava / water only.

Gaelmare:
This gets into the bit of forge magic I don't understand. These use
blockstate descriptions of the model, and not not the regular model
directories at all. I'll experiment with moving the blockstate descriptions
to the right directory and changing the registration path. May work?

Gaelmare:
So in summary, I need to
Update wooden and metal buckets to return new ActionResults
Change the item naming to match standard metal items
Add config based fluid lists, defaulting to the 3 water types or lava.

Thanks!

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
Gaelmare: Completed.

src/main/java/net/dries007/tfc/objects/items/metal/ItemMetalBucket.java
Gaelmare: Done.

PR #731, opened 2020-01-09
Some more bug fixes

DisasterMoo:


src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockFruitTreeTrunk.java
```diff
@@ -261,6 +285,17 @@ public void randomTick(World worldIn, BlockPos pos, IBlockState state, Random ra
...
 
+    @Override
+    @SuppressWarnings("deprecation")
+    public void neighborChanged(IBlockState state, World worldIn, BlockPos pos, Block blockIn, BlockPos fromPos)
+    {
+        super.neighborChanged(state, worldIn, pos, blockIn, fromPos);
+        if (!BlocksTFC.isGrowableSoil(worldIn.getBlockState(pos.down())))
+        {
+            worldIn.destroyBlock(pos, false);
+        }
```
alcatrazEscapee: Does this not break any trunks that are > 1 high, rendering fruit trees impossible to grow?

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockFruitTreeBranch.java
```diff
@@ -39,9 +40,29 @@
...
-    private static final AxisAlignedBB HORIZONTAL_X_AABB = new AxisAlignedBB(0.0D, 0.375D, 0.375D, 1.0D, 0.625, 0.625);
+    /* Connection sides
+     * 0 = no connection
+     * 1 = connected, use vertical model
+     * 2 = connected=, use horizontal model */
+    public static final PropertyInteger NORTH = PropertyInteger.create("n", 0, 2);
+    public static final PropertyInteger EAST = PropertyInteger.create("e", 0, 2);
+    public static final PropertyInteger SOUTH = PropertyInteger.create("s", 0, 2);
+    public static final PropertyInteger WEST = PropertyInteger.create("w", 0, 2);
+    public static final PropertyInteger UP = PropertyInteger.create("u", 0, 2);
```
alcatrazEscapee: These should use actual names, i.e. "north", "south", "up", for consistency with vanilla, and general readability.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockFruitTreeTrunk.java
DisasterMoo: you're right! I should've added another condition for trunks!

alcatrazEscapee: This is still broken - fruit tree branches currently break on every block update.

Issue #728, opened 2020-01-04
Typo in the spindle recipe name


Mrthomas20121:
![image](https://user-images.githubusercontent.com/34064092/71764365-45b47980-2ee7-11ea-914d-f2bbc47256a3.png)
"recipe ID": tfc:spindel
TFC version : 0.27.7.107

Issue #720, opened 2020-01-02
Rock Knife Blade Large Item instead of Small


KaneHart:
**Describe the bug**
Rock Knife Blade shows up as Large instead of Small and can't be stored correctly.
![image](https://user-images.githubusercontent.com/425853/71652828-86996c00-2cf6-11ea-88a6-5611e7908d3a.png)
Rock Knife Blade should be set to small. 

**Meta Info**
 - TFC Version: 0.27.7.107

Gaelmare:
When we knapp something out of a tiny, light rock, it should also be tiny and light.

Issue #591, opened 2019-11-27
Ceramic jugs delete liquids it can't be filled with


rhapsodyman2000:
Empty ceramic jugs delete liquid blocks when you attempt to fetch unassigned liquids, such as lava or hot water. The ceramic jug is left empty.
TerraFirmaCraft version 0.26.6.97

alcatrazEscapee:
Seems the issue is that any fluid not pick-up-able by ceramic jugs still get removed from the world but don't fill.

rhapsodyman2000:
I believe that there is no ID for Jugs filled with Hot water.

alcatrazEscapee:
Different fluids don't have a different ID. There's a list in the config with the valid fluids pickupable with the ceramic bucket. Try adding hot water to that list and seeing if the jug can pick that up.

rhapsodyman2000:
Wouldn’t know how too 😫. Besides, there is no texture for it that I can see.

alcatrazEscapee:
> Besides, there is no texture for it that I can see.

The texture uses the forge bucket blockstate, so this is meaningless. Fluid textures are made using an overlay at runtime. The issue here is just that the ceramic jug will allow picking up fluids despite not actually filling the fluid container.

Issue #561, opened 2019-11-20
cobblestone no longer fits in small vessels


agilmore2:
In classic, cobblestone was heavy/very small, and fit inside small vessels:
![2019-11-19_21 43 34](https://user-images.githubusercontent.com/6852298/69213460-66846380-0b21-11ea-807c-87da78a9f850.png)

In TNG, cobblestone does not fit in small vessels, and is also heavy/very small.

**Meta Info**
 - TFC Version: build 94

agilmore2:
![2019-11-19_23 11 22](https://user-images.githubusercontent.com/6852298/69213679-03df9780-0b22-11ea-8f8a-ff5120aa1c75.png)

alcatrazEscapee:
Vessels are limited by size: < Large and < Heavy. We currently default to very small / heavy for `ItemBlock`s. To keep the same stack size, this would need to change to Small / Normal, here:
https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/40d0ac3709e8ea0c24a2285060e8bf310b88c164/src/main/java/net/dries007/tfc/api/capability/size/CapabilityItemSize.java#L129

Any anywhere else that value is hardcoded. Would also need to investigate possible side effects caused by this change (likely none, but possible interaction with other size limited containers)

Gaelmare:
It does seem strange to me that 4 Light/Tiny rocks can make a Heavy/V.Small cobble block, so Small/Normal makes sense for cobble blocks and maybe raw blocks.

Gaelmare:
So this was broken in #519 to make the size annotations match Classic. I think the right answer is Medium/Small to match the 32 stack size of Classic while also allowing carrying them in small vessels.

PR #732, opened 2020-01-10
Bugfixes

Gaelmare:
More simple bugfixes.


Gaelmare:
More detail, sorry.
Fix #728 by moving the recipe to the right filename, spindle.json
Fix #720 by changing the size of toolheads to the same size of the rocks they were knapped from
Fix #561 by changing block item size values to match discussed solution
Fix #591 by ensuring the container can hold the fluid contained in the source

alcatrazEscapee:
f9b5d26553083519723bc39dc4ed8f1388ef2f0e

PR #733, opened 2020-01-10
Add Chisel Behavior

DavidTriphon:
Some things still need to be sorted out:
- A symbol to signify that a hammer is missing from the required slot.
- Slabs jump around when they are dropped in the world when slabs are carved.
- Possible enhancement for creative to remove dropping slabs when the player is in creative.


DavidTriphon:
replaced by PR #734

PR #744, opened 2020-01-15
And more bug fixes!

DisasterMoo:
See commits for more details

### DIFF
--- a/src/main/java/net/dries007/tfc/CommonEventHandler.java
+++ b/src/main/java/net/dries007/tfc/CommonEventHandler.java
@@ -117,16 +117,31 @@ public final class CommonEventHandler
     }
 
     /**
-     * Make leaves drop sticks
+     * Update harvesting tool before it takes damage
      */
+    @SubscribeEvent
+    public static void breakEvent(BlockEvent.BreakEvent event)
+    {
+        final EntityPlayer player = event.getPlayer();
+        if (player != null)
+        {
+            IPlayerData cap = player.getCapability(CapabilityPlayerData.CAPABILITY, null);
+            if (cap != null)
+            {
+                cap.setHarvestingTool(player.getHeldItemMainhand());
+            }
+        }
+    }
+
     @SubscribeEvent
     public static void onBlockHarvestDrops(BlockEvent.HarvestDropsEvent event)
     {
-        final EntityPlayer harvester = event.getHarvester();
-        final ItemStack heldItem = harvester == null ? ItemStack.EMPTY : harvester.getHeldItemMainhand();
+        final EntityPlayer player = event.getHarvester();
+        final ItemStack heldItem = player == null ? ItemStack.EMPTY : player.getHeldItemMainhand();
         final IBlockState state = event.getState();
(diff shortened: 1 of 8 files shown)


src/main/java/net/dries007/tfc/objects/container/ContainerKnapping.java
```diff
@@ -130,6 +130,30 @@ private void resetMatrix()
...
+     * Used in client to set a slot state in the matrix
+     * JEI won't cause issues anymore see https://github.com/TerraFirmaCraft/TerraFirmaCraft/issues/718
+     *
+     * @param index the slot index
+     * @param value the value you wish to set the state to
+     */
+    public void setSlotState(int index, boolean value)
+    {
+        matrix.set(index, value);
+    }
```
alcatrazEscapee: Nowhere do these sync the server matrix state to the client, is that intentional? i.e. the client is tracking it's matrix completely independently of the server.

DisasterMoo: yes, that's intentional. We don't need to sync the server matrix->client because that would be redundant/cause lag issues. While recipe logic is done on server, the client is the one who draws the matrix (sending packets to server on which slot he clicked).

The lag issue i mentioned would be like block breaking lag: Client click on slot, slot is sent to server, server send to client a resync packet while client already clicked at another slot, the slot is visibly returning to true state before going false again.

src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -140,6 +156,26 @@ public static void onBlockHarvestDrops(BlockEvent.HarvestDropsEvent event)
...
+        // Apply durability modifier on tools
+        if (player != null)
+        {
+            ItemStack tool = ItemStack.EMPTY;
+            IPlayerData cap = player.getCapability(CapabilityPlayerData.CAPABILITY, null);
+            if (cap != null)
+            {
+                tool = cap.getHarvestingTool();
+            }
+            if (tool != ItemStack.EMPTY)
```
alcatrazEscapee: Don't equality check against `ItemStack.EMPTY`, always use `.isEmpty()`

src/main/java/net/dries007/tfc/objects/blocks/wood/BlockLogTFC.java
```diff
@@ -152,10 +155,20 @@ public boolean removedByPlayer(IBlockState state, World world, BlockPos pos, Ent
...
         if (!state.getValue(PLACED))
         {
-            final ItemStack stack = player.getHeldItemMainhand();
+            ItemStack stack = ItemStack.EMPTY;
+            IPlayerData cap = player.getCapability(CapabilityPlayerData.CAPABILITY, null);
+            if (cap != null)
+            {
+                stack = cap.getHarvestingTool();
+            }
+            if (stack == ItemStack.EMPTY)
```
alcatrazEscapee: Same as above, use `.isEmpty()`

Issue #652, opened 2019-12-10
Planting seed then digging it up sometimes gives more than one seed


p0kes:
Planting seed then digging it up sometimes gives more than one seed

(Agriculture master skill)

**Meta Info**
 - TFC Version: 0.X.X.105 on test server

DavidTriphon:
If crops die by placing their seed on the ground, doesn't it stand to reason that killing a plant sooner by hand should also drop exactly 1 seed? Maybe a better way would be to only drop extra seeds when the plant also drops food, but still drop exactly 1 seed up until that point.

DavidTriphon:
Gaelmare pointed out to me that this issue wasn't closed. This has been fixed by the crop fixes PR.

Issue #741, opened 2020-01-13
Crops should have different temperature ranges


DavidTriphon:
**Describe the request**
1. What problem or issue would this enhancement solve?
TFC Classic had different temperatures that each plant required in order to stay alive or grow. Currently all TNG crops require exactly 5-35C.

2. Describe the feature itself - what would change, what would happen. In general more detail is good!
Change the TNG crop temperature values to more closely match those in classic. Classic only cared about minimum ranges, but TNG doesn't make a difference between alive and growing, so there's some leanway to be done.

**Meta Info**
 - TFC Version: (at time of request)
0.28.0.110

 - Is this a compatibility feature for other mods? Or a enhancement for TFC alone?
This is an enhancement for TFC alone.

 - If necessary, what other mods (including versions) would this require?
none

alcatrazEscapee:
> TNG doesn't make a difference between alive and growing

Yes, it does, or at least should. See <https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/838dd798de233841987cf2f00aabaaf53b0eb648/src/main/java/net/dries007/tfc/api/types/ICrop.java#L33> vs <https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/838dd798de233841987cf2f00aabaaf53b0eb648/src/main/java/net/dries007/tfc/api/types/ICrop.java#L42>

DavidTriphon:
Somehow I missed that. But it's unfortunate that the alive parameters aren't explicit. I think that would be an improvement to allow crops to not die with a greater tolerance than 5 degrees.

alcatrazEscapee:
#749

PR #749, opened 2020-01-17
Lots of Crop Fixes

DavidTriphon:
Gave crops unique temperature ranges.
Fixed crops growing rates to grow at a rate that could be witnessed within a single lifetime.
Fixed crops not growing as much as they should when reloaded after an entire year.
Fixed spreading crops not dying and dropping their seeds properly.
Balanced crops to only drops extra seeds from skill when the crop drops food.


DavidTriphon:
Fixes #652 and #741

DavidTriphon:
You asked me to merge timeScaleCrops into this PR but I realized that it's not entirely done because I need to overhaul how they track age in order to tick them solely from TEs for ICalendarTickable.

DavidTriphon:
So it's going to be a while longer before I get TE's working so that crops can be updated on chunk load. The current state of the PR works well, so if you like it Alcatraz I have no objections to getting it merged. I'll do the `onCalendarUpdate()` behavior in a future PR that includes other things like dead crops probably.

DavidTriphon:
I've got dead crop blocks ready to add to this PR but I need https://github.com/TerraFirmaCraft/TFC-Assets/pull/136 to be accepted first or else I'd be adding purple and black blocks to the game.

DavidTriphon:
You know what, to save time I'll just add it now.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropDead.java
```diff
@@ -0,0 +1,127 @@
...
+import net.minecraft.world.World;
+import net.minecraftforge.common.EnumPlantType;
+
+import net.dries007.tfc.api.types.ICrop;
+import net.dries007.tfc.objects.items.ItemSeedsTFC;
+
+public class BlockCropDead extends BlockBush
+{
+    /* true if the crop spawned in the wild, means it ignores growth conditions i.e. farmland */
+    public static final PropertyBool WILD = PropertyBool.create("wild");
```
alcatrazEscapee: Why do dead crops need a wild condition? They shouldn't have growth checks, nor should they particularly care i.e. if the soil beneath them is not farmland, as maybe that's exactly what killed them in the first place.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropDead.java
```diff
@@ -0,0 +1,127 @@
...
+    {
+        return MAP.get(crop);
+    }
+
+    public static Set<ICrop> getCrops()
+    {
+        return MAP.keySet();
+    }
+
+    // fields
```
alcatrazEscapee: These are unnecessary comments. Our code style already sorts various class members so there's no need to delimit sections, and the members themselves tell you what they are. i.e. a constructor is obviously not a field.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropSimple.java
```diff
@@ -101,52 +64,47 @@ public boolean onBlockActivated(World worldIn, BlockPos pos, IBlockState state,
...
+                };
+            case 8:
+                return new BlockCropSimple(crop, isPickable)
+                {
+                    @Override
+                    public PropertyInteger getStageProperty()
+                    {
+                        return STAGE_8;
+                    }
+                };
```
alcatrazEscapee: Instead of a switch over the property for each different crop type (imo ugly, long, duplicated code), consider seperating these into two methods:
```java
public static void getStagePropertyForValue(int maxStages) {
    // return the stage property based on maxStages, perhaps even using an array index
}
public static BlockCropSimple create(int maxStages) {
    PropertyInteger property = getStagePropertyForValue(maxStages);
    return new BlockCropSimple(...) {
        @Override public PropertyInteger getStageProperty() { return property; }
    }
}
```

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropSpreading.java
```diff
@@ -6,44 +6,34 @@
 package net.dries007.tfc.objects.blocks.agriculture;
```
alcatrazEscapee: What's the plan for spreading crops? I thought the decision was to remove / replace them for the time being as their growth method heavily relies on randomticks?

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropTFC.java
```diff
@@ -57,8 +86,12 @@ public static BlockCropTFC get(ICrop crop)
         return MAP.keySet();
     }
 
+    // fields
+
```
alcatrazEscapee: Eww

src/main/java/net/dries007/tfc/objects/te/TECropBase.java
```diff
@@ -0,0 +1,69 @@
...
+@ParametersAreNonnullByDefault
+public class TECropBase extends TETickCounter implements ICalendarTickable
+{
+    // static constants
+
+    private static final String NBT_LAST_TICK_CAL_CHECKED = "lastTickCalChecked";
+
+    // fields
+
+    protected long _lastTickCalChecked;
```
alcatrazEscapee:
```suggestion
    protected long lastTickCalChecked;
```

src/main/java/net/dries007/tfc/objects/te/TECropBase.java
```diff
@@ -0,0 +1,69 @@
...
+import net.minecraft.nbt.NBTTagCompound;
+
+import net.dries007.tfc.objects.blocks.agriculture.BlockCropTFC;
+import net.dries007.tfc.util.calendar.CalendarTFC;
+import net.dries007.tfc.util.calendar.ICalendarTickable;
+
+@ParametersAreNonnullByDefault
+public class TECropBase extends TETickCounter implements ICalendarTickable
+{
+    // static constants
```
alcatrazEscapee: More Ew

src/main/java/net/dries007/tfc/util/agriculture/Crop.java
```diff
@@ -25,58 +25,116 @@
...
-    JUTE(() -> new ItemStack(ItemsTFC.JUTE), () -> ItemStack.EMPTY, 5f, 35f, 100f, 400f, 6, 0.5f, SIMPLE),
-    PUMPKIN(() -> new ItemStack(Blocks.PUMPKIN), () -> ItemStack.EMPTY, 5f, 35f, 100f, 400f, 8, 0.5f, SPREADING),
-    MELON(() -> new ItemStack(Blocks.MELON_BLOCK), () -> ItemStack.EMPTY, 5f, 35f, 100f, 400f, 8, 0.5f, SPREADING);
-
-    public static final PropertyInteger STAGE_8 = PropertyInteger.create("stage", 0, 7);
-    public static final PropertyInteger STAGE_7 = PropertyInteger.create("stage", 0, 6);
-    public static final PropertyInteger STAGE_6 = PropertyInteger.create("stage", 0, 5);
-    public static final PropertyInteger STAGE_5 = PropertyInteger.create("stage", 0, 4);
+    // todo: unique rain tolerances for each crop
+    BARLEY(Food.BARLEY,
```
alcatrazEscapee: Everything here is automatically generated via spreadsheet. This means:
 - No extra newlines between arguments, it's ugly
 - Any values changed here should be changed in the spreadsheet first, as it's much easier to peruse that for balance changes than raw source code.

src/main/java/net/dries007/tfc/util/agriculture/Crop.java
```diff
@@ -123,59 +181,44 @@ else if (currentStage == getMaxStage() - 1)
...
+        }
+
+        public Range(float minAlive, float minGrow, float maxGrow, float maxAlive)
+        {
+            this.minAlive = minAlive;
+            this.minGrow = minGrow;
+            this.maxGrow = maxGrow;
+            this.maxAlive = maxAlive;
+        }
+    }
```
alcatrazEscapee: This is rather pointless abstraction, it's not used anywhere, and the entire point seems to avoid a calculation which could just as easily be done in constructor arguments for `Crop` instead.

DavidTriphon:
> What's the scope of remaining changes? i.e. things still to be considered:

You can find my list of tasks that I've remembered to track on my project page here: https://github.com/DavidTriphon/TerraFirmaCraft/projects/1

>     what's happening with spreading crops (and by extension, pumpkins and melons?)

In this patch currently there is a temporary change turn melons and pumpkins and etc into simple crops since patching the Spreading class behavior will require some redesigning. I wanted to try to introduce different crop types in a different patch after discussion with you and Clay.

>     what are conditions for crops turning into dead versions, and what are conditions for dead crops getting destroyed (i.e. obviously block beneath them getting removed, but others?)

Currently the death conditions are exactly the same. Anything that set them to an item placed on the ground before now turns it into a dead crop.

>     are ICalendarTickable changes finalized or is it still WIP?

IcalanderTickable should be working correctly. I don't think anything needs to change on that anymore.

>     are crop growth rates randomized or are they all exactly the same?

growth rates are not randomized. randomTicks are used to checkGrowth() and update the blockstate at random times, but if they were all planted at the same time and randomTicked at the same time, then they would update their growths at the same time.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropDead.java
DavidTriphon: I've removed all the comments I've found and reordered functions according to the reformat code style.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropSimple.java
DavidTriphon: I've removed the switch and replaced it with a method like you've suggested that uses the stage mapping.

src/main/java/net/dries007/tfc/objects/te/TECropBase.java
DavidTriphon: I've removed all the underscores for fields.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropSpreading.java
DavidTriphon: They are defunct. This code is basically useless. I left it in for the time being until something is done with it.

src/main/java/net/dries007/tfc/util/agriculture/Crop.java
DavidTriphon: I thought it would be useful for making the constructors more clear and readable, rather than a series of a bunch of numbers in the arguments on all one line, which is also why I opted to add newlines at designated points in the code, helps readability...

alcatrazEscapee: These constructors are auto-generated - there's no use in making them more readable as it's much easier to look and modify a spreadsheet instead.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropDead.java
DavidTriphon: WILD condition has been removed.

src/main/java/net/dries007/tfc/util/agriculture/Crop.java
DavidTriphon: values have been changed in the spreadsheet and the code now represents those changes.

src/main/java/net/dries007/tfc/util/agriculture/Crop.java
DavidTriphon: Range has been removed.

generateResources.py
```diff
@@ -980,6 +982,31 @@ def item(filename_parts, *layers, parent='item/generated'):
 # LEATHER / HIDES
 blockstate(('placed_hide',), 'tfc:hide_rack', {})
 
+# AGRICULTURE
+
+for crop in CROPS.keys():
```
alcatrazEscapee:
```suggestion
for crop, data in CROPS.items():
```
Rather than accessing each property by value

generateResources.py
```diff
@@ -980,6 +982,31 @@ def item(filename_parts, *layers, parent='item/generated'):
...
+    texture = CROPS[crop]['texture']
+    mymodel = CROPS[crop]['model']
+
+    variants = {'stage': {}}
+    for stage in range(stages):
+        variants['stage'][stage] = {
+            'textures': {
+                texture: "tfc:blocks/crop/%s_%d" % (crop, stage)
+            }
+        }
```
alcatrazEscapee: This can be reduced to
```python
{'stage': dict((str(i), { 'textures': { 'texture': 'tfc:blocks/crop%s_%d' % (crop, stage)}) for i in range(stages)}
```
and since it's only used once, just inline it in the method call

generateResources.py
```diff
@@ -980,6 +982,31 @@ def item(filename_parts, *layers, parent='item/generated'):
...
+        filename_parts=('crop', crop),
+        textures={texture: "tfc:blocks/crop/%s_0" % crop},
+        model=mymodel,
+        variants=variants
+    )
+    blockstate(
+        filename_parts=('dead_crop', crop),
+        textures={texture: "tfc:blocks/crop/%s_dead" % crop},
+        model=mymodel
+    )
```
alcatrazEscapee: Python styling: don't separate arguments with newlines, don't name arguments unless they are necessary (i.e, non-positional), don't call things `mymodel`

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropSimple.java
```diff
@@ -90,63 +75,27 @@ public boolean onBlockActivated(World worldIn, BlockPos pos, IBlockState state,
...
             {
+                ItemStack seedDrop = new ItemStack(ItemSeedsTFC.get(crop));
+                seedDrop.setCount(0);
+                SimpleSkill skill = CapabilityPlayerData.getSkill(playerIn, SkillType.AGRICULTURE);
+
+                if (skill != null)
+                {
+                    foodDrop.setCount(1 + Crop.getSkillFoodBonus(skill, RANDOM));
+                    // omit the +1 because the plant stays alive.
+                    seedDrop.setCount(Crop.getSkillSeedBonus(skill, RANDOM));
```
alcatrazEscapee: Would be nicer to have
```java
ItemStack seedDrop = ItemStack.EMPTY;
```
and then
```java
seedDrop = new ItemStack(ItemSeedsTFC.get(crop), Crop.getSkillSeedBonus(skill, RANDOM));
```
As `.setCount(0)` is really ugly.

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropSpreading.java
```diff
@@ -6,44 +6,69 @@
...
 
 import net.dries007.tfc.api.types.ICrop;
-import net.dries007.tfc.objects.items.ItemSeedsTFC;
 import net.dries007.tfc.objects.te.TECropSpreading;
-import net.dries007.tfc.objects.te.TETickCounter;
 import net.dries007.tfc.util.Helpers;
 
-import static net.dries007.tfc.util.agriculture.Crop.STAGE_8;
-
 @ParametersAreNonnullByDefault
```
alcatrazEscapee:
```suggestion
@Deprecated // For now, unused, needs design work
@ParametersAreNonnullByDefault
```

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropTFC.java
```diff
@@ -73,67 +115,52 @@ public static BlockCropTFC get(ICrop crop)
         setHardness(0.6f);
     }
 
+    // public methods
```
alcatrazEscapee: ew
```suggestion
```

DavidTriphon: gosh darnit I thought I got them all

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropSimple.java
DavidTriphon: Would putting the 0 in the constructor be okay? I feel like `seedDrop` should always be a seed item during the variable's lifespan in case someone needs to change how the amount of seeds dropped is calculated outside the scope of that if statement.

generateResources.py
DavidTriphon: I used `mymodel` even though it's bad naming convention because when I used `model` it overloaded the function `model()` since it all runs on the import scope.

generateResources.py
DavidTriphon: While I do appreciate the nature of creating the dictionary in one call with the embedded for, we definitely disagree stylistically about the use of newlines. You love compressing code but I love separating code across lines.

PR #750, opened 2020-01-17
Crops know how time affects them even when they aren't loaded.

DavidTriphon:
If a crop is planted in October, and then left unloaded until May, it should die because it went through the winter, even though May is a growing month. If a crop is planted in May and left unloaded until October, it should grow because it went through the summer, even though October is a cold month (but not cold enough to die). This commit adds this behavior.

This PR is built on top of #749.


alcatrazEscapee:
Can you merge this branch into #749? Lots of changes are showing up duplicated here and it's easier if this is all one fixer-upper

DavidTriphon:
I merged this into crop stuff PR. #749

PR #785, opened 2020-02-03
Changes in JEI

### LINKED ISSUE #778: JEI Metals issues.
- Currently Casting is missing from all the JEI metal recipes.
- Melting recipes should be using the fluids instead of cast ingots.
- Alloys are only looking at ingots and ores not nuggets.
- Unmounted recipes should should show one per metal
- Metal heating should be search able by input and output.

DisasterMoo:
Should now be possible to search for any input -> output.
Closes #778


alcatrazEscapee:
bd40578060eea74ef990a360b107d01f9d5f3477

PR #809, opened 2020-02-11
A few enhancements

### LINKED ISSUE #586: HUD display changes
Configurable style options on how to display health, thirst and hunger bars.

Health
- [x] Vanilla (Heart Textures)
- [x] TFC Classic (Health bar with customizable format)

Hunger
- [x] Vanilla (10 meat chunks)
- [x] TFC Classic (green bar)

Thirst
- [x] TFC Classic (blue bar)
- [x] None (disable thirst bar)

### LINKED ISSUE #620: Crucible GUI improvements / rewrite
- [x] Add more input slots (9 ?)
- [x] Render the fluid texture in GUI
- [x] Scroll or some other form of handling a lot of materials (see #544 )

### LINKED ISSUE #651: Anvil plan selection GUI needs scrollbar / page buttons to display large amounts of recipes
When adding many new recipes in anvil GUI (e.g. TCon parts), the recipe buttons begin to overlap the rest of the GUI. This is a problem because anvil work needs to be fast. Preselection of a recipe is one option.

 - TFC Version: build 105
 - Is this a compatibility feature for other mods? Yes.

### LINKED ISSUE #655: Bloom percentage to be visible on standard tooltip
Bloom percentage can only be seen on Advanced Tooltip, which makes it rather obscure.
Adding this to the standard tooltip would make it clear why <100% split bloom can't be worked into an ingot.

**Meta Info**
 - TFC Version: 0.X.X.105

### LINKED ISSUE #721: Implement falling block missing features?
Falling blocks/cobblestone don't do the following:
destroy loose ore items
destroy ore blocks
damage the player

Play would be easier with these gone, but they are classic features.

### LINKED ISSUE #767: Vanilla eggs can still spawn vanilla baby chickens.
**Describe the bug**

Not really a bug, but it's not in line with TFC to spawn a different variant of chicken that can just drop eggs freely and be bred using vanilla seeds.

**To Reproduce**
Throw egg.

**Meta Info**
 - TFC Version: 1.12.2-0.28.2.113

### LINKED ISSUE #786: Forging Safety
Just a newbie helping hand: Prevent any forging to start using red rules.

### LINKED ISSUE #803: Log pile interaction tweaks
Reasoning:
 - a few people have mentioned it can be a pain to try and fill log piles (it requires 16 clicks to put logs in, or opening a GUI and shift clicking / mouse moving logs into the inventory)
 -  we want to preserve the ability to place log piles without consuming all possible logs (because in occasion they are still used as building materials)

Summary:
 - shift click log onto block = place log pile (use one log)
 - right click log pile holding log = place one log into pile without opening gui
 - shift right click log pile holding log = place all the logs in your hand in the pile without opening gui
 - [shift|] right click log pile without holding a log = open gui

DisasterMoo:
A few enhancements:
- Log pile interaction (Closes #803).
- Safety Forging (Closes #786).
- Missing falling features from classic (Closes #721).
- Bloom tooltips (Closes #655).
- Disable  vanilla animals entirely (Closes #767).
- Crucible tweaks / GUI Rewrite (Closes #620).
- Added page selection buttons to Anvil  Plan GUI (Closes #651).
- Customizable HUD (Closes #586).


src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -565,6 +566,16 @@ public static void onLivingSpawnEvent(LivingSpawnEvent.CheckSpawn event)
...
     }
 
+    @SubscribeEvent
+    public static void onEntityJoinWorldEvent(EntityJoinWorldEvent event)
+    {
+        // Prevent vanilla animals (that have a TFC counterpart) from mob spawners / egg throws / other mod mechanics
+        if (ConfigTFC.GENERAL.forceNoVanillaAnimals && Helpers.isVanillaAnimal(event.getEntity()))
+        {
+            event.setCanceled(true);
+        }
```
alcatrazEscapee: There's two concerns here:
1) Should we try and spawn a TFC replacement mob instead? If the list of vanilla animals is hardcoded we should be able to guess / replace the spawning of other mobs here. And if so,
2) Eggs should not spawn TFC chickens when thrown as we have separate mechanics for that. So we need to stop that regardless, and I don't think a blanket check for animals is the right way to go about that. Look into `ProjectileImpactEvent.Throwable` instead for thrown eggs on impact.

src/main/java/net/dries007/tfc/client/gui/GuiAnvilPlan.java
```diff
@@ -100,16 +116,72 @@ protected void drawGuiContainerBackgroundLayer(float partialTicks, int mouseX, i
...
         {
-            // This fires when you select a plan in the Plan GUI
-            TerraFirmaCraft.getLog().info("Pressed the plan button");
-            ResourceLocation recipeName = ((GuiButtonAnvilPlanIcon) button).getRecipeName();
-            TerraFirmaCraft.getNetwork().sendToServer(new PacketGuiButton(button.id, new NBTBuilder().setString("recipe", recipeName.toString()).build()));
+            int posX = 7 + (i % 9) * 18;
+            int posY = 25 + ((i % 18) / 9) * 18;
+            addButton(new GuiButtonAnvilPlanIcon(recipeList.get(i), ++buttonID, guiLeft + posX, guiTop + posY));
+        }
+        buttonLeft = addButton(new GuiButton(++buttonID, guiLeft + 7, guiTop + 65, 14, 14, "")
```
alcatrazEscapee: This should really be a seperate class instead of inline ones. It's cleaner, esp. considering these buttons should probably have tooltips. (See `IButtonTooltip` which we implement on buttons to do automatic on-hover-tooltip recognition)

src/main/java/net/dries007/tfc/network/PacketAnvilUpdate.java
```diff
@@ -56,7 +57,8 @@ public void fromBytes(ByteBuf buffer)
         workProgress = buffer.readInt();
         workTarget = buffer.readInt();
         pos = BlockPos.fromLong(buffer.readLong());
-        steps = ForgeSteps.deserialize(buffer.readInt());
+        steps = new ForgeSteps();
+        steps.deserializeNBT(ByteBufUtils.readTag(buffer));
```
alcatrazEscapee, line 61: Is there a reason this was dropped? It's significantly cheaper to serialize one int than an NBT tag as a string.

DisasterMoo: I've changed since serialization using bit shift didn't work correctly for negative (client deserialization only worked correctly and updated GUI after all three steps weren't null/-1). Since `ForgeSteps` NBT only use 3 integers, i don't think it will cause much overhead, ~~it's *almost* the same as writing them directly to the buffer~~. Okay, concede it's significantly costly, but should it matter? it's only sent on user input. If you want we can write all 3 integers to the buffer instead of doing bit shift shenanigans.

*please don't make me rewrite the bit shift operation to work with only one integer* 😩

src/main/java/net/dries007/tfc/CommonEventHandler.java
DisasterMoo: Not sure if we should spawn TFC mobs instead. The implications of doing so is:
- Mob spawners would spawn our mobs
- Using vanilla spawn eggs also would spawn our mobs
- **And more importantly** Any other mod trying to spawn vanilla mobs would spawn our mobs too (and if said mod would try to do something like Class Cast to `EntityChicken` without checking it's instance, BOOM crash)

At least canceling the event would let the other mod know that the spawning is prevented.

Now if we want to prevent egg throwing without config, i can look at `ProjectileImpactEvent.Throwable`

alcatrazEscapee: I very much doubt that would be an issue. We're not trying to intercept `newInstance`, we're trying to intercept as late as possible, when the entity is actually getting added to the world (`spawnEntity`) More importantly, that method does not return the entity that was actually spawned - unless a mod spawned an entity, then queried the world for the exact entity which they already have a reference to (because they constructed it), *and* then did that unsafely, there would be problems.

As for mob spawners (although what spawners?) or vanilla eggs, I think this is fine, predicate that it's configurable (which it already is)

The effect of spawning a different entity is just as bad as not spawning one - a `ClassCastException` is replaced with a `NullPointerException`

src/main/java/net/dries007/tfc/network/PacketAnvilUpdate.java
alcatrazEscapee: No, it's not important, I was mostly curious. If there's a bug then this is case and point for bit shift magic being too complicated and hard to see errors in.

(If I've done the math correctly, writing the tag is 33 bytes once you include tag names, header info, and three integers)

PR #810, opened 2020-02-12
Fix Blast Furnaces losing fuel on save/load

DavidTriphon:
What it says on the tin


alcatrazEscapee:
81eb419981b299537acc86e5cae22affd4476034

PR #812, opened 2020-02-13
Fixed support beams to only be placeable on beams at least three blocks high

DavidTriphon:
What it says on the tin.

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
+++ b/src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
@@ -293,7 +293,6 @@ public class BlockSupport extends Block
         else
         {
             if (!isConnectable(world, pos, side.getOpposite())) return false;
-            if (!isThreeTall(world, pos.offset(side.getOpposite()))) return false;
             int distance = getHorizontalDistance(side, world, pos);
             return distance > 0;
         }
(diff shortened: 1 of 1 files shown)


Fr4gtastic:
To be clear — is it only three błocks or at least three błocks?

DavidTriphon:
At least. It only checks the two blocks beneath the attachment block. Beyond that it doesn’t care if it is solid or support.

DavidTriphon:
The picture shows beams placed on 3, 4, and 5 height beams but not beneath that.

Fr4gtastic:
I see, thanks for the clarification.

src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
```diff
@@ -292,10 +292,8 @@ public boolean canPlaceBlockOnSide(@Nonnull World world, @Nonnull BlockPos pos,
         }
         else
         {
-            if (!isConnectable(world, pos, side.getOpposite()))
-            {
-                return false;
-            }
+            if (!isConnectable(world, pos, side.getOpposite())) return false;
+            if (!isThreeTall(world, pos.offset(side.getOpposite()))) return false;
             int distance = getHorizontalDistance(side, world, pos);
```
DisasterMoo, line 296: This is only checking on one side. Better to implement that logic on `getHorizontalDistance` and only return a value if both sides are at least three tall.

DisasterMoo (changes requested):

DavidTriphon:
Fixed 3 tall check to work on both sides and moved 3 tall check to the `getHorizontalDistance` method.

PR #834, opened 2020-02-19
Little adjustments on en_us

HeartyYF:


src/main/resources/assets/tfc/lang/en_us.lang
```diff
@@ -3,7 +3,7 @@ tfc.command.time.usage=/timetfc [add|query|set] [args...]
 tfc.command.time.usage_expected_first_argument=Expected [add|query|set] after /timetfc
 tfc.command.time.usage_expected_second_argument_query=Expected [daytime|day|gametime|playerticks|calendarticks] after /timetfc query
 tfc.command.time.usage_expected_second_argument_set=Expected [day|night|monthlength <value>] after /timetfc set
-tfc.command.time.usage_expected_second_argument_add=Expected [days|months|years|] <value> after /timetfc add
+tfc.command.time.usage_expected_second_argument_add=Expected [days|months|years] <value> after /timetfc add
```
alcatrazEscapee, line 6: This is correct, `/timetfc add <value>` is a valid command, hence, allowing for nothing.

HeartyYF: According to [minecraft wiki](https://minecraft.gamepedia.com/Commands#Syntax), `[]` has already shown that its inside are optional, so allowing for nothing doesn't need to be extra announced.

PR #837, opened 2020-02-19
A few improvements

### LINKED ISSUE #690: Separate Heat and Forgeability capabilities.
Currently forgeability requires heat to exist.
This makes a heat cap be required on an item to forge it even if you don't need the heat for the item you must put it on there. These are two separate systems that just happen to be used together and they should be treated as such. There isn't any reason for an anvil recipe to require heat, it was only made this way as that's hows its used within TFC.

This makes anvils operate less like a crafting table and more like a only for this mod block.

Splitting them opens up the anvil to easier customization for pack makers and possibly more use within TFC itself.
Either a new registry option can be added or the input of the recipes can be looked at and automatically be registered.

### LINKED ISSUE #796: Add configs for all fluid related containers where possible.
Things like the Crucible, Vessel, Barrel and such should all have config options for the maximum amount they can hold (if it can be done without remaking the whole thing)

### LINKED ISSUE #832: /timetfc add always shows how many ticks were added.
**Describe the bug**
When using command `/timetfc add` , it always shows how many ticks were added, rather than how many years/months/days.

Like this, using `/timetfc add years 1` shows "Added 2304000 years to the time."

**To Reproduce**
Use command `/timetfc add years 1`.

**Meta Info**
 - TFC Version: 0.29.3.118

### LINKED ISSUE #833: /timetfc has wrong auto-complete
**Describe the bug**
After typing `/timetfc set` or `/timetfc query,` pressing tab shows inconsistent alternatives with practical command.

Here should shows `calendarticks, day, daytime, gametime, playerticks`

Here lacks `monthlength`.

**To Reproduce**
After typing `/timetfc set` or `/timetfc query,` press tab.

**Meta Info**
 - TFC Version: 0.29.3.118

### LINKED ISSUE #835: Localization on %s of time.query and time.added
**Describe the request**
Now these two commands shows partly untranslated for %s is always in English, like these two: 

(not the final Chinese translation, just for test)

If more localization can be introduced, I think this can be solved. 

**Meta Info**
 - TFC Version: 0.29.3.118

DisasterMoo:
- Added localization and improved output information of `timetfc` command (Closes #835, Closes #833, Closes #832).
- Added config to change how much fluid containers can hold (Closes #796).
- Separated Forging capability from heat (Closes #690). Added an extension class to use both for the normal functionality.

### DIFF
--- a/src/main/java/net/dries007/tfc/ConfigTFC.java
+++ b/src/main/java/net/dries007/tfc/ConfigTFC.java
@@ -359,6 +359,21 @@ public class ConfigTFC
         @Config.RangeInt(min = 0, max = 2)
         @Config.LangKey("config." + MOD_ID + ".general.inventoryCraftingTableMode")
         public int inventoryCraftingTableMode = 0;
+
+        @Config.Comment("How much can a small vessel hold?")
+        @Config.RangeInt(min = 100)
+        @Config.LangKey("config." + MOD_ID + ".general.tankSmallVessel")
+        public int tankSmallVessel = 4000;
+
+        @Config.Comment("How much can a barrel hold?")
+        @Config.RangeInt(min = 100)
+        @Config.LangKey("config." + MOD_ID + ".general.tankBarrel")
+        public int tankBarrel = 10000;
+
+        @Config.Comment("How much can a crucible hold?")
+        @Config.RangeInt(min = 100)
+        @Config.LangKey("config." + MOD_ID + ".general.tankCrucible")
+        public int tankCrucible = 3000;
     }
 
     public static class ClientCFG
--- a/src/main/java/net/dries007/tfc/ConfigTFC.java
+++ b/src/main/java/net/dries007/tfc/ConfigTFC.java
@@ -360,17 +360,17 @@ public class ConfigTFC
         @Config.LangKey("config." + MOD_ID + ".general.inventoryCraftingTableMode")
         public int inventoryCraftingTableMode = 0;
 
-        @Config.Comment("How much can a small vessel hold?")
+        @Config.Comment("How much metal (units / mB) can a small vessel hold?")
         @Config.RangeInt(min = 100)
         @Config.LangKey("config." + MOD_ID + ".general.tankSmallVessel")
         public int tankSmallVessel = 4000;
 
-        @Config.Comment("How much can a barrel hold?")
+        @Config.Comment("How much fluid (mB) can a barrel hold?")
         @Config.RangeInt(min = 100)
         @Config.LangKey("config." + MOD_ID + ".general.tankBarrel")
         public int tankBarrel = 10000;
 
-        @Config.Comment("How much can a crucible hold?")
+        @Config.Comment("How much metal (units / mB) can a crucible hold?")
         @Config.RangeInt(min = 100)
         @Config.LangKey("config." + MOD_ID + ".general.tankCrucible")
         public int tankCrucible = 3000;
--- a/src/main/java/net/dries007/tfc/api/capability/forge/ForgeableHeatableHandler.java
+++ b/src/main/java/net/dries007/tfc/api/capability/forge/ForgeableHeatableHandler.java
@@ -1,3 +1,8 @@
+/*
+ * Work under Copyright. Licensed under the EUPL.
+ * See the project README.md and LICENSE.txt for more information.
+ */
+
 package net.dries007.tfc.api.capability.forge;
 
 import javax.annotation.Nonnull;
--- a/src/main/java/net/dries007/tfc/api/capability/forge/ForgeableHeatableHandler.java
+++ b/src/main/java/net/dries007/tfc/api/capability/forge/ForgeableHeatableHandler.java
@@ -1,8 +1,3 @@
-/*
- * Work under Copyright. Licensed under the EUPL.
- * See the project README.md and LICENSE.txt for more information.
- */
-
 package net.dries007.tfc.api.capability.forge;
 
 import javax.annotation.Nonnull;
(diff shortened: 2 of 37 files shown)


src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -359,6 +359,21 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
         @Config.RangeInt(min = 0, max = 2)
         @Config.LangKey("config." + MOD_ID + ".general.inventoryCraftingTableMode")
         public int inventoryCraftingTableMode = 0;
+
+        @Config.Comment("How much can a small vessel hold?")
```
alcatrazEscapee:
```suggestion
        @Config.Comment("How much metal (units / mB) can a small vessel hold?")
```
And the same for others, including the units here would be better

src/main/java/net/dries007/tfc/api/capability/forge/ForgeableHeatableHandler.java
```diff
@@ -0,0 +1,130 @@
...
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.util.EnumFacing;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.common.capabilities.Capability;
+
+import net.dries007.tfc.api.capability.heat.ItemHeatHandler;
+import net.dries007.tfc.util.forge.ForgeStep;
+import net.dries007.tfc.util.forge.ForgeSteps;
+
+public class ForgeableHeatableHandler extends ItemHeatHandler implements IForgeableHeatable
```
alcatrazEscapee, line 20: Rather than duplicating logic from `ForgeableHandler`, it would make more sense imo to have an internal reference to one, then delegate methods to that. See how `FoodHeatHandler` refers to an internal `FoodHandler` while extending `ItemHeatHandler`. It's nicer if we ever have to fix / change `ForgeableHandler`, we don't have to also remember to fix it here as well.

src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
```diff
@@ -408,9 +408,8 @@ private boolean isThreeTall(IBlockAccess world, BlockPos pos)
         if (axis != EnumFacing.Axis.Y) return false;
         // if either of the two block beneath this block are not block supports, then this isn't three tall
         if (!(world.getBlockState(pos.down()).getBlock() instanceof BlockSupport)) return false;
-        if (!(world.getBlockState(pos.down().down()).getBlock() instanceof BlockSupport)) return false;
+        return world.getBlockState(pos.down().down()).getBlock() instanceof BlockSupport;
         // passed all the checks
```
alcatrazEscapee:
```suggestion
```

src/main/java/net/dries007/tfc/client/gui/GuiAnvilPlan.java
```diff
@@ -107,6 +84,29 @@ protected void drawGuiContainerForegroundLayer(int mouseX, int mouseY)
         super.drawGuiContainerForegroundLayer(mouseX, mouseY);
     }
 
+    @Override
+    protected void actionPerformed(GuiButton button) throws IOException
+    {
+        if (button instanceof GuiButtonAnvilPlanIcon)
+        {
+            // This fires when you select a plan in the Plan GUI
+            TerraFirmaCraft.getLog().info("Pressed the plan button");
```
alcatrazEscapee:
```suggestion
```
Not your fault, just something I noticed :P

src/main/java/net/dries007/tfc/objects/te/TEBarrel.java
```diff
@@ -79,6 +79,14 @@ public TEBarrel()
...
     {
         tank.readFromNBT(nbt.getCompoundTag("tank"));
+        if (tank.getFluidAmount() > tank.getCapacity())
+        {
+            // Fix config changes
+            FluidStack fluidStack = tank.getFluid();
+            //noinspection ConstantConditions
+            fluidStack.amount = tank.getCapacity();
+            tank.setFluid(fluidStack);
+        }
```
alcatrazEscapee, line 89: Is this preventing a crash? Or trying to do some sort of data-fixer change for configuration changes. If the latter, I don't think we should, as it's going to be inconsistent (already I see crucibles just ignore possible changes). It's reasonable to assume that if you change a configuration option, it won't affect previous items / blocks destructively.

src/main/java/net/dries007/tfc/objects/te/TECrucible.java
```diff
@@ -68,7 +66,7 @@ public TECrucible()
     {
         super(10);
 
-        this.alloy = new Alloy(CRUCIBLE_MAX_METAL_FLUID);
+        this.alloy = new Alloy(ConfigTFC.GENERAL.tankCrucible); // Side effect: Maximum amount only matches config if not loading from disk
```
alcatrazEscapee, line 69: See above comment re: data fixer on config option changes

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/objects/te/TEBarrel.java
DisasterMoo: It's the latter, mostly because fluid render overflows in GUI. I will do like crucibles then

src/main/java/net/dries007/tfc/objects/te/TECrucible.java
DisasterMoo: Made Crucibles void the surplus instead to match barrels.

PR #839, opened 2020-02-20
Tfc Advancements

Mithrik:
Adds a set of two advancement trees to TFC:

1. A main line of progression, from picking a rock, all the way up to making metal buckets. While inspired in 1.7.10 Classic's achivements, it is significantly expanded to be more detailed and serve as a better in-game guide.
2. A secondary "world exploration" tree that sets up a few sidequests and secondary objectives, such as those related to finding ore veins, and farming.

Additionally, the "en_us.lang" file has been expanded to hold the text entries for these advancements. The advancements themselves can be translated by anyone who volunteers to update the other language files, of course.


src/main/resources/assets/tfc/lang/en_us.lang
```diff
@@ -3826,4 +3826,137 @@ tfc.animal.horsetfc.female=Horse♀
...
 tfc.animal.muletfc.male=Mule♂
-tfc.animal.muletfc.female=Mule♀
\ No newline at end of file
+tfc.animal.muletfc.female=Mule♀
+
+## Advancements
+
+# Main Quest
+# Stone Age
+advancements.main.root.title=TerraFirmaCraft
```
DisasterMoo: To avoid lang files collision (where another mod / resource pack use the same lang key) everything within TFC's realm should have use `<prefix.if.needed>.tfc.<item.name>` standard. So, for advancements, i recommend:
`advancements.tfc.<tree>.<advancement>.title`

DisasterMoo (changes requested):
Looking good, great work!
In my tests, just needed two things:
- Lang entries should have `tfc` prefix to avoid lang collision (see below)
- You put your files under a unnecessary  TFC folder, but your json references correctly if you move them up one folder (eg: intead of `assets/tfc/advancements/tfc/main/<files>`, use `assets/tfc/advancements/main/<files>`, Asset collisions only happens for lang files because mojang

Mithrik:
Took care of the changes pointed out oby @DisasterMoo. The lang file refenrences now go `advancements.tfc.<tree>.<file>.title/description`.

All the JSONs were modified accordingly and the `assets/tfc/advancements/tfc/*` directory has been shortened to `assets/tfc/advancements/*`

src/main/resources/assets/tfc/advancements/world/kill_bear.json
```diff
@@ -0,0 +1,27 @@
...
+        "description": {
+            "translate": "advancements.tfc.world.kill_bear.description"
+        }
+    },
+    "parent": "tfc:world/root",
+    "criteria":{
+        "native_copper": {
+            "trigger": "minecraft:player_killed_entity",
+            "conditions": {
+                "entity": [
```
DisasterMoo: While testing, this file crashed. Looking further, this is because here we have an extra `[` and `]` brackets. just remove them:
```java
"entity": {
    "type": "tfc:beartfc"
}
```

DisasterMoo (changes requested):
Almost ready! Sorry I didn't catch the following error last time, but just fixing it makes your PR ready to be merged

src/main/resources/assets/tfc/advancements/world/kill_bear.json
Mithrik: Resolved. Along the ways I noticed the icon was wrong, so fixed that too.

DisasterMoo (approved):

alcatrazEscapee:
From discord:
> Suggestion: Adding three advancement hooks. One for when a player first lights a fire pit, another for when they first familiarize an animal and a third one for when they max out the familiarization of an animal.

Would these still be useful, i.e. if we add hooks for these, would you be interested in adding advancements using the above? (I believe this would be fairly simple, via adding an `ICriterionTrigger` to `CriteriaTriggers`, but am not sure)

Mithrik:
Yes they would all be useful. Currently there's no advancements for the fire pit or familiarization precisely because vanilla can't detect these things. I think you can detect feeding and breeding TFC animals OK, but haven't tested.

I would add the advancements in if the hooks are added.

DisasterMoo:
I can help you with that. But i guess you will need to first be able to run tests using the source code instead of waiting for new versions 😉 

If you're up to it, can you write which hooks would be useful, or are those the only ones you want?

Now to test changes, check if you can use the command line (or terminal in linux distros) and run the command `gradlew runClient` in the same folder where you cloned your repo. If you need more help just DM me in discord.

Mithrik:
Just the three hooks I requested is fine. Even if there are other useful ones, there's actually very little that can't be detected by checking a player's inventory for an item.

As for the testing from the source, I did manage to run my fork from the local repo without issue, so I should be set to test when the time comes.

DisasterMoo:
I just sent a PR to your repository. Accept that and you will have the requested triggers implemented.

Mithrik:
The custom advancements, a few tweaks here and there and the addition of XP rewards as treats for completing goals and challenges have all been committed.

PR #843, opened 2020-02-21
More fixes

### LINKED ISSUE #647: Baby animals levitate, and death animation can hover in the air.
**Describe the bug**
Make sure to include:
My sheep to be normal fluffy creatures.
The baby sheep was floating.
Praise be watersheep!
image

**To Reproduce**
1. Spawn a baby mob.

**Meta Info**
 - TFC Version: 1.12.2-0.27.3.103

### LINKED ISSUE #761: Hammer+log -> sticks is not working
Reported that hammer on log breaks and drops log instead of sticks.

**Meta Info**
 - TFC Version: build 111?

### LINKED ISSUE #766: Cobble duplication using stair blocks and gravity
**Describe the bug**
When placing cobble stairs in a place where cobble becomes gravity-affected, normally you'd expect the block to turn into a single full cobble block and fall into a resting position.

However, if you place a stair in place where it has more than one side to fall into, such as a corner or a one block wide elevation, the stair block will produce a duplicate cobble block, so you end up with two cobble blocks instead of one.

**To Reproduce**
Place a cobble stair block on the edge on an elevation that triggers it to fall from gravity, which has two or more sides the block can fall towards.

**Meta Info**
 - TFC Version: 0.28.1.111

### LINKED ISSUE #780: Eggs don't cook
Eggs do not cook in firepits or forges. They eventually just burn.
JEI shows the recipe, though.

 - TFC Version: build 115

DisasterMoo:
- Swapped levitate mechanic from animals to cobble stairs (Closes #766, Closes #647)
- Eggs can now be properly destro... i mean, cooked (Closes #780)
- Stop! Hammertime! Hammers are now a proper tool to break logs, which in turn become sticks (Closes #761).

Also, missing wolf cleanup (removed `ModelWolfTFC`)


src/main/java/net/dries007/tfc/types/DefaultRecipes.java
```diff
@@ -319,7 +319,7 @@ public static void onRegisterHeatRecipeEvent(RegistryEvent.Register<HeatRecipe>
             HeatRecipe.destroy(IIngredient.of(ItemFoodTFC.get(Food.COOKED_VENISON)), 480).setRegistryName("burned_venison"),
 
             // Egg
-            HeatRecipe.destroy(IIngredient.of(Items.EGG), 480).setRegistryName("burned_egg")
+            HeatRecipe.destroy(IIngredient.of(ItemFoodTFC.get(Food.COOKED_EGG)), 480).setRegistryName("burned_egg")
```
alcatrazEscapee, line 322: This destroys cooked eggs, yes, but the issue was that eggs aren't able to be cooked in the first place? See L297: 
```java
new HeatRecipeSimple(IIngredient.of(Items.EGG), new ItemStack(ItemFoodTFC.get(Food.COOKED_EGG)), 200, 480).setRegistryName("cooked_egg"),
```
which should already be destroying eggs. I don't see how this fixes this issue (but it may just be I'm confused)

DisasterMoo: The issue is rather simple:
https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/8519ab95386ffd3d3ee5647efb270e50e7ff43c4/src/main/java/net/dries007/tfc/api/recipes/heat/HeatRecipe.java#L41-L44
returns the first valid found recipe, which was returning the destruction recipe. Just looking at the above lines you will find that the other cooking is registered by Ingredient -> Cooked, and then Cooked -> Destroy recipe (at a higher temp).

alcatrazEscapee: Perfect, thanks

PR #865, opened 2020-02-28
Tiny bugs

### LINKED ISSUE #497: Grain crafting doesn't produce extra straw
Knife + raw grains in the crafting grid should produce extra straw. Requires a separate crafting recipe type to produce extra output, similar to molds.

As of 0.25.2.86

### LINKED ISSUE #716: Do something with dirt gold pans
Either disable them or have them do something.

Gaelmare:
Generate straw when grain is threshed with knife. Fixes #497.
Stop dirt/clay filled gold pans. Fixes #716

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/recipes/ShapelessDamageFoodRecipe.java
+++ b/src/main/java/net/dries007/tfc/objects/recipes/ShapelessDamageFoodRecipe.java
@@ -9,6 +9,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 import com.google.gson.JsonObject;
+import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.inventory.InventoryCrafting;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.IRecipe;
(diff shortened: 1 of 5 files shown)


Gaelmare:
For #497, had a question about whether I should rename ShapelessDamageFoodRecipe to ShapelessDamageFoodStrawRecipe, subclass it that way, or add a check like:
`OreDictionary.getOres("grain").stream().anyMatch(s -> s.getItem() == output.getItem())
`
to it.
Threshing grain with knife is the only current use of this recipe.

src/main/java/net/dries007/tfc/objects/recipes/ShapelessDamageFoodRecipe.java
```diff
@@ -56,6 +60,20 @@ public ItemStack getCraftingResult(InventoryCrafting inv)
         return foodStack != null ? CapabilityFood.updateFoodFromPrevious(foodStack, out) : ItemStack.EMPTY;
     }
 
+    @Override
+    @Nonnull
+    public NonNullList<ItemStack> getRemainingItems(final InventoryCrafting inventoryCrafting)
+    {
+        // Give straw to player as well.
+        EntityPlayer player = ForgeHooks.getCraftingPlayer();
+        if (!player.world.isRemote)
```
DisasterMoo: Like you said, missing a check here for ore dictionary: `OreDictionaryHelper.doesStackMatchOre(stack, "grain")`

Gaelmare: Ooh, I like the oredict helper! Done.

PR #871, opened 2020-03-03
Some JEI, CT and Veins improvements

### LINKED ISSUE #608: Optional ore sample items to generate over custom registered ores.
**Describe the request**
Make sure to include:
Inability to find ores that don't have samples, other than using a prospectors pickaxe on random rock surfaces.
Players would be able to configure what items are considered the "sample", that generates above an ore vein, making ore hunting for custom ores possible.

**Meta Info**
 - TFC Version: 1.12.2-0.27.1.99 (at time of request)
Technically a TFC standalone enhancement, but most cases would use other mods.

### LINKED ISSUE #646: TFC Inventory tab buttons overlap with JEI
**Describe the request**
Make sure to include:
Annoying JEI and Inventory clashing.
The TFC inventory tabs for example would push JEI out of the way, this is standard for most mods that add slots that would otherwise conflict with JEI.

Shown here is TFC conflicting with JEI whilst Quark pushes it aside.
image

**Meta Info**
 - TFC Version: 1.12.2-0.27.3.103 (at time of request)
Both
Just Enough Items.

### LINKED ISSUE #738: Chisel JEI and CT integration
Missing from #733, should be added eventually for compatibility:
 - Chisel recipe viewing for smoothing (stairs and slabs are dynamic based on crafting recipes)
 - CT integration for chisel smoothing recipes. Recipe registry is already setup.
 - Documentation in github wiki for any CT syntax that's added

Not including extensible of chisel modes: "smoothing" is a generic recipe mode and should be used by pack makers (when possible) as such.

### LINKED ISSUE #857: Configuration Option for what "debris" spawns above specific rocks/ores/near trees.
**Describe the request**
Make sure to include:
Useful for pack developers to integrate new items to TFC.
You would be able to configure what can be found as debris, in the form of surface samples (above ores), rocks (above select or all rock types), or sticks, (around/near trees/however sticks are told to generate right now).

**Meta Info**
 - TFC Version: 1.12.2-0.29.4.119 (at time of request)
Yes.

### LINKED ISSUE #867: Default ore generation balance issues
Probably an overlook:
- Kaolinite is rarer because it can only generates at top layer(y = 110+) while it's default config is set from y 5 - 128.
- Graphite has the same issue, although less rare because it can generates at the second layer.

### LINKED ISSUE #868: JEI integration to vein generation
As per discord discussion, two new categories:
- Rock Category: Which ore generates there + possible layers.
- Ore Category: Where it generates (Which Rocks), minimum and maximum height, rarity and small ore indication.

DisasterMoo:
Fixed JEI inventory tab buttons overlap (Closes #646).
Added JEI and CT integration to Chisel recipes (Closes #738).
Made category UIDs public for addons that wants to add custom recipe wrappers to TFC categories.
(ie: https://github.com/TFC-Metallum/TFCTech/blob/b33a87a180374a98d1b515d1dde0e1d0f0b4f444/src/main/java/tfctech/compat/jei/TechJEIPlugin.java#L147 where i add a custom block and pane mold for glass recipes, and hook then in the casting category)
Added JEI integration to rock layers and veins (Closes #868).
Added loose rock configuration to vein json loading (Closes #857, Closes #608).
Tweaked ore veins to be more forgiving (Closes #867).


src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -59,30 +46,165 @@ public String getName()
...
+        if (args.length != 2 && args.length != 3) throw new WrongUsageException("2 or 3 arguments required.");
         if (sender.getCommandSenderEntity() == null) throw new WrongUsageException("Can only be used by a player");
 
-        sender.sendMessage(new TextComponentString("Veins Found: "));
+        final List<ChunkPos> chunks = new LinkedList<>();
+        final Set<BlockPos> veinsFound = new HashSet<>(); // Using BlockPos instead of vein objs lowers ram usage
+        final String filter = args[0];
+
+        // Default print to player chat
+        Consumer<Chunk> consumer = chunk ->
```
alcatrazEscapee: Using consumers here is super ugly. Is there a reason it was done this way? I see a couple possible options that are much nicer from a code smell point of view:
 - pass these directly as arguments to the worker rather than tossing them around as variables.
 - implement these as abstract class members on the worker, then instantiate with anonymous class impl. (My preferred way)
 - add all as standard functions on the worker and just use flags to differentiate what to actually do.

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -59,30 +46,165 @@ public String getName()
...
+        {
+            ChunkDataTFC chunkData = ChunkDataTFC.get(chunk);
+            chunkData.getGeneratedVeins().stream()
+                .filter(vein -> !veinsFound.contains(vein.getPos()))
+                .filter(vein -> filter.equalsIgnoreCase("all") || vein.getType().getRegistryName().equalsIgnoreCase(filter))
```
alcatrazEscapee: If you're hyper-optimizing this for performance, this `all` check can be moved outside of the filter and only called once. Additionally, you could pre-build a `HashSet` of `VeinType`s that are required and check membership in there (`O(1)`). Probably better than string comparisons on every vein.

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -59,30 +46,165 @@ public String getName()
...
+                    }
+                };
+            }
+            else if (args[2].equalsIgnoreCase("rate"))
+            {
+                sender.sendMessage(new TextComponentString("Dumping vein rates, this is gonna take a while..."));
+                // Save files to MC's root folder
+                String fileName = "tfc-veins-rate.log";
+                Map<VeinType, Integer> veinRateMap = new HashMap<>();
+                Map<Rock, Integer> rockRateMap = new HashMap<>();
```
alcatrazEscapee: Again, if you're optimizing for performance, `TObjectIntHashMap` would be better

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -59,30 +46,165 @@ public String getName()
...
+                        .filter(vein -> filter.equalsIgnoreCase("all") || vein.getType().getRegistryName().equalsIgnoreCase(filter))
+                        .forEach(vein ->
+                        {
+                            veinsFound.add(vein.getPos());
+                            String dump = "Vein: " + vein.getType() + " at " + vein.getPos();
+                            if (!LogFileWriter.isOpen())
+                            {
+                                LogFileWriter.open(fileName);
+                            }
+                            LogFileWriter.writeLine(dump);
```
alcatrazEscapee: Is there a reason you're doing file I/O on every chunk? I can't imagine this is more efficient than simply building the data per chunk, storing it in a running `List` of some sort, then writing it all at once.

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -59,30 +46,165 @@ public String getName()
...
+
+                    rockRateMap.forEach((rock, count) -> LogFileWriter.writeLine("Rock: " + rock.getRegistryName() + " chunks: " + count));
+
+                    sender.sendMessage(new TextComponentString("vein rate file saved at: " + LogFileWriter.getFilePath()));
+                    LogFileWriter.close();
+                };
+            }
+            else
+            {
+                throw new WrongUsageException("3rd argument must be dump or rate");
```
alcatrazEscapee: Translation key

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -59,30 +46,165 @@ public String getName()
     @Nonnull
     public String getUsage(ICommandSender sender)
     {
-        return "/findveins [all|<vein name>] <radius> -> Finds all instances of a specific vein, or all veins within a certain chunk radius";
+        return "/findveins [all|<vein name>] <radius> [dump|rate] -> Finds all instances of a specific vein, or all veins within a certain chunk radius, if dump or rate save it to a log file";
```
alcatrazEscapee: Translation key

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -59,30 +46,165 @@ public String getName()
...
     }
 
     @Override
     public void execute(MinecraftServer server, ICommandSender sender, String[] args) throws CommandException
     {
-        if (args.length != 2) throw new WrongUsageException("2 arguments required.");
+        if (args.length != 2 && args.length != 3) throw new WrongUsageException("2 or 3 arguments required.");
```
alcatrazEscapee: Translation key, and be more descriptive. (i.e. see `CommandTimeTFC`)

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -99,7 +221,110 @@ public int getRequiredPermissionLevel()
...
+        else if (args.length == 3)
+        {
+            return getListOfStringsMatchingLastWord(args, "dump", "rate");
+        }
         return Collections.emptyList();
     }
+
+    private static class Worker implements WorldWorkerManager.IWorker
+    {
+        private final int dimension;
```
alcatrazEscapee: TFC will only ever generate veins in the overworld. This is only ever zero (including in this implementation). Any reason? Otherwise this can be removed.

src/main/java/net/dries007/tfc/command/CommandVeinInfo.java
```diff
@@ -0,0 +1,60 @@
...
+    public String getName()
+    {
+        return "veininfo";
+    }
+
+    @Override
+    @Nonnull
+    public String getUsage(ICommandSender sender)
+    {
+        return "/veininfo -> Show a list of all registered veins";
```
alcatrazEscapee: Translation key

src/main/java/net/dries007/tfc/command/CommandVeinInfo.java
```diff
@@ -0,0 +1,60 @@
...
+
+import net.minecraft.command.CommandBase;
+import net.minecraft.command.ICommandSender;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.util.text.TextComponentString;
+
+import net.dries007.tfc.world.classic.worldgen.vein.VeinRegistry;
+
+@ParametersAreNonnullByDefault
+public class CommandVeinInfo extends CommandBase
```
alcatrazEscapee: Is there a reason this is a command? All it does is log standard data. It could be quite easily just emitted at load time to the log - that would avoid having to load a new world, get in-game, run the command to check vein registration. It's a cleaner output method (i.e. already in a text file), and available as soon as veins are loaded.

src/main/java/net/dries007/tfc/compat/jei/TFCJEIPlugin.java
```diff
@@ -293,5 +335,46 @@ public void register(IModRegistry registry)
...
         registry.addRecipeClickArea(GuiFirePit.class, 79, 37, 18, 10, HEAT_UID);
+
+        // Fix inventory tab overlap see https://github.com/TerraFirmaCraft/TerraFirmaCraft/issues/646
+        registry.addAdvancedGuiHandlers(new TFCGuiAreas<>(GuiInventory.class));
+        registry.addAdvancedGuiHandlers(new TFCGuiAreas<>(GuiCalendar.class));
+        registry.addAdvancedGuiHandlers(new TFCGuiAreas<>(GuiNutrition.class));
+        registry.addAdvancedGuiHandlers(new TFCGuiAreas<>(GuiSkills.class));
+    }
+
+    private static class TFCGuiAreas<T extends GuiContainer> implements IAdvancedGuiHandler<T>
```
alcatrazEscapee: This doesn't make any sense to be an inner class, nor does the name imo. Suggested: `TFCInventoryGuiHandler` (matches the name of the container)

src/main/java/net/dries007/tfc/compat/jei/wrappers/ChiselRecipeWrapper.java
```diff
@@ -0,0 +1,53 @@
...
+            .forEach(stack ->
+            {
+                Block block = ((ItemBlock) stack.getItem()).getBlock();
+                if (recipe.matches(block.getDefaultState()))
+                {
+                    ingredients.add(stack);
+                }
+            });
+        // Ideally we should use Block#getPickBlock but we can't have a World and EntityPlayer at this point
+        output = new ItemStack(recipe.getOutputState().getBlock());
```
alcatrazEscapee: If this fails this will be empty, could we display something separate instead? i.e. debug block maybe?

src/main/java/net/dries007/tfc/util/LogFileWriter.java
```diff
@@ -0,0 +1,157 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+import net.dries007.tfc.TerraFirmaCraft;
+
+/**
+ * Handles writing text to a log file
+ * Makes code clean :D
```
alcatrazEscapee: I disagree strongly - this seems way more code than is nessecary here. Writing a single bunch of stuff to a file, if buffered properly:
```java
List<String> lines;
try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
    for (String s : lines) {
        writer.write(s);
    }
} catch (IOException e) {
    // log or whatever
}
```

src/main/java/net/dries007/tfc/util/json/VeinTypeJson.java
```diff
@@ -39,11 +43,25 @@ public VeinType deserialize(JsonElement json, Type typeOfT, JsonDeserializationC
             throw new JsonParseException("Minimum height cannot be greater than maximum height!");
         }
         int rarity = JsonUtils.getInt(jsonObject, "rarity");
+        if (rarity <= 0)
+        {
+            throw new JsonParseException("Rarity cannot be negative or zero!");
+        }
         int density = JsonUtils.getInt(jsonObject, "density");
         int width = JsonUtils.getInt(jsonObject, "width");
         int height = JsonUtils.getInt(jsonObject, "height");
```
alcatrazEscapee: Can add checks for nonnegative values for all these

src/main/java/net/dries007/tfc/world/classic/worldgen/vein/VeinRegistry.java
```diff
@@ -139,8 +143,21 @@ public void postInit()
                         VeinType vein = GSON.fromJson(entry.getValue(), VeinType.class);
                         vein.setRegistryName(properVeinName);
 
+                        // Last check before registering, make sure to not include any bogus veins
+                        if (vein.getOre() == null && vein.getOreState(Rock.GRANITE, Ore.Grade.NORMAL).getMaterial() == Material.AIR)
+                        {
+                            TerraFirmaCraft.getLog().error("Error: " + vein.getRegistryName() + " vein registered ore block is Air! Skipping.");
+                            continue;
+                        }
+
```
alcatrazEscapee: There's no need for this. If the block failed to resolve to a name, it should be already caught and the vein not registered. This is only preventing the case the user wrote `"minecraft:air"` as the block, in which case why would we deny that?

src/main/java/net/dries007/tfc/world/classic/worldgen/vein/VeinType.java
```diff
@@ -214,4 +227,61 @@ public IBlockState getOreState(Rock rock, Ore.Grade grade)
             return this.oreState;
         }
     }
+
+    public enum Rarity
```
alcatrazEscapee: What is the point of this? Rarity can already be displayed as an integer, and you're enforcing names upon specific values that may have no applicability to whatever custom veins someone is using.

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/world/classic/worldgen/vein/VeinType.java
DisasterMoo: Seems you lost the point when i was talking about it in discord. Plain number does nothing and is horrible to the average user, while names are actually helpful to them. Here are both screenshots when i was talking about it:
screen1
screen2

And yes, I can just remove it and show nothing, but why not give more info to the player if possible?

src/main/java/net/dries007/tfc/world/classic/worldgen/vein/VeinRegistry.java
DisasterMoo: Not true. I found in my tests that if you specify something invalid (ie: `tfc:native_copper2`) the vein is parsed anyway containing air blocks. I can fix that in `VeinTypeJson` instead, but that will still prevent air blocks being registered: 
https://github.com/DisasterMoo/TerraFirmaCraft/blob/ef7fb80e6e89f5c618d761e637c5712b78170d73/src/main/java/net/dries007/tfc/util/json/VeinTypeJson.java#L111-L115

src/main/java/net/dries007/tfc/command/CommandVeinInfo.java
DisasterMoo: added it since it is referenced by `findveins` command, but i guess we can also implement it there instead

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
DisasterMoo: Sure thing. (was just a copy-paste from forge, and since i wasn't too sure if we would ever generate veins in other dimensions past 1.15, I left there)

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
DisasterMoo: It was to free memory as soon as i needn't the data anymore, but after all rewrites i did there i think it can now manage a simple `List<String>`

src/main/java/net/dries007/tfc/world/classic/worldgen/vein/VeinType.java
alcatrazEscapee: From discord:
> but what does "Rare" mean

This is exactly the point. You aren't showing more information - you're just reducing the resolution of the rarity. You could well display "Rarity: 3%", (taking `1 / rarity`) rather than a value, but "Rare", or any word you choose, is obscuring meaning and also, *completely dependent on a specific set of veins*.

src/main/java/net/dries007/tfc/command/CommandVeinInfo.java
alcatrazEscapee: Ah, then it should be removed there. That was likely leftover copy-paste when I first wrote that.

src/main/java/net/dries007/tfc/util/json/VeinTypeJson.java
```diff
@@ -68,6 +86,49 @@ public VeinType deserialize(JsonElement json, Type typeOfT, JsonDeserializationC
...
+                    Block looseBlock = ForgeRegistries.BLOCKS.getValue(looseResource);
+                    // Looks like forge parse anything to air if nothing is found
+                    if (looseBlock == null || looseBlock == Blocks.AIR)
+                    {
+                        throw new JsonParseException("Unable to parse loose rock " + looseResource + ". No registered small ore, item or block found.");
+                    }
+                    else
+                    {
+                        looseRock = new ItemStack(looseBlock, 1, metaLoose);
+                    }
```
alcatrazEscapee: This whole sequence is unused, because you construct an `ItemStack` anyway, which will either:
a. convert the `Block` to it's respective `Item` (which, if it exists, would've been caught from the item query above), or
b. if the block doesn't exist, it will return an empty item stack.

src/main/java/net/dries007/tfc/world/classic/worldgen/vein/VeinRegistry.java
alcatrazEscapee: Well that's a nuisance. In which case, [here](https://github.com/DisasterMoo/TerraFirmaCraft/blob/ef7fb80e6e89f5c618d761e637c5712b78170d73/src/main/java/net/dries007/tfc/util/json/VeinTypeJson.java#L136), switch the ordering of that call chain to:
```java
if (ForgeRegistries.BLOCKS.containsKey(oreName)) {
    Block block = ForgeRegistries.BLOCKS.getValue(oreName); // Blocks have a nonnull default key, we can ignore null checks on this
    ...
```
as that should more accurately handle this case:
 - no arbitrary restrictions on the usage
 - accurate error reporting if something invalid was specified

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
DisasterMoo: Changed to using the VeinType obj directly, i bet that it's better than using string comparison

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -99,7 +131,244 @@ public int getRequiredPermissionLevel()
...
+                    }
+
+                }
+                else if (type == 2)
+                {
+                    final String fileName = "tfc-veins-rate.log";
+                    try
+                    {
+                        final File file = new File(fileName);
+                        final BufferedWriter writer = new BufferedWriter(new FileWriter(file));
```
alcatrazEscapee:
```suggestion
                    try(BufferedWriter writer = new BufferedWriter(new FileWriter(file))
                    {
```

src/main/java/net/dries007/tfc/command/CommandFindVeins.java
```diff
@@ -99,7 +131,244 @@ public int getRequiredPermissionLevel()
...
+
+                        for (Rock rock : rockRateMap.keySet())
+                        {
+                            String line = String.format("%s: %d", rock, rockRateMap.get(rock));
+                            writer.write(line);
+                            writer.newLine();
+                        }
+
+                        writer.close();
+
```
alcatrazEscapee:
```suggestion
```

src/main/java/net/dries007/tfc/world/classic/worldgen/WorldGenOreVeins.java
```diff
@@ -97,9 +99,19 @@ public void generate(Random random, int chunkX, int chunkZ, World world, IChunkG
...
+                if (generated)
+                {
+                    chunkData.markVeinGenerated(vein);
+                }
+                else if (ConfigTFC.GENERAL.debug)
+                {
+                    // Failed to generate, debug info
+                    // This can be by a number of factors, mainly because at each expected replacing position we didn't find a matching raw rock.
+                    // Some possible causes: Width / Height / Shape / Density / Y / Rock Layer
+                    TerraFirmaCraft.getLog().warn("Failed to generate vein '{}' in chunk ({}, {}). Vein center pos ({}x, {}y, {}z)", vein.getType().getRegistryName(), chunkX, chunkZ, vein.getPos().getX(), vein.getPos().getY(), vein.getPos().getZ());
```
alcatrazEscapee:
```suggestion
                    TerraFirmaCraft.getLog().debug("Failed to generate vein '{}' in chunk ({}, {}). Vein center pos ({}x, {}y, {}z)", vein.getType().getRegistryName(), chunkX, chunkZ, vein.getPos().getX(), vein.getPos().getY(), vein.getPos().getZ());
```

src/main/java/net/dries007/tfc/world/classic/worldgen/vein/VeinRegistry.java
```diff
@@ -141,6 +142,12 @@ public void postInit()
 
                         veinTypeRegistry.put(properVeinName, vein);
                         weightedVeinTypes.add(vein.getWeight(), vein);
+
+                        // Debug info
+                        if (ConfigTFC.GENERAL.debug)
+                        {
+                            TerraFirmaCraft.getLog().info("Registered new vein " + vein.toString());
+                        }
```
alcatrazEscapee:
```suggestion
                            TerraFirmaCraft.getLog().info("Registered new vein " + vein.toString());
```

src/main/resources/assets/tfc/lang/en_us.lang
```diff
@@ -104,6 +109,12 @@ tfc.tooltip.time_command_set_calendar_time=Set Calendar Time to %s.
 tfc.tooltip.time_command_disabled=This command has been disabled by TFC. Use /timetfc instead!
 tfc.tooltip.skill_set=Set %s to level %.2f
 tfc.tooltip.skill_add=Added %.2f levels to %s
+tfc.tooltip.veins_command.dump_veins=Dumping all veins in specified radius, this is gonna take a while...
+tfc.tooltip.veins_command.rate_veins=Dumping vein rate in specified radius, this is gonna take a while...
```
alcatrazEscapee:
```suggestion
tfc.tooltip.veins_command.rate_veins=Dumping vein rates in specified radius, this is gonna take a while...
```

alcatrazEscapee (changes requested):
A couple nitpicks

PR #901, opened 2020-03-21
Quiver Implementation

Gaelmare:
Implements leather knapped quiver, storage of javelins and arrows, and resupply of arrows and javelins when thrown/shot and none of that item is available in the inventory.


src/main/java/net/dries007/tfc/api/util/IQuiverAmmo.java
```diff
@@ -0,0 +1,50 @@
...
+
+import net.minecraft.entity.player.InventoryPlayer;
+import net.minecraft.item.ItemStack;
+
+import net.minecraftforge.items.CapabilityItemHandler;
+import net.minecraftforge.items.IItemHandler;
+
+import net.dries007.tfc.objects.items.ItemQuiver;
+
+public interface IQuiverAmmo
```
alcatrazEscapee: So this interface is just a collection of two helper functions, neither of which do anything implementation specific. This does not need to exist, the helper functions can be folded into `ItemQuiver`, as they make more sense to be there.

src/main/java/net/dries007/tfc/api/util/IQuiverAmmo.java
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    playerInv.setInventorySlotContents(playerInv.currentItem, newJav);
+                }
+            }
+        }
+    }
+
+    default ItemStack findQuiver(InventoryPlayer playerInv)
+    {
+        if (playerInv.offHandInventory.get(0).getItem() instanceof ItemQuiver)
```
alcatrazEscapee: What is the point of only searching hotbar + offhand? Quiver is meant to be worn on your back, i.e. chestplate, not held.

src/main/java/net/dries007/tfc/objects/container/ContainerQuiver.java
```diff
@@ -0,0 +1,52 @@
...
+        this.itemIndex += 8;
+    }
+
+    @Override
+    public boolean isItemValid(int slot, @Nonnull ItemStack stack) //duplicated in ItemQuiver?
+    {
+        //noinspection ConstantConditions
+        return OreDictionaryHelper.doesStackMatchOre(stack, "javelin") ||
+               //OreDictionaryHelper.doesStackMatchOre(stack, "arrow") ||
+               stack.getItem().getRegistryName().getPath().endsWith("arrow"); // no oreDict for vanilla arrows
```
alcatrazEscapee: Never do this. Just add an ore dictionary entry for arrows.

src/main/java/net/dries007/tfc/objects/container/ContainerQuiver.java
```diff
@@ -0,0 +1,52 @@
...
+public class ContainerQuiver extends ContainerItemStack implements ISlotCallback
+{
+    public ContainerQuiver(InventoryPlayer playerInv, ItemStack stack)
+    {
+        super(playerInv, stack);
+        this.itemIndex += 8;
+    }
+
+    @Override
+    public boolean isItemValid(int slot, @Nonnull ItemStack stack) //duplicated in ItemQuiver?
```
alcatrazEscapee: Then instead of duplicating code, just delegate to `ItemQuiver`, i.e. how all Tile Entity containers work.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+        return new ActionResult<>(EnumActionResult.SUCCESS, stack);
+    }
+
+    @Override
+    @Nonnull
+    public String getTranslationKey(ItemStack stack)
+    {
+        return super.getTranslationKey(stack);
+    }
```
alcatrazEscapee:
```suggestion
```

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @Nullable
+    @Override
+    public ICapabilityProvider initCapabilities(ItemStack stack, @Nullable NBTTagCompound nbt)
+    {
+        return new QuiverCapability(nbt);
+    }
+
+    // Extends ItemStackHandler for ease of use. Duplicates most of ItemHeatHandler functionality
```
alcatrazEscapee: Excuse me?

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
```diff
@@ -0,0 +1,192 @@
...
+            for (ItemStack slot : super.stacks)
+            {
+                if (!slot.isEmpty())
+                {
+                    text.add(1, I18n.format(TerraFirmaCraft.MOD_ID + ".tooltip.small_vessel_item", slot.getCount(), slot.getItem().getItemStackDisplayName(slot)));
+                    hasContent = true;
+                }
+            }
+        }
+*/
```
alcatrazEscapee:
```suggestion
```
Don't leave large block commented code. Either make the change or leave it.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
```diff
@@ -0,0 +1,192 @@
...
+        public boolean isItemValid(int slot, @Nonnull ItemStack stack)
+        {
+            //noinspection ConstantConditions
+            return OreDictionaryHelper.doesStackMatchOre(stack, "javelin") ||
+                   //OreDictionaryHelper.doesStackMatchOre(stack, "arrow") ||
+                   stack.getItem().getRegistryName().getPath().endsWith("arrow"); // no oreDict for vanilla arrows
+        }
+
+        @Override
+        public NBTTagCompound serializeNBT()
```
alcatrazEscapee: Why have you changed the serialization + deserialization from `ItemStackHandler`

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @Nullable
+    @Override
+    public ICapabilityProvider initCapabilities(ItemStack stack, @Nullable NBTTagCompound nbt)
+    {
+        return new QuiverCapability(nbt);
+    }
+
+    // Extends ItemStackHandler for ease of use. Duplicates most of ItemHeatHandler functionality
+    public class QuiverCapability extends ItemStackHandler implements ICapabilityProvider
```
alcatrazEscapee: Prefer extending `ItemStackHandlerCallback` as that implements a few prerequisite methods that simply dealing with item validity checks, for instance.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
```diff
@@ -0,0 +1,192 @@
...
+            return hasCapability(capability, facing) ? (T) this : null;
+        }
+
+        @Override
+        public boolean isItemValid(int slot, @Nonnull ItemStack stack)
+        {
+            //noinspection ConstantConditions
+            return OreDictionaryHelper.doesStackMatchOre(stack, "javelin") ||
+                   //OreDictionaryHelper.doesStackMatchOre(stack, "arrow") ||
+                   stack.getItem().getRegistryName().getPath().endsWith("arrow"); // no oreDict for vanilla arrows
```
alcatrazEscapee: Never do this, just add the ore dictionary entry

alcatrazEscapee (changes requested):
Ultimately, I don't much see the point of this, and I don't think "because classic had it" is a good reason here. As it stands, this does one thing: makes it easier after you throw a javelin to throw another one by moving one from your inventory to the thrown slot.
 - The quiver hasn't been made into a piece of armor, which would make sense. Requiring it in hotbar / offhand seems like compensation due to wanting to wear a chestplate, which is reasonable, but unrealistic.
 - You haven't implemented this for arrows at all, nor would it help, because bows already search your entire inventory for arrows - this is not something I think we need to or should change, but it makes the quiver for bows kind of useless.

IMO, the best solution here is just to implement the restock mechanism as part of the javelin as a QoL feature.

(Note: To support arrows, we could registry replace the vanilla bow and override `findAmmo`, but that leads to problems with literally any other modded bow (not really acceptable).)

src/main/java/net/dries007/tfc/objects/container/ContainerQuiver.java
Gaelmare: I see your point. This was following after the duplicated code in ItemSmallVessel and ContainerSmallVessel. I at least said something about it. :/

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: This code was copied from net/dries007/tfc/objects/items/ceramics/ItemSmallVessel.java
Don't blame me. :P

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: Done.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: Was cleaning out leftovers from ItemSmallVessel, and missed these. Thanks.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: I didn't like this either. I guess it depends on whether we want the arrow functionality at all. Seems silly that Forge doesn't add an oredict for arrows. What if two mods both decide to do it?

alcatrazEscapee: Then the first one that gets called goes through and the second one will do a no-op. It's a non-issue.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: I'm not seeing why this would be done? isItemValid is already implemented, and the slotCallback stacksize isn't an issue? Not sure where the ISlotCallback would be implemented. On the ItemQuiver? Can't implement it on the Capability, because can't pass reference to this to the super constructor before have constructed self.
ItemSmallVessel is implemented the same way as ItemQuiver.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: copypasta at its finest. Ugh.

src/main/java/net/dries007/tfc/objects/container/ContainerQuiver.java
Gaelmare: The isItemValid code in ItemSmallVessel and ContainerSmallVessel disagree. DRY violation, and something's funky here. I don't yet see the right way to fix both of these messes yet.

src/main/java/net/dries007/tfc/api/util/IQuiverAmmo.java
Gaelmare: Can we call functions on an Item that we may not have an ItemStack for? The functions here search the inventory for a quiver ItemStack. Feels strange for the javelins to call functions on an item the player may not have a copy of. I think I'm missing something here.

Gaelmare:
Thank you very much for the detailed review. I still miss this item from Classic. What do you think about someone's suggestion to make this a Bauble?

I agree that making this only work for vanilla bows would make no sense. I'd really like to add the arrow functionality, but maybe the only way would be some sort of replacement/addition to ItemArrow? Now I'm dreaming.

Gaelmare:
> (Note: To support arrows, we could registry replace the vanilla bow and override `findAmmo`, but that leads to problems with literally any other modded bow (not really acceptable).)

After looking through the code here, the onArrowLoose event is actually fired after the findAmmo call. The wrinkle is that arrow stacks coming out of the quiver would have to go into an empty slot, not into the slot that's being fired from as the stack in that slot is deleted at the end of onPlayerStoppedUsing. I think upon catching this event, checking that the stack size was 1 and that it wasn't infinite, creative mode, etc, we could resupply ammo from the quiver to an empty slot in the player inventory. If adding this feature makes this more likely to result in accepting this PR, I'll give it a shot. What do you think?

src/main/java/net/dries007/tfc/objects/container/ContainerQuiver.java
Gaelmare: ItemBow uses instanceof ItemArrow. I'll use that.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: fixed via instanceof

src/main/java/net/dries007/tfc/objects/container/ContainerQuiver.java
Gaelmare: I'm doing the same thing here as in ContainerSmallVessel and ItemSmallVessel. I'd like to see a fix for all of these. Is it as simple as return ItemSmallVessel.isItemValid(args)?

src/main/java/net/dries007/tfc/api/util/IQuiverAmmo.java
Gaelmare: done

src/main/java/net/dries007/tfc/objects/container/ContainerQuiver.java
Gaelmare: I've implemented a delegation here, and would love your feedback.

src/main/java/net/dries007/tfc/api/util/IQuiverAmmo.java
Gaelmare: I've changed it so it searches Armor and Main inventories.

Gaelmare:
This is ready for another review, but still needs a player model for the quiver. It shows up as a leather chestplate for now.

I have implemented arrow refills and equipable quiver!

From your earlier review, only open question I have is the ItemStackHandlerCallback.
ItemSmallVessel isn't implemented this way, and I'm not clear where I'd implement this.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: As discussed in discord, implemented ISlotCallback in QuiverCapability, mirroring SmallVessel.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public Weight getWeight(ItemStack stack)
+    {
+        return Weight.MEDIUM;
+    }
+
+    static QuiverCapability findQuiver(InventoryPlayer playerInv)
```
alcatrazEscapee, line 77: This should only check chestplate slot, no? Not scan the entire inventory?

alcatrazEscapee (approved):
One question, otherwise looks good.

src/main/java/net/dries007/tfc/objects/items/ItemQuiver.java
Gaelmare: Do you feel it's too OP for it to work anywhere else? If so, sure. Whole inventory may be too much, but would think chestplate slot or hotbar?

Gaelmare: Going to make it a config option, cause we don't have enough.

Gaelmare: I believe this is ready for merge

Issue #1086, opened 2020-04-23
Food combination recipes ignore traits


alcatrazEscapee:
Split off from #1066 because it's a separate issue.

Description:
 - When combining food items with different traits in a crafting grid, the rule of always preserving the soonest expiry date is violated, and the result does not correctly take into account traits present on the inputs.
 - [Example](https://clips.twitch.tv/AthleticRockyPelicanNotATK)

The section in question which does not take into account `FoodTrait`s:
https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/30b05fe351a82d2c7493a1436dd16df5b585453e/src/main/java/net/dries007/tfc/objects/recipes/FoodStackingRecipe.java#L61

I believe the current mechanic is not going to take into account traits at all. This is a bit of a delicate (mathematically) operation. The only possible result I can see, would be:
   - First, collect all traits that are present among all input items.
   - Then, copy all input items (avoid mutating the existing items in the crafting grid), and strip all traits that aren't present among all items, reducing or changing the expiration date accordingly.
    - Finally, perform the existing combination of items, now that each one should have the same traits (including the same decay modifier, which is important for choosing a minimum creation date)

Gaelmare:
Or, just refuse to stack items with different traits? That's what I was expecting to occur.

alcatrazEscapee:
Yeah, that works as well and is probably easier to do and would be more intuitive.

alcatrazEscapee:
b28f42bf356aeeba7c33a28c4c0f645bbde99852

PR #1089, opened 2020-04-23
Added food stacking function

### LINKED ISSUE #1026: QoL improvements for stacking food in inventory (former TFC Classic addon features)
I think this was in TFC Tweaks in Classic, a Dries addon. "s" was a hotkey in the inventory to stack all foods of the same type when pressed on one stack of that food. There was also a config option to stack food automatically when picked up.

Because stacking is probably even more of an issue in TNG, we should consider adding both of these this as core features.

 - TFC Version: build 129

TEXHIK:
fixes #1026

### DIFF
--- a/src/main/java/net/dries007/tfc/network/PacketStackFood.java
+++ b/src/main/java/net/dries007/tfc/network/PacketStackFood.java
@@ -1,0 +1,129 @@
+/*
+ * Work under Copyright. Licensed under the EUPL.
+ * See the project README.md and LICENSE.txt for more information.
+ */
+
+package net.dries007.tfc.network;
+
+import io.netty.buffer.ByteBuf;
+import net.dries007.tfc.TerraFirmaCraft;
+import net.dries007.tfc.api.capability.food.CapabilityFood;
+import net.dries007.tfc.api.capability.food.IFood;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.inventory.Slot;
+import net.minecraft.item.ItemStack;
+import net.minecraftforge.fml.common.network.simpleimpl.IMessage;
+import net.minecraftforge.fml.common.network.simpleimpl.IMessageHandler;
+import net.minecraftforge.fml.common.network.simpleimpl.MessageContext;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+
+public class PacketStackFood implements IMessage
+{
+    private int slotNumber;
+
+    public PacketStackFood()
+    {
+    }
+
+    public PacketStackFood(int slotNumber)
+    {
+        this.slotNumber = slotNumber;
+    }
+
+    public static final class Handler implements IMessageHandler<PacketStackFood, IMessage>
+    {
+        @Override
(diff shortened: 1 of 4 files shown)


src/main/java/net/dries007/tfc/network/PacketStackFood.java
```diff
@@ -0,0 +1,129 @@
...
+                        Slot slot = slotIterator.next();
+                        ItemStack stack = slot.getStack();
+                        IFood cap = stack.getCapability(CapabilityFood.CAPABILITY, null);
+
+                        if (cap == null || cap.isRotten()) continue;
+
+                        if (cap.getCreationDate() < minCreationDate)
+                        {
+                            minCreationDate = cap.getCreationDate();
+                        }
```
alcatrazEscapee, line 79: This suffers from the exact same problem as #1086. It needs to do a comparison on traits. Either stack ignoring traits (and then properly collecting common traits and stripping unused ones as per the description of #1086. Or, only stack foods that have the same traits. Might be worth adding a utility method for both cases, `CapabilityFood#areItemsStackable` which does exactly the checks here:
 - Are the foods both not rotten
 - Are the traits exactly the same

src/main/java/net/dries007/tfc/network/PacketStackFood.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+
+public class PacketStackFood implements IMessage
+{
+    private int slotNumber;
+
+    public PacketStackFood()
```
alcatrazEscapee, line 30:
```suggestion
    @SupressWarnings("unused")
    @Deprecated
    public PacketStackFood()
```
Helps avoiding IDE warnings for not using it (since it's called through forge internals), and accidental uses of it (since it's a common mistake to call this constructor instead of passing in the slot number)

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/network/PacketStackFood.java
TEXHIK: Rotten check I already have, but instacking logic. It's a good point to put that in capability, thanks.

alcatrazEscapee: As of b67eab6a018d6a74b1903bca2b545b11b7309c46 there are two functions added to `CapabilityFood` that may be used for this purpose:
 - `CapabilityFood#mergeItemStacksIgnoreCreationDate`, and
 - `CapabilityFood#areStacksStackableExceptCreationDate`.
The latter does not check rotten, as it completely ignores their creation date.

TEXHIK:
May be crafting recipe for stacking could be removed if we have this?

alcatrazEscapee:
I don't think it should.
 - It already exists, why remove it?
 - It offers more precision for combining different meats. Sometimes you want to combine several stacks, other times you'd want to keep stack separate to keep some food lasting longer. It represents the actual trade off of storing new stuff and older stuff in separate places, to avoid cross contamination.

alcatrazEscapee (approved):

PR #1104, opened 2020-04-26
Mass extinction of open issues. (Lots of bug and QoL fixes)

### LINKED ISSUE #436: Config option to disable the ability to place ingot stacks

### LINKED ISSUE #554: Javelin Thrown Position Should Match Throwing Hand
**What happens:** When throwing a javelin, if you're facing north, the thrown javelin throws from the right of the camera. If facing south, it throws from the left side of the camera. If facing east or west, it throws from above/center.

**What should happen:** Javelin should always be offset to the side matching the throwing hand. If held in main hand, throw from right of camera. If held in offhand, throw from left of camera. Reverse this behavior if Main Hand is set to Left in Skin Customization options.

**Meta Info**
 - TFC Version: 0.26.4.94
 - Occurs with/without other mods installed.

### LINKED ISSUE #707: Plucking feathers
This was a feature that existed in 1.7 that allowed for obtaining feathers from pheasants. I've tried right clicking and shift right clicking but it doesn't work, leading me to believe that it isn't implemented yet.

### LINKED ISSUE #759: Lighting Bug on Farmland with Fast Lighting
**Describe the bug**
When running TFC with fast lighting, farmland blocks look completely dark on the top of the block.

Fast Lighting:

Minimum/Maximum Lighting:

**To Reproduce**
In the Menu:
1. Go to Settings.
2. Click Video Settings.
3. Change Lighting Mode to Fast.

Afterwards, make some farmland with a hoe or visit already existing farmland.

**Meta Info**
 - TFC Version: 0.28.1.111
 - This bug was originally spotted when running Optifine and Phosphor, however further tests reproduced it with TFC alone, so neither of them is at fault.

### LINKED ISSUE #775: Jute texture is far too big and has massive z-fighting issues due to it.
TFC: 107

### LINKED ISSUE #799: Add config to disable crop withering
**Describe the request**
- When playing on a server with multiple people playing at different times crops will wither.

2. Describe the feature itself
- By Adding a simple true/false config to enable or disable the chance that a crop will wither would prevent this.

**Meta Info**
 - TFC Version: 0.29.0.115 (at time of request)
 - TFC Enhancement

### LINKED ISSUE #808: "Thirst" potion effect similar to "Hunger" potion effect.
**Describe the request**
Make sure to include:
A means of granting players a hunger like potion effect that affects players thirst bars.
Better thirst decrease when drinking salt water.
Useful for mod integration in modpacks with unclean water, could be used for TFC salt water jugs instead of suddenly removing water from their thirst bar. Would likely be an option part of the CraftTweaker integration too.

**Meta Info**
 - TFC Version: 1.12.2-0.29.0.115 (at time of request)

### LINKED ISSUE #918: Breaking Log with a hand
**Describe the bug**
When you break a log with your hand, it drop the log.

**To Reproduce**
just break a log.

**What Should happen?**
You should be able to break a log but it shouldn't drop a log. it should just drop nothing.

**Meta Info**
 - TFC Version: 0.30.1.122
- Forge version:14.23.5.2847

### LINKED ISSUE #933: Cloud visible trough some blocks
**Describe the bug**
Make sure to include:
1. What did you expect to happen? See the block
2. What actually happened instead (i.e. what was the bug). Clouds (and beams from journey map marker) show through some blocks, the specific one I found are workbench and any TFC liquid. I thought it could be some kind of transparency issue with the texture, but the bottom of the crafting table is just wood.
3. If they would help to explain the issue, include screenshots, sake, workbench and molten metal.

**To Reproduce**
1. place block and look at the sky.

**Meta Info**
 - TFC Version: 124, forge 2847
 - Were any other mods included? no, I removed everything else.

### LINKED ISSUE #939: Knives should destroy leaves faster
**Describe the request**
Make sure to include:
In classic, knives destroyed leaves faster than a bare fist or non-scythe item. This was useful on trees with a small number of leaves, as it allowed you to break the leaves faster without decreasing sapling chance. This is balanced by the fact that this eats through the durability of the knife very quickly.

**Meta Info**
 - TFC Version: 0.31.0.125 (at time of request)
 - TFC alone
 - No other mods needed

### LINKED ISSUE #946: Wooden buckets (and other items which use `FluidWhitelistHandler`) save fluid data to NBT causing different items not to stack.
**Describe the bug**
Make sure to include:
1. What did you expect to happen? all empty buckets to stack
2. What actually happened instead (i.e. what was the bug) empty bucket and empty bucket used in a recipe (dough) are not the same and do not stack
First is a newly crafted bucket, second is as bucket from a dough recipe.
image

**To Reproduce**
1. Do craft 2 buckets, use one to make dough, try to stack both, they don't.

**Meta Info**
 - TFC Version: 125

### LINKED ISSUE #969: Config to disable sapling drops from leaves.
**Describe the request**
Make sure to include:
Self explanatory, Classic did this, would be nice in TNG.

**Meta Info**
 - TFC Version: 0.31.1.126 (at time of request)

### LINKED ISSUE #971: You can still plant seeds on grass.
**Describe the bug**
Make sure to include:
No, just no.
Seeds can be planted on grass, which was a test feature before farmland was added.

**Meta Info**
 - TFC Version: 1.12.2-0.31.1.126

### LINKED ISSUE #988: Better indication for lit log piles
If the charcoal pit is to get any sort of configuration it should use another block.
Also if you play with Minimal particles the smoke is missing and you can't tell if a pit is lit. Changing the texture/block like Classic did would help with this.

### LINKED ISSUE #1036: Bad command fail message for /stripworld.
128

>:| You were supposed to fix all this....

/stripworld gives useless feedback on fail.

Plz no repeat in 1.15

### LINKED ISSUE #1038: Red apple sapling have banana sapling texture
**Describe the bug**
Make sure to include:
Red apple sapling have banana sapling texture. It is also the only tree I have not encountered in the wild. 

image

**Meta Info**
 - TFC Version: 0.X.131

### LINKED ISSUE #1050: Dynamic vaule tooltip for ores.
Currently we have Ore values on ores but perhaps should add a dynamic count of the current value of the whole stack to make things a bit easier to do math wise.

### LINKED ISSUE #1054: Fix firepit hitbox when things are attached.
Alc was lazy and didn't do this. Plz fix
TFC 132

### LINKED ISSUE #1067: Blue bed needs regular Dye recipe.
TFC 133
~Beds shouldn't use the dye recipes like the blue bed.~
Blue bed should get it's dye recipe back.

### LINKED ISSUE #1068: Add recipe to recycle unfired mold to 5 clay.
TFC 133
There isn't really a reason this shouldn't exist. Raw clay is Raw clay.

### LINKED ISSUE #1072: Rename Hot water to Spring water in the lang.
Too many people get this confused. Changing this should help people understand it better.

### LINKED ISSUE #1095: Allow users to pick up items from a placed item block if off hand holds something.
TFC 133
Title.

### LINKED ISSUE #1097: Fence gates don't line up with walls.
TFC 133
Fence gates need to be moved down by -.3 to line up with walls.

Claycorp:
- Fixed #436
- Fixed #554
- Fixed #707 
- Fixed #759 
- Fixed #775 (sorta, This is a deviation for the original idea of a scale down but people liked it.)
- Fixed #799 
- Fixed #808 
- Fixed #933 
- Fixed #918 
- Fixed #939 
- Fixed #946 (Other mods may cause this issue too, it was fucking annoying though. Should help Could be fixed with CT scripts in other places more than likely.)
- Fixed #969 (69 lamo)
- Fixed #971 
- Fixed #988 
- Fixed #1072 
- Fixed #1036 
- Fixed #1068 
- Fixed #1038 
- Fixed #1050 
- Fixed #1054 
- Fixed #1067 
- Fixed #1097 
- Fixed #1095 

Misc non-issue changes.
- Javelins are infinite for creative users.
- Improved thrown javelin rotation.
- Added XZ deviations to all saplings.
- Metal Buckets don't empty in creative anymore.
- Barrels will not hold the contents they had anymore when sealed unless it's a valid recipe. (You can't hide the cheese from me...)
- Campfires aren't wusses anymore and won't disappear from a simple slap.


src/main/java/net/dries007/tfc/command/CommandStripWorld.java
```diff
@@ -40,13 +40,13 @@ public String getName()
...
     {
-        return "/stripworld <radius> -> [DANGER] Strips all stone variant blocks + water.";
+        return "/stripworld <radiusInBlocks> -> [DANGER] Strips all stone variant blocks + water.";
     }
 
     @Override
     public void execute(MinecraftServer server, ICommandSender sender, String[] args) throws CommandException
     {
-        if (args.length != 1) throw new WrongUsageException("1 argument required.");
+        if (args.length != 1) throw new WrongUsageException("/stripworld <radiusInBlocks>");
```
DisasterMoo: Please use lang entries, refer to `CommandFindVeins` and it's lang entries if needed.

src/main/java/net/dries007/tfc/objects/potioneffects/PotionThirst.java
```diff
@@ -0,0 +1,50 @@
...
+            float thirst = foodStatsTFC.getThirst();
+
+            foodStatsTFC.setThirst(thirst - 0.05F * (float) (amplifier + 1));
+        }
+    }
+
+    @Override
+    public boolean isReady(int duration, int amplifier)
+    {
+        return true;
```
DisasterMoo, line 42: Minor nitpick: I think it would be better to do like hunger and drains the thirst over time instead of instantly.

src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -807,6 +810,30 @@ public static void pickupQuiverItems(EntityItemPickupEvent event) //only pickups
...
+            if (!event.getTarget().getEntityWorld().isRemote && player.getHeldItemMainhand().isEmpty() && player.isSneaking())
+            {
+                if (target.hurtResistantTime == 0)
+                {
+                    target.dropItem(Items.FEATHER, 1);
+                    target.attackEntityFrom(PLUCKING, (float) ConfigTFC.GENERAL.damagePerFeather);
+                }
+                if (target instanceof EntityAnimalTFC)
+                {
+                    ((EntityAnimalTFC) target).setFamiliarity(((EntityAnimalTFC) target).getFamiliarity() - 0.4f);
```
DisasterMoo: If this is the intention you can mark this as resolved, but 0.4f = 40% of the total familiarity. Shouldn't it be 4% instead?

src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -475,6 +501,16 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
...
         @Config.LangKey("config." + MOD_ID + ".client.hideThirstBar")
         public boolean hideThirstBar = false;
+
+        @Config.Comment({"Ore tooltip info mode.",
+            "0 = No tooltip info.",
+            "1 = Ore Unit only info.",
+            "2 = Stack unit total info.",
+            "3 = All info."
+        })
+        @Config.LangKey("config." + MOD_ID + ".general.oreTooltipMode")
```
DisasterMoo, line 511: Missing lang entry

src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -426,6 +426,32 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
...
+        public boolean shouldCropsDie = true;
+
+        @Config.Comment("If false, logs will drop logs when punched.")
+        @Config.LangKey("config." + MOD_ID + ".general.doLogsRequireAxe")
+        public boolean doLogsRequireAxe = true;
+
+        @Config.Comment("If false, leaves will not drop saplings.")
+        @Config.LangKey("config." + MOD_ID + ".general.doLeavesDropSaplings")
+        public boolean doLeavesDropSaplings = true;
+
```
DisasterMoo, line 454: Missing lang entries

DisasterMoo (changes requested):
Need a couple changes, most notably the missing lang entries, but otherwise seems good.

src/main/java/net/dries007/tfc/objects/blocks/devices/BlockFirePit.java
```diff
@@ -56,6 +56,7 @@
 
     private static final AxisAlignedBB DEFAULT_AABB = new AxisAlignedBB(0, 0, 0, 1, 0.125, 1);
     private static final AxisAlignedBB FIREPIT_AABB = new AxisAlignedBB(0, 0, 0, 1, 0.125, 1);
+    private static final AxisAlignedBB FIREPIT_WITH_SHIT_AABB = new AxisAlignedBB(0, 0, 0, 1, 0.7, 1);
```
alcatrazEscapee:
```suggestion
    private static final AxisAlignedBB FIREPIT_WITH_ATTACHMENT_AABB = new AxisAlignedBB(0, 0, 0, 1, 0.7, 1);
```
Can we not?

src/main/java/net/dries007/tfc/objects/items/metal/ItemMetalBucket.java
```diff
@@ -124,6 +124,10 @@ else if (fluid.getBlock() instanceof BlockLiquid)
                             }
                         }
                         worldIn.playSound(null, pos, SoundEvents.ITEM_BUCKET_EMPTY, SoundCategory.PLAYERS, 1.0F, 1.0F);
+                        if (stack.hasTagCompound() && stack.getTagCompound().isEmpty())
+                        {
+                            stack.setTagCompound(null);
+                        }
```
alcatrazEscapee, line 130: So I understand the issue that this is trying to solve, but this is a really hack and slash way of getting around it and doesn't resolve the root issue. There are still plenty of situations where the NBT tag would still apply. As I mentioned in the issue, we need to tackle the root problem, namely, the situations in `FluidHandlerItemStackSimple` which saves the fluid to NBT, is not the same assumption that we operate under (that the fluid is never saved to NBT). For consistency with the rest of the mod we should go with the latter, but for simplicity it might be better to go with the former.

src/main/java/net/dries007/tfc/objects/items/wood/ItemWoodenBucket.java
```diff
@@ -146,6 +146,10 @@ else if (fluid.getBlock() instanceof BlockLiquid)
                             }
                         }
                         worldIn.playSound(null, pos, SoundEvents.ITEM_BUCKET_EMPTY, SoundCategory.PLAYERS, 1.0F, 1.0F);
+                        if (stack.hasTagCompound() && stack.getTagCompound().isEmpty())
+                        {
+                            stack.setTagCompound(null);
+                        }
```
alcatrazEscapee, line 152: See above.

src/main/java/net/dries007/tfc/objects/recipes/ShapelessFluidFoodRecipe.java
```diff
@@ -59,7 +59,12 @@ public ShapelessFluidFoodRecipe(ResourceLocation group, NonNullList<Ingredient>
             else
             {
                 handler.drain(Fluid.BUCKET_VOLUME, true);
-                ret.set(i, handler.getContainer().copy());
+                ItemStack updatedItem = handler.getContainer().copy();
+                if (updatedItem.hasTagCompound() && updatedItem.getTagCompound().isEmpty())
+                {
+                    updatedItem.setTagCompound(null);
+                }
+                ret.set(i, updatedItem);
```
alcatrazEscapee, line 67: See above.

src/main/java/net/dries007/tfc/objects/potioneffects/PotionThirst.java
Claycorp: This is exactly like hunger? It's a copy paste of the hunger debuff set with a higher default (one order higher) so you don't need a 40 modifier on it to notice the drop.

Edit: Helps if I read the thing you highlight.
Testing it gives the same effect as hunger. I don't know why this option even exists as it does nothing. The only thing that uses it is health.

src/main/java/net/dries007/tfc/command/CommandStripWorld.java
Claycorp: I just edited what was there, I can update that to match though.

src/main/java/net/dries007/tfc/CommonEventHandler.java
Claycorp: Correct. Bad conversion from classic math to %. Should be .04!

PR #1124, opened 2020-04-27
Lamp implementation

### LINKED ISSUE #722: lamps should be completed
TNG lamps are not yet functional. This should be fixed.

Gaelmare:
* attach to the top or bottom of blocks
* use same anvil recipe as classic and can be from any metal.
* lit with anything torches can and snuffed out by an empty hand or any tool
* respond to levers, so can use a lightswitch
* burn olive oil only by default
* hold 250mb, last for 2000 hours by default same as classic
* new lamp models use new "lit/unlit" lamp texture and current metal block textures
* new item textures to match
* blockstates are autogenerated
* use new fluidwhitelisthandlercomplex that handles partially full items (I can drop mine in Flasks!)
* refill from player held containers that can hold partial fluids if not lit
* Closes #722


src/main/java/net/dries007/tfc/objects/blocks/metal/BlockMetalLamp.java
```diff
@@ -0,0 +1,514 @@
...
+
+    private final Metal metal;
+
+    public static final PropertyDirection FACING = PropertyDirection.create("facing", new Predicate<EnumFacing>()
+    {
+        public boolean apply(@Nullable EnumFacing p_apply_1_)
+        {
+            return p_apply_1_ == EnumFacing.DOWN || p_apply_1_ == EnumFacing.UP;
+        }
+    });
```
alcatrazEscapee:
```suggestion
    public static final PropertyDirection FACING = PropertyEnum.create("facing", EnumFacing.class, EnumFacing.Plane.VERTICAL);
```

src/main/java/net/dries007/tfc/objects/blocks/metal/BlockMetalLamp.java
```diff
@@ -0,0 +1,514 @@
...
+                break;
+            case 3:
+                iblockstate = iblockstate.withProperty(FACING, EnumFacing.SOUTH);
+                break;
+            case 4:
+                iblockstate = iblockstate.withProperty(FACING, EnumFacing.NORTH);
+                break;
+            case 5:
+            default:
+                iblockstate = iblockstate.withProperty(FACING, EnumFacing.UP);
```
alcatrazEscapee: This is both redundant (as the facing is restricted to only UP / DOWN), and overly verbose. For example:
```java
return state.withProperty(FACING, EnumFacing.byIndex(meta % 2)).withProperty(LIT, meta >= 2);
```

src/main/java/net/dries007/tfc/objects/blocks/metal/BlockMetalLamp.java
```diff
@@ -0,0 +1,514 @@
...
+                break;
+            case NORTH:
+                i = i | 4;
+                break;
+            case UP:
+            default:
+                i = i | 5;
+        }
+
+        return i + (state.getValue(LIT) ? 6 : 0);
```
alcatrazEscapee: Same here
```java
return state.getValue(Facing).getIndex() + (state.getValue(LIT) ? 2 : 1);
```

src/main/java/net/dries007/tfc/objects/items/itemblock/ItemBlockMetalLamp.java
```diff
@@ -0,0 +1,167 @@
...
+import net.dries007.tfc.api.capability.metal.IMetalItem;
+import net.dries007.tfc.api.types.Metal;
+import net.dries007.tfc.objects.blocks.metal.BlockMetalLamp;
+import net.dries007.tfc.objects.fluids.capability.FluidWhitelistHandlerComplex;
+import net.dries007.tfc.objects.te.TELamp;
+import net.dries007.tfc.util.Helpers;
+
+public class ItemBlockMetalLamp extends ItemBlockTFC implements IMetalItem
+{
+    public final static int CAPACITY = TELamp.CAPACITY;
```
alcatrazEscapee: This should be a config option.

src/main/java/net/dries007/tfc/objects/te/TELamp.java
```diff
@@ -0,0 +1,127 @@
...
+import net.dries007.tfc.ConfigTFC;
+import net.dries007.tfc.objects.fluids.capability.FluidHandlerSided;
+import net.dries007.tfc.objects.fluids.capability.FluidTankCallback;
+import net.dries007.tfc.objects.fluids.capability.IFluidHandlerSidedCallback;
+import net.dries007.tfc.objects.fluids.capability.IFluidTankCallback;
+import net.dries007.tfc.objects.items.itemblock.ItemBlockMetalLamp;
+
+public class TELamp extends TETickCounter implements IFluidTankCallback, IFluidHandlerSidedCallback
+{
+    public final static int CAPACITY = ConfigTFC.GENERAL.metalLampCapacity;
```
alcatrazEscapee: This defeats the purpose of making it a config option. You need to query it from the config, not just statically reference it.

src/main/java/net/dries007/tfc/objects/te/TELamp.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private final FluidTank tank = new FluidTankCallback(this, 0, CAPACITY);
+    private boolean powered = false;
+
+    public TELamp()
+    {
+        super();
```
alcatrazEscapee:
```suggestion
```

src/main/java/net/dries007/tfc/objects/te/TELamp.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    public void readFromNBT(NBTTagCompound nbt)
+    {
+        super.readFromNBT(nbt);
+        tank.readFromNBT(nbt.getCompoundTag("tank"));
+        if (tank.getFluidAmount() > tank.getCapacity())
+        {
+            // Fix config changes
+            FluidStack fluidStack = tank.getFluid();
+            //noinspection ConstantConditions
```
alcatrazEscapee: This can be null, don't ignore this, null check instead.

alcatrazEscapee (changes requested):

alcatrazEscapee:
There's a couple of code style inconsistencies, could you import + reformat please.

src/main/java/net/dries007/tfc/objects/te/TELamp.java
Gaelmare: As with some of these, you're finding problems with code I copied verbatim from other places in TFC: TEBarrel#readFromItemTag and #readFromNBT
That's fun.

src/main/java/net/dries007/tfc/objects/blocks/metal/BlockMetalLamp.java
Gaelmare: I implemented this way in case horizontal/wall mount options do show up. It will be just a blockstate json file update then. Ok to leave it in this state, or should I remove the horizontal facings?

Gaelmare:
I thought I was careful with code formatting. I'll double check that too. Thanks for review!

src/main/java/net/dries007/tfc/objects/blocks/metal/BlockMetalLamp.java
alcatrazEscapee: This change does nothing except clean up your predicate. If it gets changed to horizontal, both places where I've suggested property changes will need to be made, both should still change now because they're overly verbose and can be written much simpler.

Gaelmare: If horizontal gets added, existing blocks in worlds will be broken, I believe?

alcatrazEscapee: No, they'll fix themselves.

src/main/java/net/dries007/tfc/objects/te/TELamp.java
Gaelmare: Fixed here and in the two places in TEBarrel

Gaelmare:
Code reformat done.

Gaelmare:
Ready for merge

alcatrazEscapee (approved):

PR #1128, opened 2020-04-28
Improve Passive AI

Gaelmare:
Question: Concerns about adding more Entities into EntityAIAvoidEntity scans? Since none of the predators share a common class, would have to attempt to avoid each one separately.
* Pulled most Livestock AI tasks into common class. More could be done.
* Added swimmingAI to cows
* Eating grass AI to all
* Tweaked distances and avoidance speeds for avoid AI
* Added avoid predators and eat grass to deer
* Added avoid wolves to pheasant, increased panic speed
* Added avoid wolves to rabbits!


src/main/java/net/dries007/tfc/api/types/ILivestockAI.java
```diff
@@ -0,0 +1,57 @@
...
+import net.dries007.tfc.objects.entity.animal.EntitySaberToothTFC;
+import net.dries007.tfc.objects.entity.animal.EntityWolfTFC;
+
+/**
+ * Use this to add AI Tasks that are the same between all TFC Livestock
+ * Does not apply to Vanilla derived mobs: Horse, Donkey, Mule, Llama, Camel
+ * Does apply to at least Pig, Cow, Chicken, Sheep, Goat, Duck, Alpaca,
+ * Used only in TFC worlds.
+ */
+public interface ILivestockAI extends ILivestock
```
DisasterMoo, line 35: This should go into an enum instead, and then you could implement there some "common behaviors" which could be called from `initEntityAI` ie: `commonQuadrupedTasks`, `avoidLargeCarnivores`, etc.

See how i did a similar thing in `AnimalGroupingRules`

DisasterMoo (changes requested):
I think you should change that interface for an enum, since it would make more sense to run tasks there (imo, interfaces should be used when you need `instaceof` checks someplace else).

src/main/java/net/dries007/tfc/api/types/ILivestockAI.java
Gaelmare: Thanks for review. Too bad IPredator can't be used by the Avoid Entity AI. :/
I'll look into making an Enum.

alcatrazEscapee: That, or, just a static method somewhere. There's no point to implementing this interface - it's just a helper method to add common tasks to an entity. imo an enum doesn't make sense here particularly (it would work, as would this), but there's reason to think that some entities would want different combinations of behaviors.

`EntitiesTFC.addCommonLivestockAI(Entity entity)` would work fine.

Gaelmare:
So since you're assigned, @DisasterMoo, why don't all the predators (leaving wolves out) inherit from a single class? Would be helpful, I think?

src/main/java/net/dries007/tfc/api/types/ILivestockAI.java
Gaelmare: Done. Also incorporated all other TFC specific prey animal AI (not vanilla derived entities).
Is it really ok to have 9 EntityAIAvoidEntity tasks for the all the predators, plus EntityPlayer, and EntityMob?

Gaelmare:
closed in favor of #1139 which doesn't have merge/rebase weirdness

PR #1130, opened 2020-04-28
Some entity field synchronization with the client for Waila

repo-alt:


alcatrazEscapee (changes requested):
Please import and use our code style.

repo-alt:
Oh, I thought Idea does that automatically. Ok.

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -473,4 +478,25 @@ public static ResourceLocation readResourceLocation(ByteBuf buf)
...
+            return new DataParameter(id, this);
+        }
+
+        public Long copyValue(Long value) {
+            return value;
+        }
+    };
+    static {
+        DataSerializers.registerSerializer(LONG_DATA_SERIALIZER);
+    }
```
DisasterMoo: Seems this is still missing code style (your `{}` is't matching). Also, there's already an static initializer above, just use that. Lastly, please put static variables with the rest (`JEI`, `JOINER_DOT`) at the start of this class.

DisasterMoo (changes requested):
One thing, see below

src/main/java/net/dries007/tfc/util/Helpers.java
repo-alt: Ok. Btw, is it a good place for this stuff? Didn't find where else to put it
I added second static{} because it works only on previous declarations, and I didn't want to put new stuff at the start. I believe they are combined at compile time anyway

repo-alt: Done.

DisasterMoo: It's fine to be at `Helpers`, no need to create a class for this.

DisasterMoo (approved):

PR #1146, opened 2020-05-01
Add Waila info for TFC blocks/entities

AnodeCathode:
Adds most blocks and animal support. Requires PR https://github.com/TerraFirmaCraft/TerraFirmaCraft/pull/1130 for animal support.
Optional Air Temperature display in waila.cfg for BlockRockVariant blocks.
Blast furnace and bloomery, charcoal forge tbd, may require TFC client/data sync changes.


HeartyYF:
And also, can you please add Waila info for stalactite?

AnodeCathode:
> 
> 
> And also, can you please add Waila info for stalactite?

Not sure what value this would add. What info would you look to receive from a stalactite? It doesn't show up in Waila because it doesn't have an item. It doesn't have an item because it's purely decorative. If it had some use beyond being a pretty decoration in caves then I'd support this. But as fixing this would require giving the stalactite a pickblock return, it needs a case for why.

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,502 @@
...
+    @Override
+    public List<String> getWailaBody(ItemStack itemStack, List<String> currenttip, IWailaDataAccessor accessor, IWailaConfigHandler config)
+    {
+        Block b = accessor.getBlock();
+        TileEntity te = accessor.getTileEntity();
+
+        // Mechanics
+        if (b instanceof BlockOreTFC) currenttip = OreTFCBody(itemStack, currenttip, accessor, config);
+        else if (te instanceof TEPlacedItemFlat)
+            currenttip = PlacedItemFlatBody(itemStack, currenttip, accessor, config);
```
alcatrazEscapee: Please expand this:
```suggestion
        if (b instanceof BlockOreTFC) 
        {
            return OreTFCBody(itemStack, currenttip, accessor, config);
        }
        else if (te instanceof TEPlacedItemFlat)
        {
            return PlacedItemFlatBody(itemStack, currenttip, accessor, config);
        }
```
etc.

I'm not a fan of this giant if-else statement but if we were to change it I'm not sure it would be much better. So it stays.

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,502 @@
...
+        if (b instanceof BlockCropDead) currenttip = CropDeadHead(itemStack, currenttip, accessor, config);
+        if (b instanceof BlockFruitTreeTrunk | b instanceof BlockFruitTreeLeaves | b instanceof BlockFruitTreeBranch)
+        {
+            currenttip = FruitTreeHead(itemStack, currenttip, accessor, config);
+        }
+        return currenttip;
+    }
+
+    @Override
+    public List<String> getWailaBody(ItemStack itemStack, List<String> currenttip, IWailaDataAccessor accessor, IWailaConfigHandler config)
```
alcatrazEscapee:
```suggestion
    public List<String> getWailaBody(ItemStack stack, List<String> currentTooltip, IWailaDataAccessor accessor, IWailaConfigHandler config)
```
As per both java conventions on field / paramater naming, and MCP coding style conventions

DisasterMoo:
IIRC there is a missing lang entry for stalactites and stalagmites, which is probably what he means here. Don't bother, I will check and if needed fix that for the next build.

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,502 @@
...
+    public List<String> getWailaTail(ItemStack itemStack, List<String> currenttip, IWailaDataAccessor accessor, IWailaConfigHandler config)
+    {
+        if (config.getConfig("tfc.displayTemp"))
+        {
+            currenttip.add(new TextComponentTranslation("waila.temperature").getFormattedText() + " : " + String.valueOf(Math.round(ClimateTFC.getActualTemp(accessor.getWorld(), accessor.getPosition(), 0))) + new TextComponentTranslation("waila.tempsymbol").getFormattedText());
+        }
+        return currenttip;
+    }
+
+    public List<String> FirepitBody(ItemStack itemStack, List<String> currenttip, IWailaDataAccessor accessor, IWailaConfigHandler config)
```
alcatrazEscapee: A couple things that are common themes with all the below methods:

 - All these methods need to be changed to adhere to java naming conventions - `lowerCamelCase` for methods. In my opinion they should all be prefixed with `get`, i.e. `getFirepitBody`, `getPitKilnBody`, `getOreStack`, etc.
 - There is also no specific need to include `TFC` in the name, despite it being the name of the tile entity.
 - There should be a consistient access level between all of them, imo `private` makes the most sense.

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,502 @@
...
+    {
+        return currenttip;
+    }
+
+    public List<String> PitKilnBody(ItemStack itemStack, List<String> currenttip, IWailaDataAccessor accessor, IWailaConfigHandler config)
+    {
+        Block b = accessor.getBlock();
+        TEPitKiln te = (TEPitKiln) accessor.getTileEntity();
+        String key;
+        Boolean isLit = te.isLit();
```
alcatrazEscapee:
```suggestion
        boolean isLit = te.isLit();
```

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,502 @@
...
+        Long litTick = te.getLitTick();
```
alcatrazEscapee:
```suggestion
        long litTick = te.getLitTick();
```
Don't use boxed primitives unless nessecary.

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,502 @@
...
+    private ItemStack CropSimpleStack(IWailaDataAccessor accessor, IWailaConfigHandler config)
+    {
+
+        BlockCropTFC b = (BlockCropTFC) accessor.getBlock();
+        ICrop crop = b.getCrop();
+        IBlockState blockstate = accessor.getBlockState();
+        Integer curStage = blockstate.getValue(b.getStageProperty());
+        ItemStack foodDrop = crop.getFoodDrop(curStage);
+
+
```
alcatrazEscapee:
```suggestion
        BlockCropTFC b = (BlockCropTFC) accessor.getBlock();
        ICrop crop = b.getCrop();
        IBlockState blockstate = accessor.getBlockState();
        int curStage = blockstate.getValue(b.getStageProperty());
        ItemStack foodDrop = crop.getFoodDrop(curStage);
```
Some odd whitespace, and boxed primitives.

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,502 @@
...
+    {
+        currenttip.add(new TextComponentTranslation("waila.deadcrop").getFormattedText());
+        return currenttip;
+    }
+
+    private List<String> FarmlandTFCBody(ItemStack itemStack, List<String> currenttip, IWailaDataAccessor accessor, IWailaConfigHandler config)
+    {
+        //anything to do here?
+        return currenttip;
+    }
```
alcatrazEscapee: If not, this can be removed?

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,502 @@
...
+        currenttip.add(text.trim());
+
+        return currenttip;
+    }
+
+    private List<String> CropSpreadingBody(ItemStack itemStack, List<String> currenttip, IWailaDataAccessor accessor, IWailaConfigHandler config)
+    {
+        //Currently unused
+        return currenttip;
+    }
```
alcatrazEscapee: It's unused and effectively deprecated. This can be removed.

alcatrazEscapee (changes requested):
Mostly just code style issues. Actual content I have no issues with. Common ones:
 - methods should all be `lowerCamelCase`
 - don't use boxed primitives, i.e. favor `int` over `Integer` if possible.

And give this a pass through Intellij's code style reformat.

src/main/resources/assets/tfc/lang/en_us.lang
```diff
@@ -4443,3 +4443,45 @@ advancements.tfc.world.find_nickel.description=Find Garnierite Ore.
...
+waila.canshear=Shearable
+waila.fruit_trees.banana.name=Banana Tree
+waila.fruit_trees.cherry.name=Cherry Tree
+waila.fruit_trees.green_apple.name=Green Apple Tree
+waila.fruit_trees.lemon.name=Lemon Tree
+waila.fruit_trees.olive.name=Olive Tree
+waila.fruit_trees.orange.name=Orange Tree
+waila.fruit_trees.peach.name=Peach Tree
+waila.fruit_trees.plum.name=Plum Tree
+waila.fruit_trees.red_apple.name=Red Apple Tree
```
DisasterMoo: This probably can cause lang issues with other mods if we share the same lang key. To make all these more unique to TFC, please change all to `waila.tfc.<key>`

DisasterMoo (changes requested):
Lang issue, see below

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
AnodeCathode: TFC is in the name of getOreTFC as the blockname is BlockOreTFC. So for consistency with other methods I've included it there.

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
AnodeCathode: done

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
AnodeCathode: Other suggestions have been commited.

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
AnodeCathode: Yeah I looked at a number of other methods to resolve. Maps and enums and various OO pretty but awkward solutions.. In the end it really is the marginally best of a set of choices.

AnodeCathode:
> 
> 
> IIRC there is a missing lang entry for stalactites and stalagmites, which is probably what he means here. Don't bother, I will check and if needed fix that for the next build.

Well I also think that because it does not return a pickblock, the waila hud will not even display for this block. Again, not sure if it really matters. But it's essentially the same issue that was fixed with the pitkiln and placed items. (Give it a pick block and also ensure it drops nothing but air)

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
```diff
@@ -0,0 +1,483 @@
...
+            currentTooltip = getBlastFurnaceBody(stack, currentTooltip, accessor, config);
+        else if (te instanceof TEFirePit) currentTooltip = getFirepitBody(stack, currentTooltip, accessor, config);
+            // Crops and Trees
+        else if (te instanceof TECropBase) currentTooltip = getCropSimpleBody(stack, currentTooltip, accessor, config);
+        else if (b instanceof BlockFruitTreeLeaves)
+            currentTooltip = getFruitTreeLeavesBody(stack, currentTooltip, accessor, config);
+        else if (b instanceof BlockBerryBush) currentTooltip = getBerryBushBody(stack, currentTooltip, accessor, config);
+        else if (b instanceof BlockCropDead) currentTooltip = getCropDeadBody(stack, currentTooltip, accessor, config);
+        else if (b instanceof BlockFruitTreeLeaves)
+            currentTooltip = getFruitTreeLeavesBody(stack, currentTooltip, accessor, config);
```
DisasterMoo: Code style issue, which makes this harder to read and maintain later. (also, Alc already said this earlier)
Use this
```java
if (b instanceof BlockOreTFC)
{
    // do something
}
else if (b instanceof TEPlacedItemFlat)
{
    // do something else
}
else if...
```
instead.

But the optimal solution here is to code an interface here and put `getWailaBody()` and `getWailaHead()` there, then you just need to create classes for each `TileEntity` and/or `Block` that you made it does something when looked at and register everything in a `List<Interface> listWaila`, then the solution would be as clean as:
```java
    public List<String> getWailaBody(ItemStack stack, List<String> currentTooltip, IWailaDataAccessor accessor, IWailaConfigHandler config)
    {
        Block block = accessor.getBlock();
        TileEntity te = accessor.getTileEntity();
        for(<Interface> wailaInterface : listWaila)
        {
            if (waila.isLookingAt(b, te))
            {
                return waila.getBody()
            }
        }
    }

File: src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java [hunk 11]
```diff
@@ -0,0 +1,483 @@
...
+
+        if (b instanceof BlockOreTFC)
+        {
+            stack = getOreTFCStack(accessor, config);
+        }
+        else if (b instanceof BlockCropSimple)
+        {
+            stack = getCropSimpleStack(accessor, config);
+        }
+        return stack;
```
DisasterMoo, line 96: If you do the other solution i mentioned it also applies here: You can code in that new interface a `getWailaStack()` there and instead of a lot of if-else, a for loop.

DisasterMoo (changes requested):
Still has issues, see below

src/main/java/net/dries007/tfc/compat/wailatop/WailaBlocks.java
AnodeCathode: I took option A. B may be optimal and if someone is eager to do it, please feel free

PR #1170, opened 2020-05-04
Add TOP plugin info

AnodeCathode:
Slightly different implementation to Waila as TOP doesn't support annotations. Otherwise follows the setup of the Waila impl.

CTRL+C / CTRL+V to the rescue

### DIFF
--- a/src/main/java/net/dries007/tfc/compat/top/BlastFurnaceProvider.java
+++ b/src/main/java/net/dries007/tfc/compat/top/BlastFurnaceProvider.java
@@ -1,0 +1,68 @@
+package net.dries007.tfc.compat.top;
+
+
+import java.util.function.Function;
+import javax.annotation.Nullable;
+
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.tileentity.TileEntity;
+import net.minecraft.util.text.TextComponentTranslation;
+import net.minecraft.world.World;
+
+import mcjty.theoneprobe.api.*;
+import net.dries007.tfc.api.capability.heat.Heat;
+import net.dries007.tfc.objects.blocks.devices.BlockBlastFurnace;
+import net.dries007.tfc.objects.te.TEBlastFurnace;
+import net.dries007.tfc.util.Helpers;
+
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+
+public class BlastFurnaceProvider implements Function<ITheOneProbe, Void>, IProbeInfoProvider
+{
+    public static ITheOneProbe probe;
+
+    @Override
+    public String getID()
+    {
+        return MOD_ID + ":top_blastfurnace_provider";
+    }
+
+    @Override
+    public void addProbeInfo(ProbeMode probeMode, IProbeInfo iProbeInfo, EntityPlayer entityPlayer, World world, IBlockState state, IProbeHitData iProbeHitData)
+    {
+        TileEntity te = world.getTileEntity(iProbeHitData.getPos());
+        if (state.getBlock() instanceof BlockBlastFurnace && te instanceof TEBlastFurnace)
+        {
+            TEBlastFurnace blastFurnace = Helpers.getTE(world, iProbeHitData.getPos(), TEBlastFurnace.class);
+            int chimney = BlockBlastFurnace.getChimneyLevels(world, iProbeHitData.getPos());
+            if (chimney > 0)
+            {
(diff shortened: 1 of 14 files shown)


src/main/java/net/dries007/tfc/compat/top/TOPPlugin.java
```diff
@@ -0,0 +1,63 @@
...
+    @Mod.EventHandler
+    public void fmlPreInitEvent(final FMLPreInitializationEvent event)
+    {
+        if (Loader.isModLoaded(MODID_TOP))
+        {
+            RegisterProviders(event);
+        }
+
+
+    }
```
DisasterMoo, line 41: Instead of duplicating pre init event making our jar contains two mods, just pass on the preInit event to this class if TOP mod is present (this is necessary since there's no annotation method in TOP). Also, your `@Optional` annotation will be unnecessary after this change.

DisasterMoo (changes requested):
Two things:
- Don't duplicate `@Mod` annotation, that will make our jar contains two mods. Use `TerraFirmaCraft#preInit()` and check if TOP is loaded there before passing on the registration to `TOPPlugin`
- Centralize the code that does "waila" stuff: Create a class for every provider with header / body / tail or anything more that is needed with common code there (you may use waila ones, just remove waila specific code there and create these specifics in another class). Functions should receive the `TileEntity`, `Block`, maybe `World` and `BlockPos`, anything that would be needed to work and returns a `List<String>` containing the tooltips that you may convert it to HWYLA or TOP in it's specific class. 

You may think it's a PITA to do the 2nd one but it will make our life maintaining this code later much easier (just one place to change instead of 2 or more. Animals is a "good" bad example of this since I always have to copy paste things there. I can't properly extend vanilla and use our common code in one place since it needs variables saved in the entity's NBT).

src/main/java/net/dries007/tfc/compat/top/BlastFurnaceProvider.java
```diff
@@ -0,0 +1,68 @@
...
+        if (state.getBlock() instanceof BlockBlastFurnace && te instanceof TEBlastFurnace)
+        {
+            TEBlastFurnace blastFurnace = Helpers.getTE(world, iProbeHitData.getPos(), TEBlastFurnace.class);
+            int chimney = BlockBlastFurnace.getChimneyLevels(world, iProbeHitData.getPos());
+            if (chimney > 0)
+            {
+                int maxItems = chimney * 4;
+                int oreStacks = blastFurnace.getOreStacks().size();
+                int fuelStacks = blastFurnace.getFuelStacks().size();
+                int temperature = blastFurnace.getField(TEBlastFurnace.FIELD_TEMPERATURE);
```
alcatrazEscapee, line 44: Two things need to be modified here:
 - We do not sync fields by default when the container is not open, meaning this will suffer from a desync. There are two things needed to resolve this:
   - If TOP has a sync mechanism for TEs that it's looking at, we need to exploit that, or otherwise allow that to sync
   - Based on what sync is happening, this may need to be changed to use other data (for instance, the NBT data directly, see how walia handles it)

src/main/java/net/dries007/tfc/compat/top/CrucibleProvider.java
```diff
@@ -0,0 +1,60 @@
...
+        TileEntity te = world.getTileEntity(iProbeHitData.getPos());
+        if (te instanceof TECrucible)
+        {
+            TECrucible crucible = Helpers.getTE(world, iProbeHitData.getPos(), TECrucible.class);
+            if (crucible.getAlloy().getAmount() > 0)
+            {
+                Metal metal = crucible.getAlloyResult();
+                TOPPlugin.outputHorizontalText(iProbeInfo, new TextComponentTranslation("waila.tfc.metal.output", crucible.getAlloy().getAmount(), new TextComponentTranslation(metal.getTranslationKey()).getFormattedText()).getFormattedText());
+            }
+            int temperature = crucible.getField(TECrucible.FIELD_TEMPERATURE);
```
alcatrazEscapee, line 43: Same issue as blast furnace.

alcatrazEscapee (changes requested):

alcatrazEscapee:
> You may think it's a PITA to do the 2nd one but it will make our life maintaining this code later much easier (just one place to change instead of 2 or more.

I'm not sure how simple this will be to do, given the API hooks for both cases have potential to be radically different. So if this can be done, great. Otherwise I'm fine with copy paste on this one.

DisasterMoo:
Closing this in favor of #1209

PR #1180, opened 2020-05-05
Synchronization standardization

### LINKED ISSUE #1179: Trying to put food in Salad GUI instantly crashes the client
**Describe the bug**
- not crash when making a salad
- crashed
3. If it was a crash, upload the full log file (`latest.log`)
- https://docs.google.com/document/d/1WHYrhOz3z8qLynLGbvcV-jLMOoKGio51cVbELbvXrZE/edit?usp=sharing

**To Reproduce**
- just try and create any salad

**Meta Info**
 - TFC Version: tested 138 and 139
-  no other mods

alcatrazEscapee:
- Fixes #1179 
 - Code cleanup for all tile entities, removes unnecessary packets and block updates, establishes a synchronization standard for future use:
   - No client side simulation to compensate for lackluster syncing
   - TE sync using `TEBase#markForSync` (doesn't cause block updates), and `TEBase#markForBlockUpdate`)
   - Batched implementation for tickable tile entities in `TETickableBase` and `TETickableInventory`
   - Hwyla switched to use built in NBT request based synchronization for data that would typically be present in fields. There still may be more to do here.

Still should be done:
 - Any outstanding issues with tile entities? (I've tested as well as I can and not found anything major)
 - I noticed a number of TEs don't implement `ICalendarTickable` when they probably should. Some probably have logic for this already, others idk. I didn't look into it extensively.
 - A lot of the Hwyla stuff was done very shotgun. There's definite potential for this to be either overdone or missing certain stuff. The only one I'm confident is working correctly and not unnecessary is `CrucibleProvider`. Others I added on a guesswork basis


src/main/java/net/dries007/tfc/objects/te/TEBlastFurnace.java
```diff
@@ -231,10 +232,89 @@ public int getField(int index)
...
+            IBlockState state = world.getBlockState(pos);
+            if (state.getValue(LIT))
+            {
+                // Update bellows air
+                if (--airTicks <= 0)
+                {
+                    airTicks = 0;
+                }
+
+                if (!world.isRemote && --burnTicksLeft <= 0)
```
DisasterMoo: cleanup (there's already a check for `world.isRemote` above)

src/main/java/net/dries007/tfc/objects/te/TEBloomery.java
```diff
@@ -124,15 +124,51 @@ public void onBreakBlock(World worldIn, BlockPos pos, IBlockState state)
...
+        checkForCalendarUpdate();
         if (!world.isRemote && --delayTimer <= 0)
         {
+            IBlockState state = world.getBlockState(pos);
+            if (state.getValue(LIT))
+            {
+                if (--burnTicksLeft <= 0)
+                {
+                    burnTicksLeft = 0;
+                    if (!world.isRemote)
```
DisasterMoo: Same as above

DisasterMoo (changes requested):
Outstanding job out there. 
TIL 2 things: 
- Vanilla has it's own `SPacketUpdateTileEntity` to update NBT data directly that we can override `TileEntity#getUpdatePacket` to only send sensitive data. Seems good to me since it would not clog network with a lot of block data / tick / player.
- Waila also has it's own way of asking server about updated data by registering to `registerNBTProvider`. I didn't took a look at their description, my bad.

Now, i only found 2 nitpicks, see below.

PR #1243, opened 2020-05-10
restrict berry bush / fruit tree spawning based on rainfall and temp

eerussianguy:
Changelog message for Moo to use:

- Fruit trees and bushes now will spawn only in areas where the rainfall and temperature allows them to, instead of all over the map.
    - Plums, oranges and bananas prefer the hotter environments
    - Lemons and olives spawn in hot as well as temperate environments.
    - Apples  like slightly colder and temperate environments, and prefer areas with less rainfall
    - Peaches will survive anywhere that isn't hot or rainy
    - Cherries extend the furthest into colder regions than any other fruit tree
    - Berries universally prefer colder weather. Wintergreen can even grow in negative temperatures
    - Elderberries, gooseberries, bunch berries, and strawberries are the only berries that are able to spawn in hotter regions

### DIFF
--- a/src/main/java/net/dries007/tfc/util/agriculture/BerryBush.java
+++ b/src/main/java/net/dries007/tfc/util/agriculture/BerryBush.java
@@ -18,17 +18,17 @@ import static net.dries007.tfc.api.types.IBerryBush.Size.*;
 
 public enum BerryBush implements IBerryBush
 {
-    BLACKBERRY(Food.BLACKBERRY, Month.MAY, 4, 5f, 25f, 125f, 400f, 0.8f, LARGE, true),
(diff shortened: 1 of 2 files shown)


eerussianguy:
OK it looks like the berry portion of the spreadsheet doesn't actually concatenate properly, hold off accepting until I get that fixed

eerussianguy:
all good to go

nuuno2000:
Will this only change the spawning to be dependent on temp/rainfall or also change the fruiting to be dependant on those factors, as right now all fruit trees/bushes need the sme rainfall/temp to fuit

Azaysol:
@nuuno2000 by the looks of the code i would say it also edits the growths of fruits after your replant them

DisasterMoo (approved):

PR #1263, opened 2020-05-13
More fixes

### LINKED ISSUE #342: Crops don't take into account farmland moisture,
Farmland moisture could probably be averaged with rainfall to find a medium between strict growth conditions and manual watering. Farmland moisture calculation should be adjusted to take this into account - be 30 / 70 influenced by area vs surrounding water. Crops should then only check the farmland moisture instead

Crops also are missing nutrients checks - this needs to be part of larger changes to farmland. Nutrients was discussed as being saved to chunk data rather than part of farmland. Possible avenues for replenishment via grass growing, or random ticks on fallow farmland blocks? (Note: this would largely eliminate the need to rotate crops around, rather just have mixed farms in a chunk)

### LINKED ISSUE #888: Add CT integration to bloomery and blast furnace recipes
Title says all (missing from #756).

### LINKED ISSUE #907: JEI "show recipes" is displayed above the output slot rather than on the "recipe/processing" arrow.
**Describe the bug**
This is self explanatory. "show recipes" JEI button should be on the recipe/processing "arrow", like most mods, and Vanilla tend to have.
image

**Meta Info**
 - TFC Version: 1.12.2-29.5.120

### LINKED ISSUE #942: Hot liquids can be stored in barrels.
Hot liquids, such as lava and molten metals can be stored in barrels. I expected the barrel to either burn or not allow the fluid, but it did. 

**To Reproduce**
1.  put lava bucket (vanilla or blue steel) into the input slot of the barrel.
2. the lava will appear in the barrel, this happens with molten metals as well.

**Meta Info**
 - TFC Version: 0.30.3.124
 - Other mods were included, but I don't believe any of them caused this issue

### LINKED ISSUE #996: Fix food poison saying death is from magic.
TFC 128

Dying to food poison = killed by magic. The damage source should be changed to something else.

### LINKED ISSUE #1017: TFCT: Hitboxes on quern do not show when looking down on it
If you look down from too high an angle at a quern, the nice hitboxes disappear.

**Meta Info**
 - TFC Version: build 129

### LINKED ISSUE #1018: TFCT: JEI does not show salting meat recipes
When JEI brings up the uses of salt powder (tfc:powder/salt) only the fresh->salt water recipe appears.

**To Reproduce**
1. Search for salt in JEI
2. press u to see usages
3. Expect to see salted meat, but only see saltwater recipe.

 - TFC Version: build 129

### LINKED ISSUE #1022: Natural predators should hunt passive mobs when aggro
All predators should hunt some passive mobs when aggro like wolves.

 - TFC Version: build 129

### LINKED ISSUE #1060: [Request] Config option for breaking speeds
So, I understand the realism that it's hard to break stone and chop trees, but it's super grindy. For the users' sake, would it be possible to add a config option to customize the break speed of logs and stone? Just one number you could change and the different stone materials and pickaxe materials would scale to that number. Maybe something standard, like copper against granite/oak.

Thanks for your consideration.

### LINKED ISSUE #1073: Animal spawn config should be per-animal instead of per-interface
I was looking through the animal code the other day, and noticed a number of inconsistencies. Pheasants implement `ILivestock`, which means they cannot respawn, and should probably be huntable instead (this is likely due to copying the code from chicken). Initial spawning rarity (but not respawning, which is handled in worldgen) is set directly on every animal entity, but I feel it would be better to set it in the animal type interfaces, as all animals except wolves and ocelots (and technically pheasants due to the above bug) use the correct spawn rarity for their type (wolves and ocelots can still have it easily overridden). Also, since there are now two "tameable"/"pet" animals which hunt other animals (wolves and ocelots), you could maybe create a new interface to share between them instead of making them special exceptions to rules like spawning and interaction (and possibly make adding new similar animals easier). Alpaca are intended to be the "colder" variant for sheep, however, unlike sheep, they do not drop sheepskin upon being killed (this may very well be intentional, I just wanted to check). Lastly, polar bears currently use the default food of grain, unlike other predators, which use meat and/or fish (though I'm not sure if this affects anything at the moment, as you cannot feed predators).

 - TFC Version: 1.0.6.133

### LINKED ISSUE #1099: Change heart icon to heart with a red X over it for things that can't be tamed.
TFC 133
Title.
Red around the outside isn't clear enough it can't be tamed.

### LINKED ISSUE #1109: Sandwiches and salads rot in JEI
**Meta Info**
 - TFC Version: 1.12.2-1.0.6.133
image

### LINKED ISSUE #1148: Reintroduce gemQualityGEMNAME oreDict's for better mod integration.
**Describe the request**
image
At the moment TFC has no oreDicts for gems like this: ``gemExquisiteDiamond``, just ``gemExquisite``, unlike some other mods, adding this would allow better integration with said mods.
image

**Meta Info**
Yes.

### LINKED ISSUE #1154: Large Vessel does not pop off when unsupported.
Large vessels need a solid block below to be placed but don't pop off when this block is removed.

### LINKED ISSUE #1165: Lamps and Nest Boxes float if supporting blocks broken like #1154
These TFC blocks do not pop off if their supporting blocks are broken. See #1154

 - TFC Version: build 137 + dev

### LINKED ISSUE #1223: Differences with hwyla after update to 141
After the update to 141 hwyla doesnt display horses jump and running strength nor animals hp

**Meta Info**
 - TFC Version: 1.0.12.141

### LINKED ISSUE #1229: Animals going through fences
Sometime i find that one of my animals has gone through the fence is outside, im not sure how to replicate it becasue ive never seen it happen, ive always just found the animals outside the fenced area

**Meta Info**
 - TFC Version: 1.0.12.141

### LINKED ISSUE #1234: Fertilized eggs don't display fertilized tooltip
Title

**Meta Info**
 - TFC Version: 1.0.12.141

### LINKED ISSUE #1237: Support Beams
**Describe the request**
Make sure to include:
 - Support beams being more convinient to use in certain situations
- Vertical support beams can be only place on top of each other or on the bottom of the block if there is a vertical beam 1 block below it. **Would be nice to be able to place them by shift clicking on the placed vertical support beam and be able to extend it upwards**
- Can't place support beams on bottom half slabs, upside stairs

**Meta Info**
 - TFC Version: 0.X.X.XX (at time of request)

### LINKED ISSUE #1240: Player Editing Livestock Feed Lists
**Describe the request**
Make sure to include:
Hi guys. While I was playing with terrafirmacraft, I had a good idea.
After harvesting, a lot of straw remains, which in real life goes to compost or livestock feed. The cattle in the terra firm eats only grain, which is necessary for baking bread, and straw is not used in any way.
What I propose: in the configuration menu, configure the products used by animals. Then those who are not happy with what the animals are eating now will be able to change the desired values.

**Meta Info**
 - TerraFirmaCraft-MC1.12.2-1.0.12.141
 - Еnhancement for TFC alone

### LINKED ISSUE #1248: Skeleton horses spawn
Im not sure but i doubt keep this feature was intencional. If it was i suggest making them behave like TFC horses, ie the need for a lead when riding etc

**Meta Info**
 - TFC Version: 1.0.12.141

### LINKED ISSUE #1249: Crops dont require Sky access to grow
**Describe the bug** crops can be grown undergound
Make sure to include:
1. What did you expect to happen? crops to require sky access
2. What actually happened instead (i.e. what was the bug) grew crops underground

**To Reproduce**
1. Do '...' plant crops underground
2. Then do '...' watch them grow

**Meta Info**
 - TFC Version: 141
 - Were any other mods included? Can you reproduce it without these other mods? no

### LINKED ISSUE #1258: Crucible dupe
Recent changes to small vessel heat recipe opened a bug in `TECrucible`. Basically, since the cached item isn't updated you end up with a duplicated small vessel because `HeatRecipe` does a copy

`ItemStack` and `Capability` instances both have to be updated after all this: https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/254ee67e5d554df9844f4b09a97dc80743a46465/src/main/java/net/dries007/tfc/objects/te/TECrucible.java#L116-L134
(hint here is that inventory is re-updated from recipe but the instance isn't)

### LINKED ISSUE #1259: Blast furnace multiblock only accepts wrought iron, should accept any metal >= Tier 3
blast furnace `Multiblock` needs updating.

Build 142

### LINKED ISSUE #1262: WAILA should show how long trees will grow
**Describe the request**
Make sure to include:
1. What problem or issue would this enhancement solve? People who use WAILA who want to see how long it will take for trees to grow
2. Describe the feature itself - what would change, what would happen. In general more detail is good! Knowing how long a tree will take to grow/mature. We can already check the progress of crops, fruit tree saplings and 
Animals, why not trees?
**Meta Info**
 - TFC Version: 0.X.X.142 (at time of request)
 - Is this a compatibility feature for other mods? Or a enhancement for TFC alone? More of a compatability, since WAILA does work with other tfc items
 - If necessary, what other mods (including versions) would this require? WAILA, of course

### LINKED ISSUE #1265: Raw eggs decays too fast
- Even though fertilized ducks now display "Fertilized" in HWYLA, the eggs they lay dont display the same tag, making it hard to distinguish them from normal eggs.
- Raw eggs now rot, but way too fast, they should take at least 60 days to rot
 - TFC Version: 1.1.0.142

### LINKED ISSUE #1266: Issues regarding preservation of food in vinegar
- When preserving food in vinegar the barrel doesnt display "Preserving" even though it works.
- There needs to be at least 1000ml of vinegar in the barrel for the food inside to be preserved (might be intentional)

**Meta Info**
 - TFC Version: 1.1.0.142

### LINKED ISSUE #1268: Animals are selected to die (by old age mechanics) regardless of age
**Describe the bug**
Make sure to include:
In previous versions, there was a setting for the chance of unexpected cattle death or starvation. There is no configuration in this version, but this happens completely unexpectedly.
Hi guys. In the new update, it became impossible to play. Inexplicably animals disappear. Young and children. This was not the case in previous versions

### LINKED ISSUE #1272: Placing item in donkey/mule saddlebags for first time crashes client
**Describe the bug**
Placing item in donkey/mule saddlebags for first time crashes client. Seems OK thereafter.

**To Reproduce**
Add saddlebags chest, then place item inside.

**Meta Info**
 - TFC Version: 0.X.X.142 on test server

[crash-2020-05-14_17.30.33-client.txt](https://github.com/TerraFirmaCraft/TerraFirmaCraft/files/4629758/crash-2020-05-14_17.30.33-client.txt)

### LINKED ISSUE #1275: Placing torches on large vessels
**Meta Info**
 - TFC Version: 0.X.X.142

### LINKED ISSUE #1276: Rain should turn off fire pits and charcoal forges
Rain doesnt turn off forges like i assume its suposed to (as thats how it was in 1.7.10)
havent tested pit kilns
**Meta Info**
 - TFC Version: 0.X.X.142

### LINKED ISSUE #1277: Goats' milking cooldown isn't working
Hello,

I found some problems after new update. I can milk my goats indefinitely and also I noticed that some of my animals just vanished. Before last update this didn't happen before.

I play TFC with mods: Gravestone, Harvest, Hwyla, JEI, NEI, TFCTech, WailaHarvestability.

DisasterMoo:
- Blast Furnace multiblock structure allows Tier III and above metal sheets (Closes #1259).
- Fixed small vessel dupe in crucible (Closes #1258).
- Added salting recipe wrapper to JEI and fixed log output (Closes #1018).
- Added tree growth to waila provider (Closes #1262).
- Crops now require sky and moisturized farmland, otherwise it will reset growth to the start of current stage (Closes #342, Closes #1249).
- Fixed animals dying regardless of age, also set default config back to disabled (Closes #1268).
- Changed cooked eggs to spoil faster and made raw eggs be the choice for preserving it (Closes #1265)
- Milking now ignore creative and fills the bucket (since it drains from the cow/goat and they have a cooldown, should stop people thinking it is a bug)
- Added missing mating AI to get mules
- Fixed issue in hwyla integration (Closes #1223).
- Fixed crash in chested horses (Closes #1272).
- Made some blocks drop itself if not "supported" (Closes #1154, Closes #1165).
- Fixed quern hitboxes not rendering when looked upon above (Closes #1017).
- Skeleton horses will not spawn on surface anymore (Closes #1248).
- Fixed animals bounding boxes (Closes #1229).
- Fixed torches being able to be placed on large vessels (Closes #1275).
- Added hunting lists to carnivores (Closes #1022, Closes #1073).
- Familiarity overlay now renders only on familiarizable animals (Closes #1099).
- Added food lists to livestock animals (Closes #1240).
- Added convenient support placement on sneaking right click (Closes #1237).
- Added config to change mining time of rock and log blocks (Closes #1060).
- Added CT integration to bloomery and blast furnace (finally, Closes #888).
- Knapping GUI now uses the processing arrow instead of a question mark for JEI click areas (Closes #907)
- Fixed goats being able to be milked without cooldown (Closes #1277).
- Refactored vinegar preservation to use a barrel recipe instead of fluid property. Also, vinegar preservation now requires 125mB of vinegar for each food piece (Closes #1266).
- Rains now make fire pits and charcoal forges consumes more fuel (Closes #1276).
- TFC foods will not decay in any creative tab and in JEI (Closes #1109).
- Fertilized eggs' tooltip now works! (Closes #1234).
- Added gem[Quality][gemName] ore tag to gems (Closes #1148).
- Food poisoning now has it's own poison effect (Closes #996).
- Barrels now uses a fluid whitelist (Closes #942)


src/main/java/net/dries007/tfc/compat/jei/wrappers/SaltingRecipeWrapper.java
```diff
@@ -0,0 +1,68 @@
...
+            for (ItemStack matchingStack : ingredient.getMatchingStacks())
+            {
+                ItemStack stack = matchingStack.copy(); // Avoid changing recipe
+                slot.add(stack);
+                IFood food = stack.getCapability(CapabilityFood.CAPABILITY, null);
+                if (food != null)
+                {
+                    // Reset capability to avoid rotten foods
+                    food.setCreationDate(CalendarTFC.PLAYER_TIME.getTicks());
+                    food.getTraits().clear();
```
alcatrazEscapee: Should not have to do this. This is either
 - loaded before a world exists, which means no item stacks created will decay (and also `CalendarTFC.PLAYER_TIME` is meaningless, or
 - should be replaced with a `food.setNonDecaying()`

DisasterMoo: Yeah, my bad. I didn't notice the first time it was not decaying by default. Now what i found strange is that it is coming with brined from `Ingredient#getMatchingStacks`. But nonetheless, the recipe is salting non salted foods as expected, so i left `food.getTraits().clear()` there.

src/main/java/net/dries007/tfc/objects/blocks/devices/BlockBlastFurnace.java
```diff
@@ -47,29 +47,35 @@
     static
     {
         Predicate<IBlockState> stoneMatcher = state -> state.getBlock() instanceof BlockFireBrick;
-        Predicate<IBlockState> ironSheetMatcher = state -> (state.getBlock() instanceof BlockMetalSheet)
-            && ((BlockMetalSheet) state.getBlock()).getMetal() == Metal.WROUGHT_IRON;
+        Predicate<IBlockState> sheetMatcher = state -> {
+            if (state.getBlock() instanceof BlockMetalSheet)
+            {
+                BlockMetalSheet block = (BlockMetalSheet) state.getBlock();
+                return block.getMetal().getTier().isAtLeast(Metal.Tier.TIER_III);
```
alcatrazEscapee: One thing to consider: As is, this allows metals that are technically "unfinished products", i.e. high carbon steels, weak steels, pig iron, etc. This *may* be fixed, likely just by requiring this as to also be `isToolMetal()` (which would restrict the list in TFC to iron, steel, or colored steel). However I'm unsure of if this change should actually be made

src/main/java/net/dries007/tfc/objects/entity/animal/EntityGoatTFC.java
```diff
@@ -70,6 +75,61 @@ public void birthChildren()
...
+    public boolean isFood(@Nonnull ItemStack stack)
+    {
+        // Check for rotten
+        IFood cap = stack.getCapability(CapabilityFood.CAPABILITY, null);
+        if (!ConfigTFC.Animals.GOAT.acceptRotten && cap != null && cap.isRotten())
+        {
+            return false;
+        }
+        // Check if item is accepted
+        for (String input : ConfigTFC.Animals.GOAT.food)
```
alcatrazEscapee: Firstly, this is *horrendously* gross. It's also duplicated verbatim all over various animals, which just makes it worse. It's using what looks to be a completely custom format (which I don't see a formal spec for anywhere), and you're also doing the parsing of the actual ingredient on demand, every time.

It's a hodgepodge of formatting and conventions:
 - craft tweaker esque "ore" syntax that has no business here.
 - checking for metadata `-1` values, instead of `OreDictionary.WILDCARD_VALUE` or other wildcards.
 - space seperated metadata and item name

An alternate proposal:
 - Change the config to a single string.
 - The format of this string must be a valid `Ingredient` (vanilla) json. (although obviously newlines removed because it's a config and all of these config values are hacked onto what really should be (and will be) json eventually).
 - Then, to parse this, use `gson` and call `CraftingHelper#getIngredient(JsonElement, JsonContext)` to return a vanilla `Ingredient` instance. Lazily cache this as part of the animal data, and invalidate it when configs change (there's an event for that).

src/main/java/net/dries007/tfc/objects/te/TECharcoalForge.java
```diff
@@ -87,6 +87,15 @@ public void update()
...
             {
+                // Have to check the above block, since minecraft think this block is "roof"
+                if (world.isRainingAt(pos.up()))
+                {
+                    // Instantly consume last fuel and turn off
+                    burnTicks = 0;
+                    burnTemperature = 0;
+                    world.setBlockState(pos, state.withProperty(LIT, false));
+                    return;
+                }
```
alcatrazEscapee: Is there a particular reason this both 
 - checks every tick for rain instead of on random ticks or % chances, and
 - has a 100% chance to go out on said exact tick?

Given the charcoal forge has a requirement that it must see sky light (although it can be indirect), I would assume a much lower chance of going out in the rain. It's also a charcoal bleedin forge that can reach temperatures > 1500 C, but rain puts it out *instantly*?

src/main/java/net/dries007/tfc/objects/te/TEFirePit.java
```diff
@@ -125,6 +125,15 @@ public void update()
...
             {
+                // Have to check the above block, since minecraft think this block is "roof"
+                if (world.isRainingAt(pos.up()))
+                {
+                    // Instantly consume last fuel and turn off
+                    burnTicks = 0;
+                    burnTemperature = 0;
+                    world.setBlockState(pos, state.withProperty(LIT, false));
+                    return;
+                }
```
alcatrazEscapee: Same as above, is there a reason this isn't done on random tick? That would make more sense thematically.

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/objects/entity/animal/EntityGoatTFC.java
DisasterMoo: Copy from discord:

> I concede the first implementation i did for animal feeding list is gross, but putting a json in a config? Don't seem right to me, if we want to use json (and i prefer to do that in 1.15 for certain, although i would use the datapacks if possible), I prefer to use an actual json file, like i did for damage resistances
so, the counter-proposal is that, intead of a config, a json file containing all current animals food with a loader=animalFoodList

Thoughts?

DisasterMoo: Done!

alcatrazEscapee (approved):

PR #1278, opened 2020-05-16
How about we make worldgen deterministic again?

### LINKED ISSUE #1205: Worldgen different between SSP and SMP with same seed
Same seed used on a SSP world and SMP world have differences.

Make sure to include:
1. What did you expect to happen? The same seed should derive the same map.
2. What actually happened instead (i.e. what was the bug) Used seed -564885382. In SSP the spawn rock biome is Marble. Started a SMP server with the same seed in server.properties. level-type=tfc_classic and the spawn rock biome is dolomite. The map's appear visually similar otherwise, but no idea how far the differences go.

**To Reproduce**
1. Do 'Start SSP world with seed -564885382. Note spawn rock type (Marble).  

2. Then do 'Start server with same seed, spawn rock type is dolomite

 - TFC Version: 1.0.11.140
 - Were any other mods included? Can you reproduce it without these other mods? Just hwyla on server, journeymap on client

dries007:
**Warning: This changes worldgen! you may experience glitched borders.**

Quick explanation of changes:

1. Fixed wrong assumptions that ocean is always 0 in low level layergen (changes worldgen)
2. Fixed stone layers being non-deterministic because of call to IForgeRegistry.getValuesCollection.
3. Removed dead & commented out code
4. Removed a whole lot of calls to getIdForBiome by caching biomeIDs
5. Gave some var/par variables better names

This would also fix #1205, but actually fixes more than just that.

This will make it once again possible to generate seeds with TFC Seed Maker (update forthcoming), since now that's impossible becasue of the non deterministic issues with point 1 and 2.


src/main/java/net/dries007/tfc/world/classic/ChunkGenTFC.java
```diff
@@ -242,24 +240,23 @@ public Chunk generateChunk(int chunkX, int chunkZ)
...
+//
+//                    chunkPrimerOut.setBlockState(x, 230, z, Blocks.STAINED_GLASS.getStateFromMeta(rockLayer1[z << 4 | x] & 15));
+//                    chunkPrimerOut.setBlockState(x, 220, z, Blocks.STAINED_GLASS.getStateFromMeta(rockLayer2[z << 4 | x] & 15));
+//                    chunkPrimerOut.setBlockState(x, 210, z, Blocks.STAINED_GLASS.getStateFromMeta(rockLayer3[z << 4 | x] & 15));
+//
+//                    chunkPrimerOut.setBlockState(x, 252, z, Blocks.STAINED_GLASS.getStateFromMeta(stabilityLayer[x << 4 | z].layerID & 15));
+//                    chunkPrimerOut.setBlockState(x, 250, z, Blocks.STAINED_GLASS.getStateFromMeta(drainageLayer[x << 4 | z].layerID & 15));
+//                }
+//            }
+//        }
```
alcatrazEscapee: Why leave this commented? If you think there's a reason to remove it, then remove it. Otherwise it's a config option that is disable by default, to be used for whatever purpose.

alcatrazEscapee (commented):
My only concern is this how much of a breaking change this will be to world gen - just need to put a big flashing warning on the change log for that. (last *last* time we say worlds may need to be re-created.)

And the fact that basically all of this code isn't getting ported forward to 1.15 because it's already been rewritten ground up 😉

dries007:
I didn't test this, but I suspect the differances in terrain should be minor.
The rock layer change _should_ only affect SSP. (There will be chunk border differances.)
But yes, this release should come with a big warning.

dries007:
Just did some tests, unfortunately there can quite a bad edge, but sometimes it also perfecly matches up. I generated this world on the latest release and let it generate the full initial region, then I flew around in my patched version.

2020-05-17_00 32 23_1_overworld_day

alcatrazEscapee (approved):

PR #1303, opened 2020-05-22
Added metals, metal items and blocks

DisasterMoo:
- Metal registration via datapacks (missing fluid)
- Metal items registered via MetalItemRecipe, displays tooltip
- Some python script modifications and additions to generate resources for all this data (lang keys, models, blockstates, etc).


alcatrazEscapee:
Closed in favor of #1308 then.

PR #1308, opened 2020-05-23
Ported over some features

DisasterMoo:
- Forging capability which is applied to all items and sync on item tag.
- Heat capability which is loaded by json.
- Ported over `CalendarTFC`
Note: The other PR's commits are here


src/main/java/net/dries007/tfc/ForgeEventHandler.java
```diff
@@ -223,4 +245,42 @@ public static void onWorldTick(TickEvent.WorldTickEvent event)
...
+    public static void attachItemCapabilities(AttachCapabilitiesEvent<ItemStack> event)
+    {
+        ItemStack stack = event.getObject();
+        if (!stack.isEmpty())
+        {
+            // Every item has a forging capability
+            event.addCapability(CapabilityForging.KEY, new ForgingHandler(stack));
+            // Attach heat capability to the ones defined by datapacks
+            CapabilityHeat.getCapability(stack).ifPresent(heat ->
+                event.addCapability(CapabilityHeat.KEY, heat));
```
alcatrazEscapee: Something doesn't seem right here.

src/main/java/net/dries007/tfc/api/Metal.java
```diff
@@ -166,77 +179,158 @@ public boolean isAtMost(@Nonnull Tier requiredInclusive)
...
-        UNFINISHED_BOOTS(true, 200),
-        BOOTS(true, 3, 400, ItemMetalArmor::new),
-
-        SHIELD(true, 400, ItemMetalShield::new);
-        */
+        INGOT("ingots", Type.DEFAULT, metal -> new Item(new Item.Properties().group(TFCItemGroup.METAL))),
+        NUGGET("nuggets", Type.DEFAULT, metal -> new Item(new Item.Properties().group(TFCItemGroup.METAL))),
+        DUST("dusts", Type.DEFAULT, metal -> new Item(new Item.Properties().group(TFCItemGroup.METAL))),
+
+        SCRAP("scraps", Type.PART, metal -> new Item(new Item.Properties().group(TFCItemGroup.METAL))),
```
alcatrazEscapee: We don't use this, there are no plans to use this, it's not a common item, can we just get rid of this?

src/main/java/net/dries007/tfc/api/calendar/Month.java
```diff
@@ -0,0 +1,59 @@
...
+ * See the project README.md and LICENSE.txt for more information.
+ */
+
+package net.dries007.tfc.api.calendar;
+
+import java.util.Arrays;
+
+public enum Month
+{
+    JANUARY(66.5f),
```
alcatrazEscapee: These temperature modifiers need to change - I have plans to simplify some of the weird calculations involved in calendar / climate temperature stuff, but that can wait until I get around to it.

src/main/java/net/dries007/tfc/api/capabilities/forge/ForgeRule.java
```diff
@@ -0,0 +1,134 @@
...
+    ForgeRule(Order order, ForgeStep type)
+    {
+        this.order = order;
+        if (type == HIT_MEDIUM || type == HIT_HARD)
+            this.type = HIT_LIGHT;
+        else
+            this.type = type;
+    }
+
+    @OnlyIn(Dist.CLIENT)
```
alcatrazEscapee: No `OnlyIn` in non-vanilla (or overriding vanilla methods) code.

src/main/java/net/dries007/tfc/api/capabilities/forge/ForgeStep.java
```diff
@@ -0,0 +1,72 @@
...
+        this.u = u;
+        this.v = v;
+    }
+
+    public int getStepAmount()
+    {
+        return stepAmount;
+    }
+
+    @OnlyIn(Dist.CLIENT)
```
alcatrazEscapee: No `OnlyIn` in non-vanilla (or overriding vanilla methods) code.

src/main/java/net/dries007/tfc/api/capabilities/forge/ForgingHandler.java
```diff
@@ -0,0 +1,132 @@
...
+    public <T> LazyOptional<T> getCapability(@Nonnull Capability<T> cap, @Nullable Direction side)
+    {
+        return CapabilityForging.CAPABILITY.orEmpty(cap, capability);
+    }
+
+    /**
+     * Initialize tag if needed, returns a tag with forging data
+     * Only call this when adding work / forge step
+     */
+    private CompoundNBT getTag()
```
alcatrazEscapee, line 128: This is missing a way to detect empty tags, and/or remove the tag from the container item if it's not storing any data.

src/main/java/net/dries007/tfc/api/capabilities/heat/CapabilityHeat.java
```diff
@@ -0,0 +1,154 @@
...
+    public static class HeatWrapper
+    {
+        private final ResourceLocation id;
+        private final Supplier<IHeat> capability;
+        private final Ingredient ingredient;
+
+        public HeatWrapper(ResourceLocation id, JsonObject obj)
+        {
+            this.id = id;
+            this.capability = () -> new HeatHandler(JSONUtils.getFloat(obj, "heat_capacity"), JSONUtils.getFloat(obj, "melt_temperature"));
```
alcatrazEscapee: Evaluate json validity now, not in item stack construction. i.e. store heat capacity as a property of the `HeatWrapper`, and in `getCapability` return a new `HeatHandler`.

src/main/java/net/dries007/tfc/api/capabilities/heat/CapabilityHeat.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Gets a capability (if any) for the stack
+     *
+     * @param stack the stack to search for a capability, if defined by a datapack
+     */
+    public static Optional<IHeat> getCapability(ItemStack stack)
+    {
+        return HeatManager.INSTANCE.getValues().stream()
+            .filter(heatWrapper -> heatWrapper.isValid(stack))
+            .findFirst().map(HeatWrapper::getCapability);
```
alcatrazEscapee: This is a very misleading method. It does *not* get the capability attached to a specific stack, it generates the *template* for such a capability. As such, there's really no reason why this is a util method and should probably just be part of the `AttachCapabilitiesEvent` handler.

src/main/java/net/dries007/tfc/api/capabilities/heat/HeatHandler.java
```diff
@@ -0,0 +1,121 @@
...
+import net.minecraftforge.common.util.LazyOptional;
+
+import net.dries007.tfc.api.calendar.CalendarTFC;
+
+public class HeatHandler implements IHeat
+{
+    private final LazyOptional<IHeat> capability = LazyOptional.of(() -> this);
+    // These are "constants". Some implementations will want to change these based on other factors. (See ItemMold)
+    protected float heatCapacity; // How fast temperature rises and drops
+    protected float meltTemp; // Which temperature this item is considered "molten". Some devices (ie: Charcoal Forge) destroys the item when this is reached
```
alcatrazEscapee: This is actually unused as all practical melting functionality has been moved to `HeatRecipe`

src/main/java/net/dries007/tfc/client/ClientForgeEventHandler.java
```diff
@@ -61,6 +69,10 @@ public static void onRenderGameOverlayText(RenderGameOverlayEvent.Text event)
             BlockPos pos = new BlockPos(mc.getRenderViewEntity().getPosX(), mc.getRenderViewEntity().getBoundingBox().minY, mc.getRenderViewEntity().getPosZ());
             if (mc.world.chunkExists(pos.getX() >> 4, pos.getZ() >> 4))
             {
+                // Always add calendar info
+                //list.add(I18n.format("tfc.tooltip.date", CalendarTFC.CALENDAR_TIME.getTimeAndDate()));
+                list.add(I18n.format(MOD_ID + ".tooltip.debug_times", CalendarTFC.PLAYER_TIME.getTicks(), CalendarTFC.CALENDAR_TIME.getTicks()));
```
alcatrazEscapee: This should occur after the MOD_NAME display.

src/main/java/net/dries007/tfc/command/HeatCommand.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        final ServerPlayerEntity player = source.asPlayer();
+        final ItemStack stack = player.getHeldItemMainhand();
+        if (!stack.isEmpty())
+        {
+            stack.getCapability(CapabilityHeat.CAPABILITY).ifPresent(heat ->
+            {
+                heat.setTemperature(value);
+                source.sendFeedback(new TranslationTextComponent(MOD_ID + ".command.heat", value), true);
+                stack.setCount(2);
```
alcatrazEscapee: What is this?

src/main/java/net/dries007/tfc/objects/entities/TFCFallingBlockEntity.java
```diff
@@ -125,7 +129,7 @@ else if (!world.isAirBlock(posAt.down()) && canFallThrough(world, posAt.down(),
                     BlockState blockstate = world.getBlockState(posAt);
                     setMotion(getMotion().mul(0.7D, -0.5D, 0.7D));
 
-                    if (blockstate.getBlock() != Blocks.MOVING_PISTON)
+                    if (blockstate.getBlock() != net.minecraft.block.Blocks.MOVING_PISTON)
```
alcatrazEscapee: Why is this nessecary?

src/main/java/net/dries007/tfc/objects/items/tools/TFCAxeItem.java
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraft.inventory.EquipmentSlotType;
+import net.minecraft.item.AxeItem;
+import net.minecraft.item.IItemTier;
+
+/**
+ * This is needed so we override the damage math done by vanilla
+ *
+ * For comparison:
+ * Vanilla: Tool value (ie: 3.0 for swords) + material damage value (ie: 2.0 for iron) + 1.0 (hand) = 6.0
+ * TFC: Tool value (ie: 1.3 for maces) * material damage value (ie: 5.75 for steel) + 1.0 (hand) ~= 7.5
```
alcatrazEscapee, line 21: This would be a damn good time to stop and think "is there a reason we do this differently?", and "can we achieve the same results with less effort just calculating what vanilla values are required?

src/main/java/net/dries007/tfc/objects/recipes/MetalItemRecipe.java
```diff
@@ -0,0 +1,169 @@
...
+        getMetalItem(world, stack).ifPresent(metalItem -> {
+            text.add(new TranslationTextComponent(TerraFirmaCraft.MOD_ID + ".tooltip.metal", metalItem.getMetal().getDisplayName()));
+            text.add(new TranslationTextComponent(TerraFirmaCraft.MOD_ID + ".tooltip.units", metalItem.getAmount()));
+            text.add(metalItem.getMetal().getTier().getDisplayName());
+        });
+    }
+
+    protected final ResourceLocation id;
+    protected final Ingredient ingredient;
+    protected final ResourceLocation metal; // Have to store the resource location because at load time metals are not accessible yet
```
alcatrazEscapee: This should definitely be reload-time validated. There's *potential* to have a task-completion based reload listener but I'd have to look further into exactly how we should approach this.

Metals should **NOT** be silently defaulting, they should error if the metal is not specified correctly.

src/main/java/net/dries007/tfc/objects/recipes/MetalItemRecipe.java
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Handles mapping item stacks -> metals
+ *
+ * @see Metal
+ * @see MetalManager
+ */
+public class MetalItemRecipe implements ISimpleRecipe<ItemStackRecipeWrapper>
+{
+    public static Optional<MetalItem> getMetalItem(World world, ItemStack stack)
```
alcatrazEscapee: Why does this not just return the recipe? There seems to be no reason behind creating a separate wrapper class for this specific result query.

src/main/java/net/dries007/tfc/objects/recipes/MetalItemRecipe.java
```diff
@@ -0,0 +1,169 @@
...
+            ResourceLocation metal = new ResourceLocation(JSONUtils.getString(json, "metal"));
+            int amount = JSONUtils.getInt(json, "amount");
+            return new MetalItemRecipe(recipeId, ingredient, metal, amount);
+        }
+
+        @Nullable
+        @Override
+        public MetalItemRecipe read(ResourceLocation recipeId, PacketBuffer buffer)
+        {
+            return new MetalItemRecipe(recipeId, Ingredient.read(buffer), buffer.readResourceLocation(), buffer.readInt());
```
alcatrazEscapee: In this specific case, since the order of these calls is actually important as it's serialization order, it's preferred if these are all separate lines and then a constructor is invoked with variables.

src/main/java/net/dries007/tfc/objects/types/MetalManager.java
```diff
@@ -0,0 +1,39 @@
...
+    {
+        return new Metal(id, obj);
+    }
+
+    @Override
+    protected void postProcess()
+    {
+        super.postProcess();
+        // Sort by tier, making Tier_0 the first (only item should be metal UNKNOWN)
+        orderedTypes.sort(Comparator.comparingInt(metal -> metal.getTier().ordinal()));
```
alcatrazEscapee: Why. They are intentionally sorted as to obtain a stable ID mapping for a given set of items. This sorting makes no sense.

> (only item should be metal UNKNOWN)

This is not an assurance you are allowed to make. There may be any manner of metals (the only requirement we may specify, as in the case of rocks, is there is at least one)

alcatrazEscapee (changes requested):
Initial concerns. For the most part looks good, some of this is because this is the prime time to figure out exactly how we want to go about implementing some of this stuff.

Additionally, talked about in discord:
 - [x] No synchronization for the calendar (so it's completely broken on SMP)
 - [x] I am planning to do a pass over climate, and so a number of the climate related calculations are going to change (but the result will remain the same, it's just going to be simpler and easier to follow)

src/main/java/net/dries007/tfc/ForgeEventHandler.java
alcatrazEscapee: Whoops, this was a temp message I left for myself. I addressed it in a later note.

src/main/java/net/dries007/tfc/api/capabilities/forge/ForgingHandler.java
DisasterMoo: I've done that in `#reset`.

That function should only be called by setters (hence the javadoc i've added above), when you are certain you're adding a tag to the item. It's common code for all 3 since we need it initialized.

Getters are not using that function and instead check for the existence of the tag without creation.

resources/lang/misc.py
```diff
@@ -0,0 +1,43 @@
+from mcresources import ResourceManager
+
+from constants import *
+
+def generate(rm: ResourceManager):
+    # Item Groups
+    rm.lang(('itemGroup.tfc.rock.blocks', 'TFC Rock Blocks'))
```
alcatrazEscapee: A couple things about `rm.lang`:
 - The tuple here isn't necessary, sequential arguments work fine. But more so,
 - In cases like this where there are many different lines, it's probably simpler to use the dictionary invocation:
```python
rm.lang({
    'itemGroup.tfc.rock.blocks': 'TFC Rock Blocks',
    'itemGroup.tfc.metals': 'TFC Metals',
...
})
```

This also has benifits for grouping similar calls:
```python
rm.lang(dict((
    'tfc.enum.tier.tier_%s' % tier,
    'Tier %s' % tier.upper()
) for tier in ('0', 'i', 'ii', 'iii', 'iv', 'v', 'vi'))
```

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/api/capabilities/forge/ForgingHandler.java
alcatrazEscapee: Yes but that's the exact point - now every potential modification to the tag has to check "is this tag actually empty of any useful data, and if so, call reset". For instance, just doing `cap.setWork(0)` is not sufficient.

The idea is that the caller code (the code working with this capability) should not have to consider when doing various set operations, what the state of the data in the tag is. If the right combination (for instance, setting the work then unsetting it, or setting the recipe and removing it) are called, the tag should be empty *without* requiring `reset` to be called. Making it so that this is the caller's responsibility is asking for a future bug to appear where `reset` is not getting called from some call path.

src/main/java/net/dries007/tfc/command/HeatCommand.java
DisasterMoo: I've added for debugging, forgot to remove 😛

src/main/java/net/dries007/tfc/objects/entities/TFCFallingBlockEntity.java
DisasterMoo: Intellij did it after i refactored tags to `TFCTags.Blocks`. It's not necessary, fixing.

src/main/java/net/dries007/tfc/objects/recipes/MetalItemRecipe.java
DisasterMoo: I can only validate at execution time currently. At recipe load time, metals are not yet loaded hence i can't check if it is valid or even store the `Metal` obj directly. If you find a way lmk

src/main/java/net/dries007/tfc/objects/items/tools/TFCAxeItem.java
DisasterMoo: And I did that, my thoughts were we can't achieve the same results. It works for vanilla because there are only 4 materials and they all rise linearly. Our approach is to make a new tier worth the upgrade, while also having different kinds of weapons and not only the sword, hence the potential progression (rest assured, it's not far above vanilla, just a bit better).

That said, other reasons:
- To "achieve" the same results, we're gonna have to throw out `IItemTier#getAttackDamage` and pass out the damage value we want directly to the item class (ie: ```SwordItem(tier_ignoring_damage, damage_modifier * TFCItemTier.realDamage), attackSpeed, builder)```. This surely is gonna cause some inconsistency down the line (ie: another mod asks for the `IItemTier#getAttackDamage`).
- `SwordItem` only accepts integer as damage parameter.
Probably there were more, but i can't remember now.

alcatrazEscapee: I'll buy that.

src/main/java/net/dries007/tfc/objects/recipes/MetalItemRecipe.java
DisasterMoo: Gonna do the discussed refactor later

DisasterMoo:
Done, ready for review.
> No synchronization for the calendar (so it's completely broken on SMP)

I've finished that when you were reviewing so i guess you missed while commenting.

>I am planning to do a pass over climate, and so a number of the climate related calculations are going to change (but the result will remain the same, it's just going to be simpler and easier to follow)

So i guess i'm not touching that, but at least the files are now in place for you to use.

src/main/java/net/dries007/tfc/ForgeEventHandler.java
```diff
@@ -223,4 +247,44 @@ public static void onWorldTick(TickEvent.WorldTickEvent event)
...
+    public static void attachItemCapabilities(AttachCapabilitiesEvent<ItemStack> event)
+    {
+        ItemStack stack = event.getObject();
+        if (!stack.isEmpty())
+        {
+            // Every item has a forging capability
+            event.addCapability(CapabilityForging.KEY, new ForgingHandler(stack));
+
+            // Attach heat capability to the ones defined by datapacks
+            CapabilityHeat.HeatManager.INSTANCE.getValues().stream()
```
alcatrazEscapee: This should be tweaked to use `IndirectHashCollection` a. la. `CollapseRecipe`.
The indirect key is an `Item`, which can be extracted from the `Ingredient` via `getMatchingStacks`.

src/main/java/net/dries007/tfc/api/Metal.java
```diff
@@ -166,77 +179,157 @@ public boolean isAtMost(@Nonnull Tier requiredInclusive)
...
+
+        public String getTag()
+        {
+            return hasTag() ? tag : "";
+        }
+
+        public boolean hasTag()
+        {
+            return tag != null;
+        }
```
alcatrazEscapee, line 337: This would probably be a bit nicer if `getTag` was marked `@Nullable`, and then checked by the caller. As it is, it leaves the potential issue of calling `getTag` without `hasTag` and getting an empty string back (not intentional behavior). Alternatively, if it was marked `@Nullable` the caller's IDE should warn about incorrect usage.

src/main/java/net/dries007/tfc/api/MetalItem.java
```diff
@@ -0,0 +1,80 @@
...
+
+import net.dries007.tfc.TerraFirmaCraft;
+import net.dries007.tfc.objects.types.MetalItemManager;
+import net.dries007.tfc.objects.types.MetalManager;
+
+public class MetalItem
+{
+    public static Optional<MetalItem> get(ItemStack stack)
+    {
+        return MetalItemManager.INSTANCE.getValues().stream()
```
alcatrazEscapee: This should use `IndirectHashCollection`.

src/main/java/net/dries007/tfc/api/calendar/CalendarEventHandler.java
```diff
@@ -0,0 +1,162 @@
...
+import net.minecraftforge.event.CommandEvent;
+import net.minecraftforge.event.TickEvent;
+import net.minecraftforge.event.entity.player.PlayerEvent;
+import net.minecraftforge.event.entity.player.PlayerWakeUpEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+
+@Mod.EventBusSubscriber(modid = MOD_ID)
```
alcatrazEscapee: Prefer using an explicit bus declaration, as it's not obvious.
```suggestion
@Mod.EventBusSubscriber(modid = MOD_ID, bus = Mod.EventBusSubscriber.Bus.FORGE)
```

src/main/java/net/dries007/tfc/api/calendar/CalendarEventHandler.java
```diff
@@ -0,0 +1,162 @@
...
+    public static void onOverworldTick(TickEvent.WorldTickEvent event)
+    {
+        World world = event.world;
+        if (event.phase == TickEvent.Phase.END && world instanceof ServerWorld && event.world.getDimension().getType() == DimensionType.OVERWORLD)
+        {
+            CalendarTFC.INSTANCE.onOverworldTick((ServerWorld) world);
+        }
+    }
+
+    /** todo
```
alcatrazEscapee: This, as far as I know, is not going to happen here. We're going to have to do some brigadier messing about... that will be fun.
(aka, remove no comment)

src/main/java/net/dries007/tfc/api/calendar/CalendarTFC.java
```diff
@@ -0,0 +1,347 @@
...
+import net.minecraft.world.dimension.DimensionType;
+import net.minecraft.world.server.ServerWorld;
+import net.minecraftforge.common.util.INBTSerializable;
+import net.minecraftforge.fml.event.server.FMLServerStartingEvent;
+import net.minecraftforge.fml.network.PacketDistributor;
+
+import net.dries007.tfc.network.CalendarUpdatePacket;
+import net.dries007.tfc.network.PacketHandler;
+
+public final class CalendarTFC implements INBTSerializable<CompoundNBT>
```
alcatrazEscapee:
```suggestion
public final class Calendar implements INBTSerializable<CompoundNBT>
```
For naming consistency - if the prefix is required to avoid import collisions, `TFCCalendar` should be used.

src/main/java/net/dries007/tfc/api/calendar/ICalendarFormatted.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Gets the number of days in a month, from the calendar's instance values
+     *
+     * @return a number of days in a month
+     */
+    long getDaysInMonth();
+
+    /* todo
+    default ITextComponent getTimeAndDate()
```
alcatrazEscapee, line 78: If this is left out due to proxy not existing, it's very much the same as 1.12, just using `DistExecutor`. Although we should upgrade our forge version to get the safer classloaded changes before implenting.
```java
// in TerraFirmaCraft:
private static final IProxy PROXY = DistExecutor.safeRunForDist(() -> () -> new ClientProxy(), () -> () -> new ServerProxy());
```
The lambdas are important, do not convert to method references (suppress warnings if nessecary).

alcatrazEscapee (commented):
A couple nitpicky things / stuff that was marked `todo` that could be simply appended. Mostly updates post commits I made to master with `IndirectHashCollection`. If you could fix those that would be perfect 👍

src/main/java/net/dries007/tfc/api/calendar/ICalendarFormatted.java
DisasterMoo: Not only because of the proxy changes: We now should re-evaluate if this is needed on the server side, and if it is, implement using `ITextComponent` instead of `String`.

Issue #1306, opened 2020-05-23
TFCT: Need cold climate source for lamp fuel


Gaelmare:
Now that olives need 19 degC to ripen in October, most temperate locations won't work for olive oil. We should have another source.

Possibilities:
Tallow
Petroleum products (shale/coal oil?)
Latex/resin
Alcohol
Seed/nut oils
?

 - TFC Version: build 144

nuuno2000:
Imo 19C in october is too hot for olive trees anyways, should be 15 at the most, and id make it lower so northern reagions can get olive oil

Azaysol:
i would like to see alcohol have another used besides a roleplaying aspect. but besides that another one that would be also easy to implement is prob tallow

alcatrazEscapee:
Ah, #1309

PR #1309, opened 2020-05-24
Fruiting no longer dependent on current climate

Gaelmare:
Fruit tree gen re-randomized
Fruit Tree limits rebalanced per EER


Gaelmare:
Fixes #1306

PR #1329, opened 2020-05-25
Now empty buckets will retain if you craft dough from flour

### LINKED ISSUE #1328: Bucket is consumed when making dough
**Describe the bug**
Make sure to include:
1. What did you expect to happen?  When making dough the bucket is emptied but retained.
2. What actually happened instead (i.e. what was the bug)  The bucket was lost.

Initially was playing on official server, but also happens in single player.

**To Reproduce**
Create flour, get bucket of water.  In inventory put bucket of water and flour together to get dough.  Then dough outputs but the bucket is lost.

**Meta Info**
 - TFC Version:  TerraFirmaCraft-MC1.12.2-1.3.0.145

dixon961:
Added some logic to FluidFoodRecepies (it's only dough actually). Don't know if its right solution for #1328 but it works. Please **REVIEW** it, cos i'm not experienced in minecraft modding.

Appreciate any feedback :)

I think it can fix #1328

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/recipes/ShapelessFluidFoodRecipe.java
+++ b/src/main/java/net/dries007/tfc/objects/recipes/ShapelessFluidFoodRecipe.java
@@ -6,10 +6,15 @@
 package net.dries007.tfc.objects.recipes;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 import com.google.gson.JsonObject;
+import net.dries007.tfc.objects.items.ItemsTFC;
+import net.dries007.tfc.objects.items.wood.ItemWoodenBucket;
+import net.minecraft.init.Items;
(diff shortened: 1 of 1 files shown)


alcatrazEscapee:
This was unintentionally broken in 1ff87794845cebc849052459e5e860b7df99e097, which among other things, partially reverted changes in 6fc95b58bafd42a19fb5f449185a5af6bfb3fec4 as the logic has been moved elsewhere (both commits related to #946).

As such, the correct fix it to put said handling back, as this fix is not going to fly. Main issues being:
 - `setContainerItem` should never be getting called on items by external means.
 - `ItemWoodenBucket` already does set it's container item (in the constructor).
 - Your implementation of `checkForRetainedBucket` is woefully inadequate if it comes to any other external bucket item.

But more to the point, it's much better just to fix the bug in `getRemainingItems` rather than any of this: e58928c659ad83deb449216fe60796f0c8915aa8

dixon961:
Thanks for good explanation 👍

PR #1341, opened 2020-05-27
immature dead crops

eerussianguy:
assets only


Gaelmare:
Let's get these textures in so immature crops and dead wild crops look right.

alcatrazEscapee:
> Let's get these textures in so immature crops and dead wild crops look right.

Well then somebody best get on and rewrite the block state so the textures actually do something rather than sitting there being unused asking. 🤔

PR #1342, opened 2020-05-27
Someone wanted plants?

DisasterMoo:
- Ported 1.12 plants into 1.15.
- Added Flora json reloader, as discussed via discord.
- Loot tables, but still need to change drops to thatch when knife harvested.

Oh right, TODO:
- Flora data placement for all plants, using the spreadsheet somehow.


src/main/java/net/dries007/tfc/objects/blocks/plant/TFCBushBlock.java
```diff
@@ -0,0 +1,28 @@
...
+{
+    public TFCBushBlock(Properties properties)
+    {
+        super(properties);
+    }
+
+    @Override
+    protected boolean isValidGround(BlockState state, IBlockReader worldIn, BlockPos pos)
+    {
+        return super.isValidGround(state, worldIn, pos) || TFCTags.GRASS.contains(state.getBlock()) || Tags.Blocks.DIRT.contains(state.getBlock());
```
alcatrazEscapee: This is a great opportunity to introduce a tag `tfc:bush_plantable_on`.
Vanilla has precedent for this - bamboo does this.

src/main/java/net/dries007/tfc/objects/blocks/plant/TFCTallGrassBlock.java
```diff
@@ -0,0 +1,129 @@
...
+    {
+        BlockPos blockpos = context.getPos();
+        return blockpos.getY() < context.getWorld().getDimension().getHeight() - 1 && context.getWorld().getBlockState(blockpos.up()).isReplaceable(context) ? super.getStateForPlacement(context) : null;
+    }
+
+    @Override
+    public void harvestBlock(World worldIn, PlayerEntity player, BlockPos pos, BlockState state, @Nullable TileEntity te, ItemStack stack)
+    {
+        super.harvestBlock(worldIn, player, pos, Blocks.AIR.getDefaultState(), te, stack);
+    }
```
alcatrazEscapee: Why?

src/main/java/net/dries007/tfc/objects/blocks/plant/TFCTallGrassBlock.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    public void onBlockPlacedBy(World worldIn, BlockPos pos, BlockState state, @Nullable LivingEntity placer, ItemStack stack)
+    {
+        worldIn.setBlockState(pos.up(), this.getDefaultState().with(PART, EnumBlockPart.UPPER));
+    }
+
+    @Override
+    public void onBlockHarvested(World worldIn, BlockPos pos, BlockState state, PlayerEntity player)
```
alcatrazEscapee: no no no no no this should all be loot table.

src/main/java/net/dries007/tfc/objects/blocks/plant/PlantBlock.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        stage = IntegerProperty.create("stage", 0, Math.max(1, getPlant().getMaxStage()));
+        builder.add(stage, DAYPERIOD, AGE);
+    }
+
+    /**
+     * Only way to do without having to create a class for each plant with different number of stages
+     * Or ATing into {@link Block}'s stateContainer to make it not final
+     */
+    public abstract Plant getPlant();
```
alcatrazEscapee: Having this here basically restricts any possible usage of this class by addons. There's two options:
1) Return an interface here, and implement said interface in `Plant` (basically `IPlant`). Can also make `Plant` just a "fake" enum (e.g. static class with instances)
 - Pros: it works, it's pretty dead simple
 - Cons: it means addons now have to implement this boring interface, which is already perfectly done by an enum. Hello "generic IPlant class"
2) Detach the block from ever holding a reference to the `Plant`, and instead only reference it's *properties*. For instance, add default implementations of `getSpeedFactor`, `getMaxStage`, and `getStage` directly to `PlantBlock`. This is the approach I've taken so far with rock type blocks.
 - Pros: addons can use this directly, just by extending and/or passing in the correct constructor parameters (minimal effort). It also enforces the idea that our `Plant` enum is *only* for block registration, not for querying later.
 - Cons: more work to do.

src/main/java/net/dries007/tfc/objects/blocks/TFCBlocks.java
```diff
@@ -89,7 +93,14 @@
...
         RegistryObject<T> block = BLOCKS.register(name, blockSupplier);
-        TFCItems.ITEMS.register(name, () -> new BlockItem(block.get(), blockItemProperties));
+        TFCItems.ITEMS.register(name, () -> {
+            Block instance = block.get();
+            if (instance instanceof IBlockItemSupplier)
+            {
+                return ((IBlockItemSupplier) instance).apply(blockItemProperties);
+            }
+            return new BlockItem(block.get(), blockItemProperties);
+        });
```
alcatrazEscapee: This really grinds my gears, but I can't really articulate why. The main reason is it spreads the registration (e.g. constructor properties) out into every single block, rather then everything being consolidated in one class (e.g. `TFCBlocks`). We do consolidate most item blocks rather than have separate `ItemBlock` registry objects in `TFCItems`, I'm not sure if that's good or bad. But this marker interface I really would rather avoid.

src/main/java/net/dries007/tfc/objects/blocks/plant/TFCBushBlock.java
```diff
@@ -0,0 +1,28 @@
...
+{
+    public TFCBushBlock(Properties properties)
+    {
+        super(properties);
+    }
+
+    @Override
+    protected boolean isValidGround(BlockState state, IBlockReader worldIn, BlockPos pos)
+    {
+        return super.isValidGround(state, worldIn, pos) || TFCTags.Blocks.GRASS.contains(state.getBlock()) || Tags.Blocks.DIRT.contains(state.getBlock());
```
alcatrazEscapee: Tag for `bush_plantable_on` which references these two tags.

alcatrazEscapee (changes requested):
I haven't really tried to comment on the world generation. Given what we now know about 1.16.2 it seems that a lot of the feature style world generation (namely any configurable aspects to it) will need to be heavily rewritten to take advantage of that. However this is more comments on the non-actual-world-gen aspects, because as far as I'm aware those are not getting major changes in the near future.

DisasterMoo:
Will get to this soon:tm:

alcatrazEscapee:
Just an FYI, I may get to this next, depending on what other 1.16.x things crash and burn in the meantime. But as it stands this PR needs to be very nearly rewritten to catch up with the 1.15 -> 1.16 update, along with mapping changes and world generation rewrites.

DisasterMoo:
> Just an FYI, I may get to this next, depending on what other 1.16.x things crash and burn in the meantime. But as it stands this PR needs to be very nearly rewritten to catch up with the 1.15 -> 1.16 update, along with mapping changes and world generation rewrites.

Sure, gonna do something this weekend

src/main/java/net/dries007/tfc/objects/blocks/plant/TFCTallGrassBlock.java
DisasterMoo: I've copied vanilla's `DoublePlantBlock#onBlockHarvested`, otherwise, how would a 2 block structure drop a single item?

alcatrazEscapee: Ah, I didn't catch that, my bad.

src/main/java/net/dries007/tfc/objects/blocks/TFCBlocks.java
DisasterMoo: Just this one, i'm not sure how else I can delegate the constructor for specific `Blockitem`s when needed besides using this method. If you have something in mind let me know

DisasterMoo:
Not sure if this will be needed in the future, gonna keep the branch in my repo, otherwise, closing the PR.

Issue #1327, opened 2020-05-25
fire blocks should implement Block#isBurning to help AI pathfinding


Gaelmare:
Only BlockPitKiln implements isBurning. Should be implemented on at least Firepits, Forges, and Bloomeries.

**Meta Info**
 - TFC Version: build 145

DisasterMoo:
Bloomery is a block that is always covered from all sides but the front, it would not add any functionality. The others should light entities on fire if they walk above, so i agree.

Gaelmare:
I meant the blockMolten at the top of the bloomery and blast furnace columns, sorry.

PR #1375, opened 2020-05-31
Moar burnination, tiny bugfixes and bunches of sticks!

Gaelmare:


src/main/resources/assets/tfc/recipes/wood/stick_bunch.json
```diff
@@ -0,0 +1,45 @@
...
+    },
+    {
+      "type": "forge:ore_dict",
+      "ore": "stickWood"
+    },
+    {
+      "type": "forge:ore_dict",
+      "ore": "stickWood"
+    }
+  ],
```
DisasterMoo: eww, too much copy-pasta
should use shaped recipe instead:
```suggestion
{
  "type": "minecraft:crafting_shaped",
  "pattern": [
    "XXX",
    "XXX",
    "XXX"
  ],
  "key": {
    "X": {
      "type": "forge:ore_dict",
      "ore": "stickWood"
    }
  },
```

DisasterMoo (changes requested):

src/main/java/net/dries007/tfc/objects/blocks/devices/BlockFirePit.java
```diff
@@ -90,6 +89,20 @@ public boolean isFullCube(IBlockState state)
...
+     * to entities coming into contact with it. See #1327
+     *
+     * @param world The current world
+     * @param pos   Block position in world
+     * @return True if the block should deal damage
+     */
+    @Override
+    public boolean isBurning(IBlockAccess world, BlockPos pos)
+    {
+        return world.getBlockState(pos).getActualState(world, pos).getValue(LIT);
```
alcatrazEscapee: Don't use `getActualState`

src/main/java/net/dries007/tfc/objects/blocks/devices/BlockCharcoalForge.java
```diff
@@ -145,6 +145,20 @@ public boolean isFullCube(IBlockState state)
...
+     * to entities coming into contact with it. See #1327
+     *
+     * @param world The current world
+     * @param pos   Block position in world
+     * @return True if the block should deal damage
+     */
+    @Override
+    public boolean isBurning(IBlockAccess world, BlockPos pos)
+    {
+        return world.getBlockState(pos).getActualState(world, pos).getValue(LIT);
```
alcatrazEscapee: Don't use `getActualState`

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
```diff
@@ -254,6 +254,8 @@ public static void registerItems(RegistryEvent.Register<Item> event)
             simpleItems.add(register(r, "wood/boat/" + wood.getRegistryName().getPath(), new ItemBoatTFC(wood), CT_WOOD));
         }
 
+        simpleItems.add(register(r, "wood/stick_bunch", new ItemMisc(Size.NORMAL, Weight.LIGHT), CT_WOOD));
```
alcatrazEscapee: This should not be under `wood`, it's not a variant based item.

src/main/java/net/dries007/tfc/objects/blocks/devices/BlockFirePit.java
```diff
@@ -276,25 +289,11 @@ else if (tile.getCookingPotStage() == TEFirePit.CookingPotStage.FINISHED)
...
     @Override
     public void onEntityCollision(World worldIn, BlockPos pos, IBlockState state, Entity entityIn)
     {
-        if (state.getValue(LIT) && !entityIn.isImmuneToFire() && entityIn instanceof EntityLivingBase)
-        {
-            entityIn.attackEntityFrom(DamageSource.IN_FIRE, 1.0F);
-        }
-        super.onEntityWalk(worldIn, pos, entityIn);
+        //todo: handle fuel and item inputs from thrown entities
+        super.onEntityCollision(worldIn, pos, state, entityIn);
```
alcatrazEscapee: The point of `isBurning`: It burns every entity it touches. It's also hardcoded to entity movement as it asks the `World` if it should be taking fire damage. So this `//todo` is rather meaningless, and so there's no reason to have this override here.

alcatrazEscapee (changes requested):
Replacing existing logic with implementing `isBurning` is worse than the current situation:
 - Fire pits burn entities on top, which is implemented by `Entity#move` and `World#isisFlammableWithin`. So hardcoded (the "todo" is rather meaningless in this sense as it's basically saying "revert me")
 - `isBurning` just delegates fire damage to the above methods, which means we lose any control over the amount or power of said fire damage (which ideally we could avoid if possible, which it is?).
 - Items thrown on fire pits, or dropped, will be destroyed, which as a mechanic is rather unfair and feels bad to players. (seeing as it's an inventory you will be interacting with, as opposed to something like fire)
 - With that in mind, pit kilns implement `isBurning` because *they create actual fire*. So in that case it is justified.

Have you tried fixing #1327 with `Block#getAiPathNodeType` instead?

src/main/java/net/dries007/tfc/objects/blocks/devices/BlockFirePit.java
Gaelmare: This is consistent with the usage in BlockPitKiln. Please clarify whether BlockPitKiln should be updated to not do so. What's the drawback?

src/main/java/net/dries007/tfc/objects/items/ItemsTFC.java
Gaelmare: Thanks for clarifying.

src/main/java/net/dries007/tfc/objects/blocks/devices/BlockFirePit.java
Gaelmare: If the isBurning changes are reverted, this todo would be valid then? This is also a fix of previous copypasta.

src/main/java/net/dries007/tfc/objects/blocks/devices/BlockFirePit.java
alcatrazEscapee: - Actual states in 1.12 are a "only use if necessary to escape the 16-variant block metadata limit" as they are removed in future versions.
 - Pit kilns used to have actual states (which is why it queries it). They were removed (760d09e5802ff2cb8c0847fb7a525b1d68ddd464) and so should not be using it any more.
 - Fire pits have never had actual states and so should not be querying it.

src/main/resources/assets/tfc/recipes/wood/stick_bunch.json
Gaelmare: Thanks.

Gaelmare:
PR updated, all comments resolved.

DisasterMoo (approved):

alcatrazEscapee (approved):

PR #1446, opened 2020-07-04
Mossy Cobble

Mrthomas20121:
This Pull Request add Mossy Cobble to Terrafirmacraft.

so far i'm just missing the lang entries.
this pull request doesn't add Mossy Cobble to the world.


alcatrazEscapee:
Copying from a discord discussion:

> Because in terms of adding a feature, it's missing
> - justification for new feature: it's already on 1.15 and backporting isn't a real good reason, given:
> - no crafting recipe
> - no world generation
> - no real use other than decoration
> So given that, I'd highly [suggest] if this is something your interested in, just adding mossy cobble (or heck, many more rock variant blocks) in an addon. Because then we don't have to answer the above questions. As for vanilla parity, we're going to have mossy cobble in 1.15 (along with others), so it's already covered there.

Basically, not interested in backporting / adding this to 1.12. Design issues which make this much simpler to add to TFC than an addon have been addressed in 1.15+ already.

PR #1448, opened 2020-07-04
small bugfixes

Gaelmare:
Will cause missing recipe/item warning for pig iron lamp

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/items/metal/ItemAnvil.java
+++ b/src/main/java/net/dries007/tfc/objects/items/metal/ItemAnvil.java
@@ -62,7 +62,9 @@ public class ItemAnvil extends ItemMetal
         {
             ItemStack stack = player.getHeldItem(hand);
             IBlockState state = worldIn.getBlockState(pos.offset(facing));
-            if (state.getBlock().isReplaceable(worldIn, pos.offset(facing)))
+            IBlockState stateDown = worldIn.getBlockState(pos.offset(facing).down());
+            if (state.getBlock().isReplaceable(worldIn, pos.offset(facing)) &&
+                stateDown.isSideSolid(worldIn, pos.offset(facing).down(), EnumFacing.UP)) //forge says to do it this way, IBlockProperties#isTopSolid
             {
                 if (!worldIn.isRemote)
                 {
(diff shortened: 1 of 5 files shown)


src/main/java/net/dries007/tfc/objects/items/metal/ItemAnvil.java
```diff
@@ -62,7 +62,9 @@ public EnumActionResult onItemUse(EntityPlayer player, World worldIn, BlockPos p
         {
             ItemStack stack = player.getHeldItem(hand);
             IBlockState state = worldIn.getBlockState(pos.offset(facing));
-            if (state.getBlock().isReplaceable(worldIn, pos.offset(facing)))
+            IBlockState stateDown = worldIn.getBlockState(pos.offset(facing).down());
+            if (state.getBlock().isReplaceable(worldIn, pos.offset(facing)) &&
+                stateDown.isSideSolid(worldIn, pos.offset(facing).down(), EnumFacing.UP)) //forge says to do it this way, IBlockProperties#isTopSolid
```
Rongmario: ~~You're offsetting from an offsetted IBlockState already, and again offsetting by calling .down() on the BlockPos.~~ 

I recommend keeping `pos.offset(facing)` and `pos.offset(facing).down()` referenced, I was very confused as it wasn't as readable, and to manipulate BlockPos like that isn't recommended (not as bad in this case as its only onItemUse)

Rongmario (changes requested):

src/main/java/net/dries007/tfc/objects/items/metal/ItemAnvil.java
Gaelmare: I've cleaned this up. I'll claim I only added two .down() calls to this mess, but the rest were already there. ;)

Thanks for the review!

src/main/java/net/dries007/tfc/objects/te/TEAnvilTFC.java
```diff
@@ -285,7 +285,16 @@ else if (inventory.getStackInSlot(SLOT_INPUT_2).isEmpty())
 
                     // Reset forge stuff
                     resetFields();
-                    setRecipe(null);
+                    // if we have a valid single item recipe for our output, use it!
+                    ItemStack output1 = inventory.getStackInSlot(SLOT_INPUT_1);
+                    if (AnvilRecipe.getAllFor(output1).size() == 1 && inventory.getStackInSlot(SLOT_INPUT_2).isEmpty())
```
alcatrazEscapee: Don't call `AnvilRecipe.getAllFor` twice, it has the potential to be an expensive call. Just cache the result.

alcatrazEscapee (changes requested):

alcatrazEscapee (approved):

Rongmario (approved):

PR #1474, opened 2020-07-17
Make support beam range configurable

brndd:
A pet peeve of mine with the cave-in system has always been that it makes it impossible to safely dig up using supports. Rather than disable the system entirely, the last time I played 1.7 TFC I modified the mod to make the range configurable, and then extended it to 2 tiles vertically.

This PR adds that functionality to 1.12 TFC. I limited the maximum ranges relatively low here because of potential performance issues, but quick local testing (and past playing in the far less optimized 1.7 TFC cave-in system) suggests that these values are safe enough.

### DIFF
--- a/src/main/java/net/dries007/tfc/ConfigTFC.java
+++ b/src/main/java/net/dries007/tfc/ConfigTFC.java
@@ -191,6 +191,16 @@ public final class ConfigTFC
             @Config.LangKey("config." + MOD_ID + ".general.fallable.propagateCollapseChance")
             public double propagateCollapseChance = 0.55;
 
+            @Config.Comment("Horizontal radius of the support range of support beams.")
+            @Config.RangeInt(min = 0, max = 8)
+            @Config.LangKey("config." + MOD_ID + ".general.fallable.supportBeamRangeHor")
+            public int supportBeamRangeHor = 4;
+
+            @Config.Comment("Vertical radius of the support range of support beams.")
+            @Config.RangeInt(min = 0, max = 3)
+            @Config.LangKey("config." + MOD_ID + ".general.fallable.supportBeamRangeVert")
+            public int supportBeamRangeVert = 1;
+
             @Config.Comment("Should chiseling raw stone blocks cause collapses?")
             @Config.LangKey("config." + MOD_ID + ".general.fallable.chiselCausesCollapse")
             public boolean chiselCausesCollapse = true;
--- a/src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
+++ b/src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
@@ -54,12 +54,26 @@ public class BlockSupport extends Block
     private static final AxisAlignedBB CONNECTION_E_AABB = new AxisAlignedBB(0.6875D, 0.625D, 0.3125D, 1.0D, 1.0D, 0.6875D);
     private static final AxisAlignedBB CONNECTION_W_AABB = new AxisAlignedBB(0.0D, 0.625D, 0.3125D, 0.3125D, 1.0D, 0.6875D);
 
-    private static final int sRangeHor = ConfigTFC.General.FALLABLE.supportBeamRangeHor;
-    private static final int sRangeVert = ConfigTFC.General.FALLABLE.supportBeamRangeVert;
-    private static final int sRangeHorNeg = sRangeHor * -1;
-    private static final int sRangeVertNeg = sRangeVert * -1;
-
     private static final Map<Tree, BlockSupport> MAP = new HashMap<>();
+    private final Tree wood;
+
+    public BlockSupport(Tree wood)
+    {
+        super(Material.WOOD, Material.WOOD.getMaterialMapColor());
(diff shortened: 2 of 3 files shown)


src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
```diff
@@ -53,6 +54,11 @@
     private static final AxisAlignedBB CONNECTION_E_AABB = new AxisAlignedBB(0.6875D, 0.625D, 0.3125D, 1.0D, 1.0D, 0.6875D);
     private static final AxisAlignedBB CONNECTION_W_AABB = new AxisAlignedBB(0.0D, 0.625D, 0.3125D, 0.3125D, 1.0D, 0.6875D);
 
+    private static final int sRangeHor = ConfigTFC.General.FALLABLE.supportBeamRangeHor;
+    private static final int sRangeVert = ConfigTFC.General.FALLABLE.supportBeamRangeVert;
+    private static final int sRangeHorNeg = sRangeHor * -1;
+    private static final int sRangeVertNeg = sRangeVert * -1;
```
alcatrazEscapee: Remove these fields and instead just directly reference the config values. If these are here it's essentially forcing a MC restart to modify the config values, as they won't be changed beyond first config load.

src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
```diff
@@ -74,29 +80,27 @@ public static ItemStack get(Tree wood, int amount)
      */
     public static boolean isBeingSupported(World worldIn, BlockPos pos)
     {
-        if (!worldIn.isAreaLoaded(pos.add(-32, -32, -32), pos.add(32, 32, 32)))
-        {
-            return true; //If world isn't loaded...
+        if (!worldIn.isAreaLoaded(pos.add(-32, -32, -32), pos.add(32, 32, 32))) {
+            return true; // If world isn't loaded...
```
alcatrazEscapee:
```suggestion
        if (!worldIn.isAreaLoaded(pos.add(-32, -32, -32), pos.add(32, 32, 32)))
        {
            return true; // If world isn't loaded...
```
Formatting. If you're using IDEA please import + use `CodeStyle.xml` found in the root of the repository and try and adhere to our code style.

src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -191,6 +191,16 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
             @Config.LangKey("config." + MOD_ID + ".general.fallable.propagateCollapseChance")
             public double propagateCollapseChance = 0.55;
 
+            @Config.Comment("Horizontal radius of the support range of support beams.")
+            @Config.RangeInt(min = 0, max = 8)
+            @Config.LangKey("config." + MOD_ID + ".general.fallable.supportBeamRangeHor")
```
alcatrazEscapee, line 196: Missing the appropriate lang keys (see the config section in `en_us.lang`) for these two config options.

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
```diff
@@ -54,6 +55,25 @@
...
+        this.wood = wood;
+        OreDictionaryHelper.register(this, "support");
+        Blocks.FIRE.setFireInfo(this, 5, 20);
+        setDefaultState(blockState.getBaseState()
+            .withProperty(AXIS, EnumFacing.Axis.Y)
+            .withProperty(NORTH, false)
+            .withProperty(SOUTH, false)
+            .withProperty(EAST, false)
+            .withProperty(WEST, false));
+    }
```
brndd, line 76: IntelliJ looks to have changed this block after I imported the code style and did an automatic code reformat on the file. Should I just leave it?

src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
```diff
@@ -74,26 +94,33 @@ public static ItemStack get(Tree wood, int amount)
...
             }
         }
         return false;
     }
 
     /**
-     * This is an optimized way to check for blocks that aren't supported during a cave in, instead of checking every single block individually and calling BlockSupper#isBeingSupported
+     * This is an optimized way to check for blocks that aren't supported during a
+     * cave in, instead of checking every single block individually and calling
+     * BlockSupper#isBeingSupported
```
brndd, line 123: This comment along with the one above it were also victims of IntelliJ code formatting.

src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
alcatrazEscapee: Likely means there's a max line length / hard wrap setting I've set somewhere locally that hasn't made it's way into the code style xml, I'll take a look.

brndd: I think it moved it up there because it's a constructor and the style wants those as the first things in a class.

PR #1484, opened 2020-07-21
Make it possible to dig up safely

brndd:
Consider this something of a feedback PR (since the change is so trivial, I might as well make it into a PR rather than an issue) and feel free to close at your discretion.

Currently, it's impossible to dig up using support beams without risking cave-ins. It's my personal belief that this is an oversight on the original developers' part, as one of them [incorrectly claims that digging up safely is possible](https://terrafirmacraft.com/f/topic/6958-digging-up-safely/) on the old forums (and then backpedals saying that "you aren't going to get very many cave ins unless the RNG just hates your guts" after realizing their mistake).

The problem is that getting even a single cave-in will create unsupported blocks in the ceiling, and it's virtually impossible to re-support them from below if they extend more than two blocks above the closest block below them. Due to how the cave-in mechanic works, digging "safe" supported blocks close to these unsupported blocks can still cause the unsupported blocks to cave in, effectively destabilizing your mine and making recovery prohibitively difficult. Therefore having even one cave-in can cause a failcascade leading to a mine that's just going to keep caving in repeatedly.

The fix here is simple: extend the *upwards* support range of support beams by 1. This allows you to dig one block upwards without creating an unsupported ceiling. You can then extend support beams to the new ceiling and dig up another block. The downside is that support beams will allow blocks to float above them, but they can kinda already do that anyway and fixing this isn't possible without substantially changing the algorithm.

### DIFF
--- a/src/main/java/net/dries007/tfc/ConfigTFC.java
+++ b/src/main/java/net/dries007/tfc/ConfigTFC.java
@@ -196,10 +196,15 @@ public final class ConfigTFC
             @Config.LangKey("config." + MOD_ID + ".general.fallable.supportBeamRangeHor")
             public int supportBeamRangeHor = 4;
 
-            @Config.Comment("Vertical radius of the support range of support beams.")
+            @Config.Comment("Upwards support range of support beams.")
             @Config.RangeInt(min = 0, max = 3)
-            @Config.LangKey("config." + MOD_ID + ".general.fallable.supportBeamRangeVert")
-            public int supportBeamRangeVert = 1;
+            @Config.LangKey("config." + MOD_ID + ".general.fallable.supportBeamRangeUp")
+            public int supportBeamRangeUp = 1;
+
+            @Config.Comment("Downwards support range of support beams.")
+            @Config.RangeInt(min = 0, max = 3)
+            @Config.LangKey("config." + MOD_ID + ".general.fallable.supportBeamRangeDown")
+            public int supportBeamRangeDown = 1;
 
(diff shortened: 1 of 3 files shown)


eerussianguy:
Isn't this more appropriate to discuss in the discord? Why submit a PR?

brndd:
> Isn't this more appropriate to discuss in the discord? Why submit a PR?

I prefer Github over Discord, since it's easier to make a point cohesively in a medium like this and the discussion won't get buried as easily. Feel free to discuss it on the Discord among regular collaborators and maintainers; I've made my point and won't feel excluded (though if you want to ping me, I'm on the Discord as burneddi#8888).

alcatrazEscapee (changes requested):
Firstly, this would be *much* better done as simply making the vertical configuration range different in both positive and negative directions (as has already been done in 1.15).

Secondly, while I understand the incentive behind the change, I disagree with simply extending support range upwards to cover the entire range above. With the current WIP changes to support beams (making them *much* better than they are currently) digging upwards should be (while not immune to collapse potential), safe enough.

So: split the config option into upwards + downwards.

brndd:
> With the current WIP changes to support beams (making them much better than they are currently) digging upwards should be (while not immune to collapse potential), safe enough.

I'm not familiar with the details of the WIP changes, but if something's being done to make the mechanic better by default then I think that's good. I was very frustrated to discover that mineshafts can't be extended upwards without collapses when I first played the mod and I can't have been the only one.

That being said, splitting the existing config keys is a good idea so I'll just do that.

PR #1488, opened 2020-07-23
Added concrete barrel recipes (powder to concrete)

PyralIron:
In vanilla you would have to place the concrete powder block next to a water source block to turn it into concrete, which is slow but vanilla doesn't really have a better way to do that. It doesn't fit TFC exactly, and TFC also has the benefit of barrel recipes. Thus I added a recipe for concrete powder + 125 mb of fresh water for 1 hour to make concrete powder. It matches the recipes for dyeing and making concrete powder which 125 mb per item and 1 hour completion time. You could change it to take hot spring water if you want, but with fresh water it at least matches the vanilla way of requiring water to convert.


alcatrazEscapee:
> vanilla doesn't really have a better way to do that

Vanilla does have a way to make concrete. Belivability concerns aside for a moment (since this is not addressing that, nor our recipes do as they're just providing a method to obtain useful vanilla blocks). It's a unique and interesting method to boot, resulting in interesting automation challenges, made more interesting by TFC's restriction on moving water sources.

So given that
- A vanilla method for making concrete already exists, and (as far as I'm aware), there aren't systematic problems with it
- This just takes said mechanic, and adds an easier and more boring alternative
- You could very easily configure this yourself with craft tweaker

I don't like this at all.

PR #1492, opened 2020-07-24
backport trapdoor, door, and fruit tree assets with some color matches

### LINKED ISSUE #1449: Palm Door in-game texture dont match the same object in inventory
All in a title, i dont use any of resourcepacks

Here is screenshot:

l0ih3Vpuaq

Version: TerraFirmaCraft MC1.12.2-1.5.3.153

### LINKED ISSUE #1485: TFCT: Trapdoor models change rotation when opened
So because the classic textures are symmetrical it goes unnoticed, but when you open a trapdoor in certain orientations the texture changes orientation rather than folding downwards. This stems from how TFC uses a different model / blockstate setup for trapdoors than vanilla

eerussianguy:
Fixes #1449 and Fixes #1485


alcatrazEscapee (approved):
While normally I would complain that you edited autogenerated jsons, we're at a point with 1.12 where it's fine to be doing so, doubly so considering this is a backported from correctly, nicely autogenerated jsons.

PR #1512, opened 2020-07-31
Stop wild dead crops popping off

AnodeCathode:
Currently, when they die in the wild they pop off with any block update

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropDead.java
+++ b/src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropDead.java
@@ -31,6 +31,7 @@ import net.minecraftforge.common.EnumPlantType;
 
 import net.dries007.tfc.api.capability.player.CapabilityPlayerData;
 import net.dries007.tfc.api.types.ICrop;
+import net.dries007.tfc.objects.blocks.BlocksTFC;
 import net.dries007.tfc.objects.items.ItemSeedsTFC;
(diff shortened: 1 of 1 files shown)


src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropDead.java
```diff
@@ -140,6 +143,12 @@ public int quantityDropped(IBlockState state, int fortune, Random random)
         return count;
     }
 
+    @Override
+    public boolean canBlockStay(World world, BlockPos pos, IBlockState state) {
```
alcatrazEscapee:
```suggestion
    public boolean canBlockStay(World world, BlockPos pos, IBlockState state)
    {
```
Code style people... code style

src/main/java/net/dries007/tfc/objects/blocks/agriculture/BlockCropDead.java
```diff
@@ -140,6 +143,12 @@ public int quantityDropped(IBlockState state, int fortune, Random random)
...
+        IBlockState soil = world.getBlockState(pos.down());
+        return soil.getBlock().canSustainPlant(soil, world, pos.down(), net.minecraft.util.EnumFacing.UP, this);
```
alcatrazEscapee:
```suggestion
        return soil.getBlock().canSustainPlant(soil, world, pos.down(), EnumFacing.UP, this);
```
Don't use FQNs

Issue #840, opened 2020-02-20
JEI crashes when adding mods that override TFC fluids to an existing world


OmniFS:
**Describe the bug**
Make sure to include:
1. Expected to see the alloy ratios for bronze ingots
2. Game crashed
3. [latest.log](https://github.com/TerraFirmaCraft/TerraFirmaCraft/files/4228917/latest.log)
3a. Crash report https://paste.ee/p/VIbsr

**To Reproduce**
Attention to detail is important!
1. Attempt to view the recipe for a bronze ingot using JEI

**Meta Info**
 - TFC Version: 0.29.3.118
 - VanillaFix 1.0.10-99 was installed and the log may include that along with many many opengl errors.
 - I assume TFC: Metallum 1.0.2

OmniFS:
Further info: This happened using Poetica 0.7.2 https://www.curseforge.com/minecraft/modpacks/poetica/files/2884869

DisasterMoo:
I'm still trying to pinpoint what's the cause (i have a guess tho). But i can tell you this isn't JEI incompatibility or TFC Metallum, some other mod(or maybe a mix, which would be even harder to find) is probably taking our metal property from fluids, which would cause this `NullPointerException`.

My last test, i installed these mods:
- FoamFix (0.10.10)
- Phosphor (0.2.6-build50)
- RandomPatches (1.21.0.0)
- Tinker Contruct (2.13.0.179)
- TexFix (4.0)
- TFC: Metallum (1.0.2)
- TFCTech (1.2.6)
No crashes, JEI pages worked fine.

alcatrazEscapee:
In theory, we have a fluid wrapper system that is robust enough so that if another mod decides to override or replace our fluids, we can still map it to metal properties. My instinct is that somehow the mold was trying to be rendered with a non-metal fluid, or it got filled somehow with something that TFC didn't recognize as a metal.

alcatrazEscapee:
Based on an extensive discord discussion + series of tests, I believe we've isolated the problem:

1) Create a new world, with TFC (+ other mods)
2) Add a mod that registers fluids with the same name as TFC
3) Load the existing world. Forge will emit message of "missing minecraft:blocks, `tfc:fluid/X`"
4) In this world, despite the messages, the fluid block visible in JEI is TFCs, and viewing it in molds causes a crash because it's not the fluid instance that TFC is expecting.

Notable points:
[...]

PR #1555, opened 2020-09-03
Fixed fluid registry issue with already registered fluids

FailMelon:
Compatibility Fix for fluids: https://github.com/turbodiesel4598/NuclearCraft/issues/651#issuecomment-686262754

When attempting to register TFC metal fluids even though the Fluid already exists forge will still add reference in the FluidRegistry delegates for it, this causes Compatibility issues when passing the Fluid through a FluidStack as it grabs the TFC Fluid first instead of the correct fluid.

I'm not sure if this is the best and or intended design of how this should work but it fixes the issue above and allows copper to be melted again.

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
+++ b/src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
@@ -219,11 +219,17 @@ public final class FluidsTFC
     @Nonnull
     private static FluidWrapper registerFluid(@Nonnull Fluid newFluid)
     {
-        boolean isDefault = FluidRegistry.registerFluid(newFluid);
+        boolean isDefault = !FluidRegistry.isFluidRegistered(newFluid.getName());
+
         if (!isDefault)
         {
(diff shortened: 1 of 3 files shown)


src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
```diff
@@ -219,12 +219,18 @@ public static void registerFluids()
...
         if (!isDefault)
         {
             // Fluid was already registered with this name, default to that fluid
             newFluid = FluidRegistry.getFluid(newFluid.getName());
         }
+        else
+        {
+            // No fluid found we are safe to register our default
+            FluidRegistry.registerFluid(newFluid);
+        }
```
alcatrazEscapee, line 233: I'm intrigued. This seems like it may well be the cause of #840 ?

FailMelon: Sorry messed up the PR

PR #1556, opened 2020-09-03
Fixed fluid registry issue with already registered fluids

FailMelon:
Compatibility Fix for fluids: turbodiesel4598/NuclearCraft#651 (comment)

When attempting to register TFC metal fluids even though the Fluid already exists forge will still add reference in the FluidRegistry delegates for it, this causes Compatibility issues when passing the Fluid through a FluidStack as it grabs the TFC Fluid first instead of the correct fluid.

I'm not sure if this is the best and or intended design of how this should work but it fixes the issue above and allows copper to be melted again.

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
+++ b/src/main/java/net/dries007/tfc/objects/fluids/FluidsTFC.java
@@ -214,11 +214,17 @@ public final class FluidsTFC
     @Nonnull
     private static FluidWrapper registerFluid(@Nonnull Fluid newFluid)
     {
-        boolean isDefault = FluidRegistry.registerFluid(newFluid);
+        boolean isDefault = !FluidRegistry.isFluidRegistered(newFluid.getName());
+
         if (!isDefault)
         {
             // Fluid was already registered with this name, default to that fluid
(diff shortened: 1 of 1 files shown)


alcatrazEscapee:
I'm intrigued. This seems like it may well be the cause of #840 ?

FailMelon:
> I'm intrigued. This seems like it may well be the cause of #840 ?

Yeah could well be the cause of it, this is the forge code that happens during register.

```java
    public static boolean registerFluid(Fluid fluid) {
        masterFluidReference.put(uniqueName(fluid), fluid);
        delegates.put(fluid, new FluidRegistry.FluidDelegate(fluid, fluid.getName()));
        if (fluids.containsKey(fluid.getName())) {
            return false;
        }
        ...
    }
```

Even though it returns false when a fluid exist it for some reason puts the fluid a delegate and when you construct a FluidStack this happens.

```java
    public FluidStack(Fluid fluid, int amount) {
        ...
        this.fluidDelegate = FluidRegistry.makeDelegate(fluid);
        this.amount = amount;
    }

    static IRegistryDelegate<Fluid> makeDelegate(Fluid fl) {
        return (IRegistryDelegate)delegates.get(fl);
    }
```

PR #1561, opened 2020-09-07
Fix tree spawning values

eerussianguy:


alcatrazEscapee:
Values in spreadsheet so already pulled, thanks.

Issue #1527, opened 2020-08-09
Lanterns don't save state


alcatrazEscapee:
> pakratt0013 / Today at 9:56 AM
hey, I'm a bit late to update from Tuesday, but I'm hoping that you guys saw that lanterns are broken?
we tested them pretty extensively for a good portion of the last hour.
turning the server off and on emptied the olive oil from the lantern, whether it was on or off...though when the server loaded, it looked on.  I don't know if they just stay on after that, but I'm pretty sure the one I set up last week died at some point.

agilmore2:
Related to #1513 

NO idea how to fix this other than need to add something about serializatron in the TE?

alcatrazEscapee:
Whoops, didn't see it was already an issue.

PR #1574, opened 2020-09-21
CRASH FIX, Bugfixes and new info providers for Lamps and misc

### LINKED ISSUE #1513: TFCT: Oil disappeared from lamp
https://www.twitch.tv/videos/694447054

https://discordapp.com/channels/432522930610765835/432527966266851339/738187810590294107

I was watching this, and I never saw how the oil disappeared. I've watched it at 0.25 speed, and still don't get it.

The lamp was still cooling when filled, but that doesn't appear to change the result.

**Describe the bug**
Filled a lamp with oil and placed it on a ceiling. As soon as it was placed, it showed as a "Bronze Lamp" not as an Olive Oil Bronze Lamp, so it was empty as soon as it was placed.

I've been attempting to reproduce, but have had no luck. 
1. Created metal lamp
2. Filled lamp with oil
3. Placed lamp on ceiling
4. Attempted to light an empty lamp.

**Meta Info**
 - TFC Version: build 156

Gaelmare:
Fix synchronization and capacity setting for TELamp's fluid handler. Fixes #1513 and #1527
Adds TOP and WAILA providers for lamp contents
Fixes OreDictionary registration error of BlockMetalLamp, why register a block when an itemBlock makes more sense?


src/main/java/net/dries007/tfc/CommonEventHandler.java
```diff
@@ -195,7 +195,7 @@ public static void onBlockHarvestDrops(BlockEvent.HarvestDropsEvent event)
         }
 
         // Drop shards from glass
-        if (state.getMaterial() == Material.GLASS && Constants.RNG.nextInt(2) == 0)
+        if (state.getMaterial() == Material.GLASS && Constants.RNG.nextInt(2) == 0 && block != Blocks.GLASS_PANE)
```
DisasterMoo: I think we could instead make the number of dropped glass shard items be proportional to the broken block, ie:
```
Glass Blocks = 1-4
Glass Pane = 0-1
```
And then, either a mold, device or something to smelt those glass shards back into blocks or panes, not only blocks.

Lastly, if this idea is not agreed upon, better to just check for `instanceof BlockGlass` or `instanceof BlockStainedGlass` otherwise, your code will still drop shards for stained glass panes and other modded glass panes.

DisasterMoo (changes requested):

src/main/java/net/dries007/tfc/CommonEventHandler.java
Gaelmare: From the issue, I believe the current conversion is 1 shard = 1 glass block. How about 50% chance of dropping a shard from a block, and 25% chance from anything else? 
I'd rather not implement a new device or mechanic to close this dupe bug in 1.12.

alcatrazEscapee: No `instanceof`. No material or block equality check. Use an ore dict entry, this is a perfect example for it.

DisasterMoo: Oh yeah, I didn't consider ore dicts, they are much better for this purpose. Still, Gael's solution would make the weird scenario where a glass pane drops a shard that can be smelt into a block (where the mass for this block came?). So, for the time being better to use Alc's idea on ore dicts and just check for `blockGlass` to drop a single shard. In the future, if we do implement a glass mechanic in base TFC this could be revisited.

Gaelmare: Oh, good. I'll do this then.

Gaelmare: Well. I tried. See the new commit. Getting an itemStack of a random block that broke in the world isn't straightforward, I think? getPickBlock wants nonNull target and player. ItemAnimalHide#onItemUse is the only place I saw that does something similar, but it already knows the player, and passes a null target. Could make it so glass only drop shards if an entity breaks them?

Gaelmare: Ok, just looked up Block#getItem and did that instead. Think this is ready.

src/main/resources/assets/tfc/lang/en_us.lang
```diff
@@ -1812,741 +1812,741 @@ tile.tfc.double_slab.bricks.schist.name=Schist Brick Double Slab
 tile.tfc.double_slab.bricks.gneiss.name=Gneiss Brick Double Slab
 tile.tfc.double_slab.bricks.marble.name=Marble Brick Double Slab
 
-tile.tfc.ore.native_copper.granite.name=Granite Native Copper
```
alcatrazEscapee, line 1815: So this breaks the translation key for all ore blocks... why?

alcatrazEscapee:
Cherry picked skipping broken prospectors pick fix.

Gaelmare:
The change to propick and translation files worked. I guess we'll have to
add ".name" to the translation key delivered by the key lookup in the
propick section then. Thanks!

PR #1576, opened 2020-09-25
Groundcover PR, but 1.16

eerussianguy:
**Groundcover**
Blocks, but when you right click on them they get destroyed and drop their loot table. Rotatable and waterloggable. They have a few types:
- Miscellaneous, not really linked to anything, relatively simple
- Pebble, up to three rock items can be placed on a block to create different versions of this.
- Fallen leaves, changes with season, etc etc
- Twigs -- delineated by wood, drops a single stick

**Thatch Beds**
- Might need to be renamed to the more traditional name, Straw & Hide Bed
- Agnostic to player direction when placed
- The actual block needs to be hidden from the player somehow, since it's missingno and breaks as soon as you place it

**Texture Autogeneration**
- Currently able to autogen bookshelves, stripped log tops, log tops, workbenches, mossy bricks, mossy cobble
- Runs only when you manually run it, not under generate_resources.py

**Player Inventory**
- New assets, fixed bugs with it not showing the right stuff

**Misc**
- Added thatch, hide items, ore items
- Loot tables for lots of stuff
- Tweaked the TFCLeavesBlock class to be better-er
- Material colors for wood


resources/generate_textures.py
```diff
@@ -0,0 +1,22 @@
+from PIL import Image# pip install pillow, don't try to install PIL
```
alcatrazEscapee: This could be a note in the `README` where the other two libraries are noted.

resources/constants.py
```diff
@@ -209,20 +210,23 @@
     'diamond': Vein('diamond', 'pipe', 60, 60, 5, 140, 40, 0, 0, 0, ['gabbro']),
     'emerald': Vein('emerald', 'pipe', 80, 60, 5, 140, 40, 0, 0, 0, ['igneous_intrusive']),
 }
-ROCK_BLOCK_TYPES = ('raw', 'bricks', 'cobble', 'gravel', 'smooth', 'mossy_cobble', 'mossy_bricks', 'cracked_bricks', 'chiseled', 'spike')
-CUTTABLE_ROCKS = ('raw', 'bricks', 'cobble', 'smooth', 'mossy_cobble', 'mossy_bricks', 'cracked_bricks')
-ROCK_SPIKE_PARTS = ('base', 'middle', 'tip')
-SAND_BLOCK_TYPES = ('brown', 'white', 'black', 'red', 'yellow', 'green', 'pink')
-SOIL_BLOCK_TYPES = ('dirt', 'grass', 'grass_path', 'clay', 'clay_grass')
-SOIL_BLOCK_VARIANTS = ('silt', 'loam', 'sandy_loam', 'silty_loam')
+ROCK_BLOCK_TYPES = ['raw', 'bricks', 'cobble', 'gravel', 'smooth', 'mossy_cobble', 'mossy_bricks', 'cracked_bricks', 'chiseled', 'spike', 'pebble']
```
alcatrazEscapee: No reason to changes these from tuples -> lists

resources/generate_textures.py
```diff
@@ -0,0 +1,22 @@
...
+import os
+from constants import *
+
+path = '../src/main/resources/assets/tfc/textures/'
```
alcatrazEscapee, line 5: This should use the `main()` function pattern and not have executable code on import

src/main/java/net/dries007/tfc/common/blocks/ThatchBedBlock.java
```diff
@@ -0,0 +1,95 @@
...
+    public boolean isFlammable(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        return true;
+    }
+
+    @Override
+    public int getFireSpreadSpeed(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        int speed = 200;
+        speed += ClimateRenderCache.INSTANCE.getTemperature() * 3;
```
alcatrazEscapee: As per the javadoc comment on `ClimateRenderCache`:

> This stores the climate parameters at the current client player location, for quick lookup in rendering purposes

None of that applies or can be used here. You're making the fire spread speed dependent on the current *client player* location, which makes no logical sense but also either crash servers or be completely defunct.

As mentioned in discord I'd like to look into extending `AbstractBlock.Properties` for this, as there's a number of methods (tool type, fire spread speed, tile entity creation, etc.) that would be useful to avoid creating a pile of unnecessary `Block` subclasses.

src/main/java/net/dries007/tfc/common/blocks/ThatchBlock.java
```diff
@@ -0,0 +1,65 @@
...
+    public ThatchBlock()
+    {
+        super(Properties.of(Material.REPLACEABLE_PLANT).strength(0.6F, 0.4F).noOcclusion());
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public void entityInside(BlockState state, World worldIn, BlockPos pos, Entity entityIn)
+    {
+        entityIn.makeStuckInBlock(state, new Vector3d(0.3D, (double)0.24F, 0.3D));
```
alcatrazEscapee:
```suggestion
        entityIn.makeStuckInBlock(state, new Vector3d(0.3D, 0.24, 0.3D));
```

bad vanilla decompile copypasta

src/main/java/net/dries007/tfc/common/blocks/ThatchBlock.java
```diff
@@ -0,0 +1,65 @@
...
+    @SuppressWarnings("deprecation")
+    public VoxelShape getCollisionShape(BlockState state, IBlockReader worldIn, BlockPos pos, ISelectionContext context)
+    {
+        return VoxelShapes.empty();
+    }
+
+    @Override
+    public int getFlammability(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        return 100 + RANDOM.nextInt(50);
```
alcatrazEscapee, line 49: Don't return random here it makes no logical sense.

src/main/java/net/dries007/tfc/common/blocks/ThatchBlock.java
```diff
@@ -0,0 +1,65 @@
...
+    public boolean isFlammable(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        return true;
+    }
+
+    @Override
+    public int getFireSpreadSpeed(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        int speed = 60;
+        speed += ClimateRenderCache.INSTANCE.getTemperature() * 3;
```
alcatrazEscapee: Can't use `ClimateRenderCache` here

src/main/java/net/dries007/tfc/common/blocks/rock/PebbleBlock.java
```diff
@@ -0,0 +1,69 @@
...
+import net.minecraft.util.math.shapes.ISelectionContext;
+import net.minecraft.util.math.shapes.VoxelShape;
+import net.minecraft.world.IBlockReader;
+import net.minecraft.world.World;
+
+public class PebbleBlock extends GroundcoverBlock
+{
+    public static final BooleanProperty WATERLOGGED = BlockStateProperties.WATERLOGGED;
+    public static final DirectionProperty FACING = HorizontalBlock.FACING;
+    public static final IntegerProperty ROCKS = IntegerProperty.create("rocks", 1, 3);
```
alcatrazEscapee: Initialize properties in `TFCBlockStateProperties`, only if it's necessary and we can't borrow one from `BlockStateProperties`. The result is we don't have duplicate similarly named properties floating around that are unexchangeable due to identity equality constraints.

src/main/java/net/dries007/tfc/common/blocks/rock/PebbleBlock.java
```diff
@@ -0,0 +1,69 @@
...
+
+    protected static final VoxelShape ONE = Block.box(5.0D, 0.0D, 5.0D, 11.0D, 2.0D, 11.0D);
+    protected static final VoxelShape TWO = Block.box(2.0D, 0.0D, 2.0D, 14.0D, 2.0D, 14.0D);
+    protected static final VoxelShape THREE = Block.box(5.0D, 0.0D, 5.0D, 11.0D, 4.0D, 11.0D);
+
+    private final Rock.Default rock;
```
alcatrazEscapee: Do not store the rock field directly on the block. If it's needed indirectly there are much better options:
- If it's needed for an item / block lookup based on this, store a `Supplier` to the lookup'd item (see grass + dirt)
- If it's needed to access rock -> this, go through the rock manager and add it to the json defined optional (or not) block requirements.
- Or, consider a tag/recipe if more complex (and potentially modifiable) behavior is required)

src/main/java/net/dries007/tfc/common/blocks/wood/FallenLeavesBlock.java
```diff
@@ -0,0 +1,106 @@
...
+    public boolean isFlammable(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        return true;
+    }
+
+    @Override
+    public int getFireSpreadSpeed(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        int speed = 100;
+        speed += ClimateRenderCache.INSTANCE.getTemperature() * 3;
```
alcatrazEscapee: Can't use `ClimateRenderCache` here

src/main/java/net/dries007/tfc/common/blocks/wood/TFCLeavesBlock.java
```diff
@@ -131,25 +137,44 @@ public void randomTick(BlockState state, ServerWorld worldIn, BlockPos pos, Rand
...
     protected void createBlockStateDefinition(StateContainer.Builder<Block, BlockState> builder)
     {
         builder.add(PERSISTENT, SEASON_NO_SPRING, getDistanceProperty());
     }
 
+    @Nonnull
+    @Override
+    public BlockState getStateForPlacement(BlockItemUseContext context)
+    {
+        Season season = Calendars.SERVER.getCalendarMonthOfYear().getSeason();
```
alcatrazEscapee: You have a `IWorld` through `BlockItemUseContext` so use that to query the logical side specific calendar (e.g. `Calendars#get(IWorld)`)

src/main/java/net/dries007/tfc/common/blocks/wood/TwigBlock.java
```diff
@@ -0,0 +1,59 @@
...
+    public boolean isFlammable(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        return true;
+    }
+
+    @Override
+    public int getFireSpreadSpeed(BlockState state, IBlockReader world, BlockPos pos, Direction face)
+    {
+        int speed = 100;
+        speed += ClimateRenderCache.INSTANCE.getTemperature() * 3;
```
alcatrazEscapee: Can't use `ClimateRenderCache` here.

src/main/java/net/dries007/tfc/common/blocks/GroundcoverBlock.java
```diff
@@ -0,0 +1,168 @@
...
+    public static final BooleanProperty WATERLOGGED = BlockStateProperties.WATERLOGGED;
+    public static final DirectionProperty FACING = HorizontalBlock.FACING;
+
+    protected static final VoxelShape FLAT = Block.box(2.0D, 0.0D, 2.0D, 14.0D, 2.0D, 14.0D);
+    protected static final VoxelShape SMALL = Block.box(5.0D, 0.0D, 5.0D, 11.0D, 2.0D, 11.0D);
+    protected static final VoxelShape MEDIUM = Block.box(5.0D, 0.0D, 5.0D, 11.0D, 4.0D, 11.0D);
+    protected static final VoxelShape LARGE = Block.box(2.0D, 0.0D, 2.0D, 14.0D, 8.0D, 14.0D);
+    protected static final VoxelShape PIXEL_HIGH = Block.box(0.0D, 0.0D, 0.0D, 16.0D, 1.0D, 16.0D);
+
+    public VoxelShape shape;
```
alcatrazEscapee, line 41: Should be `private final`, doubtful if accessors are needed.

src/main/java/net/dries007/tfc/common/blocks/GroundcoverBlock.java
```diff
@@ -0,0 +1,168 @@
...
+        STICK(FLAT);
+
+        private final VoxelShape shape;
+
+        MiscCoverTypes(VoxelShape shape)
+        {
+            this.shape = shape;
+        }
+
+        public VoxelShape getShape() { return shape; }
```
alcatrazEscapee, line 166: Consistent formatting. Don't avoid newlines like this. There's a bunch of other nitpicks that could be resolved by an auto-reformat.

src/main/java/net/dries007/tfc/common/blocks/ThatchBedBlock.java
```diff
@@ -0,0 +1,95 @@
...
+        {
+            world.destroyBlock(pos, true);
+        }
+    }
+
+    @Override
+    public VoxelShape getShape(BlockState state, IBlockReader worldIn, BlockPos pos, ISelectionContext context) { return BED_SHAPE; }
+
+    @Override
+    public TileEntity newBlockEntity(IBlockReader worldIn) { return null; }
```
alcatrazEscapee, line 74: Need to also override `IForgeBlock#hasTileEntity` to return false.

alcatrazEscapee (changes requested):
Preliminary concerns

eerussianguy:
Everything fire related has been removed for now

PR #1578, opened 2020-09-28
Plants, but in 1.16

DisasterMoo:
~~Still WIP, needs tweaks on generation and placements, but otherwise the plant blocks are there and working (no water plants yet tho).~~

Edit:
PR done, although I'm having a bad time figuring out how to tweak this to perfection.


src/main/java/net/dries007/tfc/world/vein/VeinType.java
```diff
@@ -59,7 +59,7 @@ public VeinType(ResourceLocation id, JsonObject json)
             throw new JsonParseException("Vertical Size must be > 0.");
         }
         density = JSONUtils.getAsInt(json, "density", 20) / 100f;
-        if (density <= 0 || density > 100)
+        if (density <= 0 || density > 1)
```
DisasterMoo, line 61: Just wanted to point out this. I've found out what seems to be an overlook, so I've fixed it in this branch.

The problem here is that you aren't checking for the user input but rather the value after it being divided by 100. It seems you wanted it to be 0-1 (and the user input between 0-100), hence the change.

alcatrazEscapee: Yup. Everything there needs a rewrite anyway at some point. *sigh*.

DisasterMoo:
A bit of changes, seems I can make use of vanilla's `RandomPatchFeature` for placement but I still had to create another feature for rainfall and temperature check much like `ForestFeature` (read copy-pasta 😛)

And yes, still needs to write those python scripts 😩

PR #1607, opened 2020-10-17
Immersive ways of cooling heatable items - Resolves #1347

### LINKED ISSUE #1347: Item entities should cool down when in water
Right now, the only way to cool down items faster is to put them in a barrel of water. It would make sense if putting the items in a body of water would do the same. Since tossing the items into the water wouldn't work in 1.12.2, having the player stand in water to cool down items in the inventory would be a good alternative.

**Meta Info**
 - TFC Version: 1.3.2.147 (at time of request)
 - TFC Enhancement
 - None

Rongmario:
- Also adds faster rate of cool down when its a cooler liquid than water
- If the item is on top of a block that has ice, packed ice or snow material, it also cools down accordingly

This implementation utilizes EntityJoinWorldEvent and ItemExpireEvent, they go hand-in-hand with each other. 

By manually editing the tag of the EntityItem upon spawning, we can identify what EntityItems should be subjected to the cooling process. We also apply an extremely short lifespan to mimic the speed of the barrel recipe, albeit slightly longer (half a second, but modifiable via config). Then all the checks are done in ItemExpireEvent to set a new cooler temperature depending on if the conditions are met. 

First of all, if temperature is 0 or less, then nothing needs to be done and the original lifespan is restored/added on.

If no conditions are met, the same, short length of lifespan is added on so a quick check can be done once again in case the condition will be met later. Now if, this has been repeated for a long time up until it meets the item's normal lifespan, it will despawn.

Finally, if a condition is met, we extend its lifespan once again so more checks can be done to cool down even further.

Fixes #1347

### DIFF
--- a/src/main/java/net/dries007/tfc/CommonEventHandler.java
+++ b/src/main/java/net/dries007/tfc/CommonEventHandler.java
@@ -756,7 +756,7 @@ public final class CommonEventHandler
                 }
             }
         }
-        if (ConfigTFC.Devices.TEMPERATURE.coolHeatablesInWater && entity instanceof EntityItem)
+        if (ConfigTFC.Devices.TEMPERATURE.coolHeatablesInWorld && entity instanceof EntityItem)
         {
             EntityItem entityItem = (EntityItem) entity;
             ItemStack stack = entityItem.getItem();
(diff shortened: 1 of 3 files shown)


eerussianguy:
missing lang entry

Rongmario:
Fixed.

DisasterMoo (approved):
Very interesting, looks good to me!

PR #1623, opened 2020-11-03
Gems add work / check work for sluice + new command to change the work in a chunk

### LINKED ISSUE #1621: TFCT: unlimited sluicing for gems
The gem drop check in TESluice does not check for chunk work, and therefore yields unlimited gems.

 - TFC Version: build 163+
 - observed both in bare TNG playthrough and TNFC

eerussianguy:
Closes #1621 

The problem was that the thing that checked for gems didn't care about chunk work at all. I didn't do anything sophisticated here, I just made it check work and also add work to the chunk.

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/te/TESluice.java
+++ b/src/main/java/net/dries007/tfc/objects/te/TESluice.java
@@ -104,17 +104,24 @@ public class TESluice extends TEBase implements ITickable
                     }
                     if (Constants.RNG.nextDouble() < ConfigTFC.Devices.SLUICE.gemChance)
                     {
-                        Gem dropGem;
-                        if (Constants.RNG.nextDouble() < ConfigTFC.Devices.SLUICE.diamondGemChance)
+                        Chunk chunk = world.getChunk(pos);
+                        ChunkDataTFC data = ChunkDataTFC.get(chunk);
+                        if (data.canWork(1))
                         {
-                            dropGem = Gem.DIAMOND;
+                            Gem dropGem;
+                            if (Constants.RNG.nextDouble() < ConfigTFC.Devices.SLUICE.diamondGemChance)
+                            {
+                                dropGem = Gem.DIAMOND;
+                            }
+                            else
+                            {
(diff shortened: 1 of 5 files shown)


Rongmario (approved):

src/main/java/net/dries007/tfc/objects/te/TESluice.java
```diff
@@ -104,17 +104,24 @@ public void update()
                     }
                     if (Constants.RNG.nextDouble() < ConfigTFC.Devices.SLUICE.gemChance)
                     {
-                        Gem dropGem;
-                        if (Constants.RNG.nextDouble() < ConfigTFC.Devices.SLUICE.diamondGemChance)
+                        Chunk chunk = world.getChunk(pos);
```
DisasterMoo: My only concern here is that this will only check and work on the sluice's chunk, check the above lines (L73-L87) for how to do it. I also recommend moving those lines to a new function so it can be used by both ores and gems, just don't forget to add a boolean parameter for ores, since those chunks only are worked if there is a vein there.

DisasterMoo (changes requested):

src/main/java/net/dries007/tfc/objects/te/TESluice.java
```diff
@@ -281,4 +261,37 @@ private void consumeSoil()
...
+                if (chunkData.canWork(1))
+                {
+                    if (checkVeins && chunkData.getGeneratedVeins().stream().anyMatch(vein -> vein.getType() != null && vein.getType().getOre() != null))
+                    {
+                        chunks.add(chunk);
+                    }
+                    else if (!checkVeins)
+                    {
+                        chunks.add(chunk);
+                    }
```
DisasterMoo: Minor improvement 😛 
```suggestion
                    if (!checkVeins || chunkData.getGeneratedVeins().stream().anyMatch(vein -> vein.getType() != null && vein.getType().getOre() != null))
                    {
                        chunks.add(chunk);
                    }
```

DisasterMoo (approved):

PR #1631, opened 2020-11-21
Add numeric display option for Heat and Ores

AnodeCathode:
For modded TFC with Metals having melt temps > 3000 degrees, it's become clear that the Temp Enum is insufficient. So this adds an advanced option to the ore display that also applies to the Heat items and displays the specific numeric amounts.


src/main/java/net/dries007/tfc/objects/items/metal/ItemOreTFC.java
```diff
@@ -193,6 +208,15 @@ public boolean canMelt(ItemStack stack)
         return ore.canMelt();
     }
 
+    @Override
+    public float getMeltTemp(ItemStack stack)
+    {
+       if (this.canMelt(stack)){
```
DisasterMoo: Minor code style issue

src/main/resources/assets/tfc/lang/en_us.lang
```diff
@@ -87,6 +87,7 @@ tfc.tooltip.oredictionaryentry=§fOre Dictionary name:§7 %s
 tfc.tooltip.oredictionaryentries=§fOre Dictionary names:
 tfc.tooltip.metal=§fMetal:§7 %s
 tfc.tooltip.units=%d units
+tfc.tooltip.melttemp=%d°
```
DisasterMoo: This opens up debate of which units we use, which for me seems to be Celsius. If so:
```suggestion
tfc.tooltip.melttemp=%d °C
```

DisasterMoo (changes requested):
Just that minor code style issue, but also let's see what Alc think about the temperature units.

src/main/resources/assets/tfc/lang/en_us.lang
AnodeCathode: Shouldn't be a space between the value and the symbol. Whether to include the letter for the scale is entirely up to you

AnodeCathode: Added without the space but with the C symbol

DisasterMoo (approved):

PR #1632, opened 2020-11-23
Fix some minor issues with WAILA/TOP 

Rongmario:
-Fixes OreProvider not showing drop's ItemStack model as intended.

- (UP FOR DEBATE): removal of `OreProvider#getTooltip` as it is redundant, for the title of it already mentions what the ore is/contains.

+API addition `appendBody`: denotes if the particular IWailaBlock impl should have its `getTooltip` queried on showing tooltip.


DisasterMoo:
IIRC, at the time I wrote the APIs ore items didn't had the stack information they add now. Probably not a bad idea removing `OreProvider#getTooltip` altogether, if that is not cause any issues.

PR #1640, opened 2020-11-29
Check for if a metal is usable

Charnuz:


src/main/java/net/dries007/tfc/api/types/Metal.java
```diff
@@ -280,7 +285,7 @@ public static Item create(Metal metal, ItemType type)
 
         public boolean hasType(Metal metal)
         {
-            if (!metal.usable)
+            if (!metal.isUsable())
```
Rongmario, line 288: No need to change this, we're encapsulated in the same class anyways.

Rongmario (changes requested):

src/main/java/net/dries007/tfc/api/types/Metal.java
Charnuz: it's for cases where a class extends Metal and usability differs from what is passed to the constructor

Charnuz: Would like to note that it is more consistent with this change.
[Line 299 does something similar](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.12.x/src/main/java/net/dries007/tfc/api/types/Metal.java#L299)

DisasterMoo: I agree with Charnuz, it should match the function for consistency and for children classes.

src/main/java/net/dries007/tfc/TerraFirmaCraft.java
```diff
@@ -53,7 +53,7 @@
 
 @SuppressWarnings("FieldMayBeFinal")
 @Mod.EventBusSubscriber
-@Mod(modid = MOD_ID, name = TerraFirmaCraft.MOD_NAME, useMetadata = true, guiFactory = Constants.GUI_FACTORY, dependencies = "required:forge@[14.23.5.2816,);after:jei@[4.14.2,);after:crafttweaker@[4.1.11,);after:waila@(1.8.25,)")
+@Mod(modid = MOD_ID, name = TerraFirmaCraft.MOD_NAME, useMetadata = true, guiFactory = Constants.GUI_FACTORY, dependencies = "required:forge@[14.23.5.2816,);after:jei@[4.14.2,);after:crafttweaker@[4.1.11,);after:waila@(1.8.25,);before:gregtech")
```
DisasterMoo: Now this change is awkward. The other "after" dependencies are there because we are indeed soft dependant on waila and crafttweaker. Adding before gregtech implies that you want our registry stuff to happen earlier (possibly because of the metal fluids). If so, this is not the solution since other mods also have the possibility of registering earlier than TFC. And by no means adding each mod out there in this line would be the solution.

DisasterMoo (changes requested):

src/main/java/net/dries007/tfc/api/types/Metal.java
DisasterMoo: I agree with Charnuz, this should match the function for children classes.

src/main/java/net/dries007/tfc/TerraFirmaCraft.java
Charnuz: This was an accident and not meant for this branch. I am in the progress of reverting, oops

DisasterMoo (approved):

PR #1644, opened 2020-12-03
Change Berrybush to use custom damage source

AnodeCathode:
To differentiate from actual Cactus damage. For important reasons.


DisasterMoo (approved):
> For important reasons.

*I see...* Meme Advancement? 😛 
Regardless, approved.

PR #1647, opened 2020-12-09
Prospector's Pick events

LordFokas:
Add event (scoped by logical side) to fire with information regarding usage of Prospector's Pick.

Fixes inconsistency where on the dedicated server whatever config is set in `ConfigTFC.Client.TOOLTIP.propickOutputToActionBar` applies to all connected clients when players use the propick, regardless of what is configured on the client side.

### DIFF
--- a/src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
+++ b/src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
@@ -1,6 +1,7 @@
 package net.dries007.tfc.api.events;
 
 import net.dries007.tfc.objects.items.metal.ItemProspectorPick.ProspectResult.Type;
+
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.math.BlockPos;
@@ -12,16 +13,21 @@ import net.minecraftforge.fml.relauncher.Side;
  * Carries all the data relative to the result displayed to the player.
  * One of the two subclasses will be used according to the logical side.
  */
-public abstract class ProspectEvent extends Event {
+public abstract class ProspectEvent extends Event
+{
 
-    public static class Server extends ProspectEvent {
-        public Server(EntityPlayer player, BlockPos pos, Type type, ItemStack vein){
+    public static class Server extends ProspectEvent
+    {
+        public Server(EntityPlayer player, BlockPos pos, Type type, ItemStack vein)
+        {
             super(Side.SERVER, player, pos, type, vein);
         }
     }
 
-    public static class Client extends ProspectEvent {
-        public Client(EntityPlayer player, BlockPos pos, Type type, ItemStack vein){
+    public static class Client extends ProspectEvent
+    {
+        public Client(EntityPlayer player, BlockPos pos, Type type, ItemStack vein)
+        {
             super(Side.CLIENT, player, pos, type, vein);
         }
     }
@@ -32,7 +38,8 @@ public abstract class ProspectEvent extends Event {
     private Type type;
     private ItemStack vein;
 
-    protected ProspectEvent(Side side, EntityPlayer player, BlockPos pos, Type type, ItemStack vein){
+    protected ProspectEvent(Side side, EntityPlayer player, BlockPos pos, Type type, ItemStack vein)
+    {
         this.side = side;
         this.player = player;
         this.pos = pos;
@@ -40,23 +47,28 @@ public abstract class ProspectEvent extends Event {
         this.vein = vein;
     }
 
-    public Side getSide(){
+    public Side getSide()
+    {
         return side;
     }
 
-    public EntityPlayer getPlayer(){
+    public EntityPlayer getPlayer()
+    {
         return player;
     }
 
-    public BlockPos getBlockPos(){
+    public BlockPos getBlockPos()
+    {
         return pos;
     }
 
-    public Type getResultType(){
+    public Type getResultType()
+    {
         return type;
     }
 
-    public ItemStack getVein(){
+    public ItemStack getVein()
+    {
         return vein;
     }
 }
--- a/src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
+++ b/src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
@@ -1,8 +1,10 @@
 package net.dries007.tfc.api.events;
 
 import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.item.ItemStack;
 import net.minecraft.util.math.BlockPos;
 import net.minecraftforge.fml.common.eventhandler.Event;
+import net.minecraftforge.fml.relauncher.Side;
 
 /**
  * Event fired when the Prospector's Pickaxe is used.
@@ -12,14 +14,14 @@ import net.minecraftforge.fml.common.eventhandler.Event;
 public abstract class ProspectEvent extends Event {
 
     public static class Server extends ProspectEvent {
-        public Server(EntityPlayer player, BlockPos pos, ResultType type, String ore, double score){
-            super(player, pos, type, ore, score);
+        public Server(EntityPlayer player, BlockPos pos, ResultType type, ItemStack vein){
+            super(Side.SERVER, player, pos, type, vein);
         }
     }
 
     public static class Client extends ProspectEvent {
-        public Client(EntityPlayer player, BlockPos pos, ResultType type, String ore, double score){
-            super(player, pos, type, ore, score);
+        public Client(EntityPlayer player, BlockPos pos, ResultType type, ItemStack vein){
+            super(Side.CLIENT, player, pos, type, vein);
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
```diff
@@ -0,0 +1,56 @@
+package net.dries007.tfc.api.events;
+
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.util.math.BlockPos;
+import net.minecraftforge.fml.common.eventhandler.Event;
+
+/**
+ * Event fired when the Prospector's Pickaxe is used.
+ * Carries all the data relative to the result displayed to the player.
+ * One of the two subclasses will be used according to the logical side.
```
alcatrazEscapee, line 14: This is silly. Just add a `side` variable to the event with the relevant accessor.

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
```diff
@@ -0,0 +1,56 @@
...
+        ResultType(String translation){
+            this.translation = translation;
+        }
+    }
+
+    public final EntityPlayer player;
+    public final BlockPos pos;
+    public final ResultType type;
+    public final String ore;
+    public final double score;
```
alcatrazEscapee: Encapsulate these.

src/main/java/net/dries007/tfc/network/PacketProspectResult.java
```diff
@@ -0,0 +1,95 @@
...
+        this.score = score;
+    }
+
+    @Override
+    public void fromBytes(ByteBuf buf)
+    {
+        int x = buf.readInt();
+        int y = buf.readInt();
+        int z = buf.readInt();
+        this.pos = new BlockPos(x, y, z);
```
alcatrazEscapee: Consistent with everywhere else that does this, store the `BlockPos` on the packet and use `BlockPos#toLong()` and `BlockPos#fromLong()` for serialization.

To serialize the `ProspectResult` (see other comment), `ByteBufUtils` has `readItemStack` and `writeItemStack`

src/main/java/net/dries007/tfc/network/PacketProspectResult.java
```diff
@@ -0,0 +1,95 @@
...
+        this.type = ProspectEvent.ResultType.values()[buf.readByte()];
```
alcatrazEscapee: This does an unnecessary array allocation just to access the inner value, a `valueOf(int i)` method would be preferred here. See `Gem.Grade#valueOf` for example.

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    public static class Client extends ProspectEvent {
+        public Client(EntityPlayer player, BlockPos pos, ResultType type, String ore, double score){
+            super(player, pos, type, ore, score);
+        }
+    }
+
+    public enum ResultType {
```
alcatrazEscapee: This should be in `ItemProPick` with `ProspectResult`, not here.

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
```diff
@@ -0,0 +1,56 @@
...
+
+        ResultType(String translation){
+            this.translation = translation;
+        }
+    }
+
+    public final EntityPlayer player;
+    public final BlockPos pos;
+    public final ResultType type;
+    public final String ore;
```
alcatrazEscapee: First off, this `String ore` here is completely unknown as to what the actual value is. Turns out, it's the *translation key for the item stack which wraps the block that is found*. Which is a horrible API to expose on the other end. So, rather than that, add the `ProspectResult` as a field to this event and ditch `score` and `ore`. If you so choose to use the translation key as your identifier on client, be it on your own head.

src/main/java/net/dries007/tfc/objects/items/metal/ItemProspectorPick.java
```diff
@@ -74,7 +78,7 @@ public EnumActionResult onItemUse(EntityPlayer player, World worldIn, BlockPos p
                 if (!targetStack.isEmpty())
                 {
                     // Just clicked on an ore block
-                    player.sendStatusMessage(new TextComponentTranslation("tfc.propick.found").appendText(" ").appendSibling(new TextComponentTranslation(targetStack.getTranslationKey() + ".name")), ConfigTFC.Client.TOOLTIP.propickOutputToActionBar);
+                    event = new ProspectEvent.Server(player, pos, ProspectEvent.ResultType.FOUND, targetStack.getTranslationKey(), 0);
```
Rongmario: I like the work you put into this, it's a clean implementation, but...

You could have just done this on the world.isRemote branch, and avoid sending packets.

Rongmario (changes requested):

src/main/java/net/dries007/tfc/objects/items/metal/ItemProspectorPick.java
alcatrazEscapee: No. Vein information shouldn't be required on both sides.

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
LordFokas: This is actually standard stuff. I didn't come up with this by myself.
This way you can listen to events only on the side that matters...

However I do agree there should be a `side` attribute there, in case you want to listen to the `ProspectEvent` superclass. In my particular case, I don't, and that's the cleaner way to do it (avoids having to always check if side is a specific value when you're only interested in one branch; having 2 event subclasses introduces no overhead).

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
LordFokas: Right, my bad.

src/main/java/net/dries007/tfc/objects/items/metal/ItemProspectorPick.java
LordFokas: Server side does game logic, client side does presentation. There's no argument there.

src/main/java/net/dries007/tfc/network/PacketProspectResult.java
LordFokas: Will do. I didn't realize the values array is cloned every time. TIL.

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
LordFokas: Well yeah, but actually no. ProspectResult is not exposed as API, I cannot refer to that from the API package unless it gets moved to the API package as well.

src/main/java/net/dries007/tfc/network/PacketProspectResult.java
LordFokas: `BlockPos#toLong()`, yes. Serialize the `ProspectResult`, maybe. I'll have to rework some of this, let me give it some more thought.

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
LordFokas: Yes, that makes some sense.

As I said in the `ResultType` thing, I'm not going to reference internal classes in the API package. Ideally you should be able to compile a JAR with that package alone and build your addons against that without having to reference the entire TFC codebase.

So either the `ProspectResult` becomes an API class (and is moved out of ProPick), or, which is what I'm most inclined to do here, I'll remove `String ore` in favor of `ItemStack vein`.

LordFokas:
@Rongmario @alcatrazEscapee thanks for the feedback. In ~8h I expect to be able to sit down, put some more work in this, and request new reviews.

Some of these mistakes were rather silly, I'm not used to contributing to external projects, and in my own I do **a lot** of prototyping which breeds some bad habits (passing strings everywhere, for example), bear with me here 😅

Rongmario:
It's all good, the Minecraft+Forge ecosystem is huge, and a lot of us aren't aware of it all yet.

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
alcatrazEscapee: While in theory you are correct, our api package as it stands is not independent of the mod, we don't publish a separate API artifact, and there's no requirement to keeping them separate (this was decided very early on in development). It's purely an organizational thing.

LordFokas:
@alcatrazEscapee I purposefully didn't move ResultType from the event to the propick class, as it doesn't feel right to me. If it really matters to you then I suggest we discuss this deeper, either here or through Discord, hopefully with a few more members of the dev team involved.

As always, I await your feedback and we can discuss all this as much as necessary.

Oh I forgot to mention in the commit, I dropped the score attribute as it really is some internal value that has no place at this level.

src/main/java/net/dries007/tfc/api/events/ProspectEvent.java
alcatrazEscapee: >  I cannot refer to that from the API package unless it gets moved to the API package as well.

As I explained in another comment: We have no separate API artifact, the API package is *not* independent from the rest of the code base, it's not a separate source set, and we have no plans to make it so. You can say this all you want, it doesn't make it true.

LordFokas: Well, if it is not independent, and therefore you cannot abstract from the rest of the codebase, then it is not an _**Abstract** Programming Interface_, right now calling anything else, say, `foo` is more appropriate, at least `foo` isn't a deceiving name. Don't interpret me wrong here, I don't mean to sound like an arrogant ass, but this _does_ bother me.

So right now, like Tinker's Construct has (or at least used to have), what you have here is not an API but a library to interact with the mod. I'd suggest either fixing it to be what the name says it is, or renaming it to what it actually is (lib).

This being said, I have a couple questions:
- Is the existence of this class in the event a major blocker?
- Is this not-actually-an-API decision something that's set in stone? or do you guys simply not care for it? why was this path chosen? As someone who has to code against the entire codebase I'd rather not have to, being able to code against the API alone and then run against the entire thing would ease some pains and make everything more flexible on this side. I would guess others like me would agree.

Please note that I'm not trying to question your decisions, or force you to reconsider them, but if I am to line up with TFC's expectations I need to understand how things are done, and most importantly, **why.**

Thank you for your time :)

alcatrazEscapee: Your botherment is duly noted.

To answer your questions,
- Yes, it makes more sense to be in `ItemProPick`.
- Yes, since multiple years ago I can't find the exact date.
- Nope.
- Because the effort to maintain an *actual* separate source set and artifact wasn't deemed worth it given the effort required, and the fact that the majority of dependencies would be addons which tend to want to compile against the whole mod for other reasons.

> I would guess others like me would agree.

Citation needed - in a year of addon development this is the first complaint I'm hearing about the matter.

LordFokas: Very well, this is the way.

LordFokas:
I've made all the requested changes. Unless I missed anything else, my work here is done :)

alcatrazEscapee (approved):
Code style is all over the place - if you're using intellij there's a code style xml in the repo that can auto-fix everything - but that can also be fixed during merge.

LordFokas:
> Code style is all over the place - if you're using intellij there's a code style xml in the repo that can auto-fix everything - but that can also be fixed during merge.

Code style is fine, respecting the exact project specification, I don't know what you mean 😅

Issue #1075, opened 2020-04-22
TFCT: Anvil will complain items aren't hot enough to weld when they are


Gaelmare:
https://clips.twitch.tv/AssiduousCrazyElephantOMGScoots as an example

**Describe the bug**
Get a message "The item(s) are not hot enough to work" when welding them actually does work

**To Reproduce**
1. Place two weldable temperature ingots onto an appropriate tier anvil
2. Smack with hammer, hear sound of weld.
3. See warning message that is incorrect

 - TFC Version: build 133

Kalzeroth:
Can attest I saw it as well

alcatrazEscapee:
Needs update to confirm post v1.0.11.140

DisasterMoo:
Closing since it seems to be fixed. If anyone have this bug again, feel free to reopen / tell us in discord.

Gaelmare:
https://clips.twitch.tv/RealTemperedDolphinFloof

Still happening as of build 144

EntropyParadigm:
Version: TerraFirmaCraft-MC1.12.2-1.5.5.155

Still happening in 155 - Tungsten Steel - Cannot "weld" or "work" metal, including the sheet versions:
![2020-10-21_14 28 31](https://user-images.githubusercontent.com/32775097/96790119-1bb34180-13ab-11eb-8180-2550f1ecb566.png)
![2020-10-21_13 41 28](https://user-images.githubusercontent.com/32775097/96790124-1ce46e80-13ab-11eb-932b-0f9b67b4b193.png)
![2020-10-21_14 13 50](https://user-images.githubusercontent.com/32775097/96790125-1e159b80-13ab-11eb-8780-b129fe2207b9.png)
![2020-10-21_14 17 53](https://user-images.githubusercontent.com/32775097/96790131-1eae3200-13ab-11eb-9355-4de94985d62e.png)
![2020-10-21_14 28 16](https://user-images.githubusercontent.com/32775097/96790134-1fdf5f00-13ab-11eb-8a1b-fc87ddf1e27f.png)

No work around for armor crafting - but can use Tinkers to liquify the metals and pour them into casts from TFC - this however doesn't give the smithing skill bonus to the item.

alcatrazEscapee:
This issue does not apply, or has been fixed in the current MC 1.18 versions of TFC, and as MC 1.12 is receiving no more development, it will be closed. If this issue arises again in 1.18, please file a new issue report.

PR #1648, opened 2020-12-12
Make anvil welding output into the actionbar instead & add actionbar tooltip when a vessel is full of solid metal

Charnuz:
Currently when you try to open a vessel full of solid metal it can seem like the game isn't responding, and this should fix that. Also, reduced the chat spam of anvil welding messages.

Didn't change metalworking messages as that doesn't make sense.

These will need non-english localization:
- `tfc.vessel.liquid_solid`
- `config.tfc.client.tooltip.vesselOutputToActionBar`
- `config.tfc.client.tooltip.vesselOutputToActionBar.tooltip`
- `config.tfc.client.tooltip.anvilWeldOutputToActionBar`
- `config.tfc.client.tooltip.anvilWeldOutputToActionBar.tooltip`


src/main/java/net/dries007/tfc/objects/te/TEAnvilTFC.java
```diff
@@ -335,7 +335,7 @@ public boolean attemptWelding(EntityPlayer player)
             if (fluxStack.isEmpty())
             {
                 // No flux
-                player.sendMessage(new TextComponentTranslation(MOD_ID + ".tooltip.anvil_no_flux"));
+                player.sendStatusMessage(new TextComponentTranslation(MOD_ID + ".tooltip.anvil_no_flux"), ConfigTFC.Client.TOOLTIP.anvilWeldOutputToActionBar);
```
alcatrazEscapee: So as I learned as per #1647, this will actually not work on a dedicated server, as the client config will be referenced from the server, ergo not really serving the intended function. So if this is to be desired, it needs to be implemented via packet and the config checked on client.

alcatrazEscapee (changes requested):
Seems reasonable, minus the fact it won't work as a client config on dedicated servers.

src/main/java/net/dries007/tfc/objects/te/TEAnvilTFC.java
LordFokas: We can probably make a general purpose packet for "send this data to client so that they display a message in the appropriate place as per local configs" that can be used across the board...

... or maybe even intercept these messages on the client side and override where they are meant to be displayed?

Charnuz: So while I was doing some testing, turns out the tile entity has the welding method called by both client and server, but doesn't receive the world along with the player meaning you can't check if its the server so you can send the packet via SimpleNetworkWrapper. 

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.12.x/src/main/java/net/dries007/tfc/objects/blocks/metal/BlockAnvilTFC.java#L213

Charnuz: This also explains why welding error messages are received twice

alcatrazEscapee: `player.world`?

Charnuz: oh duh, I forgot that. Is there any reason we need welding to occur on both sides? (besides sound + particles)

alcatrazEscapee: It really shouldn't be, probably the cause of #1075

Charnuz:
Did test on both single player and with a dedicated server, appeared to work as expected

src/main/java/net/dries007/tfc/objects/te/TEAnvilTFC.java
```diff
@@ -44,7 +50,7 @@
 import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
 
 @ParametersAreNonnullByDefault
-public class TEAnvilTFC extends TEInventory
+public class TEAnvilTFC extends TEInventory implements ISidedInventory
```
alcatrazEscapee: This is 100% no. We do not use `ISidedInventory`, nor `IInventory`.

src/main/java/net/dries007/tfc/network/PacketBlockInventoryUpdate.java
```diff
@@ -0,0 +1,100 @@
...
+            final EntityPlayer player = TerraFirmaCraft.getProxy().getPlayer(ctx);
+            if (player != null)
+            {
+                TerraFirmaCraft.getProxy().getThreadListener(ctx).addScheduledTask(() -> {
+                    World world = player.getEntityWorld();
+                    TileEntity te = world.getTileEntity(message.pos);
+                    if (te instanceof IInventory && !message.changes.isEmpty())
+                    {
+                        // Fill the slots with the appropriate items
+                        IInventory inventory = (IInventory) te;
```
alcatrazEscapee: No no no no no no no no. Capabilities, not `IInventory`

src/main/java/net/dries007/tfc/network/PacketBlockInventoryUpdate.java
```diff
@@ -0,0 +1,100 @@
...
+            changes.put(slot, item);
+        }
+    }
+
+    @Override
+    public void toBytes(ByteBuf buf)
+    {
+        buf.writeInt(pos.getX());
+        buf.writeInt(pos.getY());
+        buf.writeInt(pos.getZ());
```
alcatrazEscapee: Can be `pos.toLong()`

src/main/java/net/dries007/tfc/network/PacketSimpleMessage.java
```diff
@@ -0,0 +1,134 @@
...
+import net.minecraft.util.text.Style;
+import net.minecraft.util.text.TextComponentString;
+import net.minecraft.util.text.TextComponentTranslation;
+import net.minecraft.util.text.TextFormatting;
+import net.minecraftforge.fml.common.network.simpleimpl.IMessage;
+import net.minecraftforge.fml.common.network.simpleimpl.IMessageHandler;
+import net.minecraftforge.fml.common.network.simpleimpl.MessageContext;
+import stanhebben.zenscript.util.ArrayUtil;
+
+public class PacketSimpleMessage implements IMessage
```
alcatrazEscapee, line 22: Rather than this rather convoluted and difficult to maintain serialization, it would be much better to use vanilla's built-in handling for this. `ITextComponent` has a serialize to JSON, which we can read + write to a string, which can be written to a packet. Much better than individually trying to serialize every text component field.

src/main/java/net/dries007/tfc/ConfigTFC.java
```diff
@@ -958,6 +958,10 @@ public static void onConfigChangedEvent(ConfigChangedEvent.OnConfigChangedEvent
             @Config.LangKey("config." + MOD_ID + ".client.tooltip.propickOutputToActionBar")
             public boolean propickOutputToActionBar = true;
 
+            @Config.Comment("Tooltips with any of the following categories will output to the actionbar (the space just above the hotbar)")
+            @Config.LangKey("config." + MOD_ID + ".client.tooltip.tooltipCategoriesOutputToActionBar")
+            public String[] tooltipCategoriesOutputToActionBar = new String[] { "anvil", "vessel", "propick" };
```
alcatrazEscapee: This is redundant when we already have config values for each of these. Just use those rather than have to invent parsing. (Never mind the fact this is a completely opaque config option that @Claycorp would yell at me for since there's no indication to the user *what* values are valid here. Just use the existing booleans.

alcatrazEscapee (changes requested):
Remove `IInventory` introduction and de-contrive the `ITextComponent` serialization.

src/main/java/net/dries007/tfc/ConfigTFC.java
Charnuz: Originally I thought it would be good to avoid having possibly many options for this and reduce it to a single config value, but you are right that it is very obscure

src/main/java/net/dries007/tfc/network/PacketBlockInventoryUpdate.java
Charnuz: Sorry I am still not super familiar with certain aspects of modern minecraft modding

PR #1652, opened 2020-12-17
The Second Great Plants PR

eerussianguy:
New Classes:

Block/Item stuff:

- BodyPlantBlock -- vanilla AbstractBodyPlantBlock (weeping/hanging vines) with a supplier for it's semantic pair
- DryPlantBlock -- TFC's PlantBlock but it explicitly allows placement on sand
- KelpTreeBlock -- ChorusPlantBlock, the stem of the giant kelp
- KelpTreeFlowerBlock -- ChorusFlowerBlock, the logical block of the giant kelp. Contains growth logic
- TFCKelpBlock -- underwater version of BodyPlantBlock
- TFCKelpTopBlock -- underwater version of TopPlantBlock
- TFCSeagrass Block -- essentially TFC's short grass block but inheriting from WaterPlantBlock for ease of use
- TFCVineBlock -- vanilla vine that updates with season
- TallWaterPlantBlock -- double tall water plant, currently only used for emergent plants and the logic reflects that
- TopPlantBlock -- logical block for twisting/weeping plants with plenty of abstraction
- WaterPlantBlock -- simple PlantBlock with a fluid property attached
- Coral -- helper class to register coral plants that are saltwater loggable
- TFCAbstractCoralPlantBlock, TFCCoralFanBlock, TFCCoralFinBlock, TFCCoralPlantBlock, TFCCoralWallFanBlock, TFCDeadCoralPlantBlock, TFCDeadCoralWallFanBlock -- all vanilla copypastes in order to add FluidProperty
- TFCSeaPickleBlock -- vanilla sea pickle that's fluidloggable

Worldgen:
- EmergentPlantPlacer -- simply calls TallWaterPlantBlock's placing helper
- FacingHorizontalBlockStateProvider -- rotates groundcover for worldgen
- TFCBlockPlacerTypes -- registration
- TFCBlockStateProviderTypes -- registration
- TallPlantPlacer -- calls TFCTallGrassBlock's placing helper
- WaterPlantPlacer -- fluidlogging helper
- ClimateTargetConfig/Decorator -- climate-restricting decorator that fades out to zero at the end of the specified range. Spread parameter corresponds to an exact +- for temperature and the distance divided by 20 for rainfall. If outside of either temperature or rainfall requirements, it won't spawn at all.
- ShorelineDecorator -- checks within a radius for a bush plantable / sea bush plantable block that has air above it and water beside it
- LooseRocksFeature -- places loose rocks of random size throughout the world
- CavePatchFeature -- place blocks randomly but cave-smart
- CaveVegetationConfig/Feature -- mossify cave floors, add cobwebs, add additional springs
- IceCaveFeature -- add ice drifts + spikes to caves
- TFCCoralClawFeature / TFCCoralMushroomFeature / TFCCoralTreeFeature / TFCCoralFeature -- copypasta
- EmergentPatchFeature -- extra checks for emergent placement
- KelpTreeFeature -- chorus feature with some handling
- RandomPatchDensityFeature -- random patch but it is denser in forests /heightmap aware
- RandomPatchWaterFeature -- title
- RandomPatchWaterLandFeature -- water agnostic feature (used for groundcover)
- TFCKelpFeature/TFCTwistingVinesFeature/TFCWeepingVinesFeature -- self explanatory
- TFCVinesFeature/VineConfig -- does X attempts per y level in the range specified and then carries the vines downwards
- TallPlantConfig -- used for any paired plant pretty much
- TFCBlocks -- new icicle block (calcite)
- TFCBlockStateProperties -- new saltwater/freshwater fluid properties

Tweaked Classes:

Block/item stuff:
- CreepingPlantBlock -- now checks a tag for canSurvive
- FloatingWaterPlantBlock -- takes a fluid supplier for fresh/salt variation, and breaks when hit by a boat
- ITallPlant stripped of almost all functionality as double-tall plants were made to work like vanilla
- Plant -- has a transform() function to let paired plants refer to each other
- Plant -- added helper functions for when blocks/items need coloring, default properties, bounding boxes
- Plant -- broadened scope of BiFunction
- TFCCactusBlock -- added canSurvive check for sand
- TFCTallGrassBlock -- simplified to not need ITallPlant as much
- IMossGrowingBlock (and implementations) -- added needsWater boolean to potentially be able to spread moss programatically on worldgen, currently unused
- TFCItems -- added coral item registration (needs special items to hide the fact that vanilla decided to use two separate blocks where one is fine)
- ClientEventHandler -- intelligent leaf / grass coloration for plants + plant items
- TFCTags - new tags (see list)
- CalciteBlock -- checks for base stone tag instead of canSurvive

Worldgen:
- ClimateConfig/Decorator -- added needsForest parameter
- IceAndSnowFeature -- attempts to convert snow pileable blocks to snow piles (half working)
- CalciteConfig/Feature -- added state parameter (to allow for more calcite types)
- CaveSpikesFeature -- sanity check for the state (unused)
- ForestConfig -- added params for log, leaves, twig, fallen leaves, for placement
- ForestFeature -- added bushes, fallen leaves and twigs
- ForestFeature -- Re-added dominance in the form of prioritizing trees by distance from their climate target. Has the effect of removing the '5 tree types in one chunk' thing.
- VeinFeature -- places on the ocean floor w/ random rotation now

New Plants:
- Badderlocks
- Cattail
- Coontail
- Dead Bush
- Eel Grass
- Gutweed
- Laminaria 
- Manatee Grass
- Marigold
- Milfoil
- Sago
- Star Grass
- Turtle Grass
- Hanging Vines (pair)
- Liana (pair)
- Tree Fern (pair)
- Arundo (pair)
- Winged Kelp (pair)
- Leafy Kelp (pair)
- Giant Kelp (pair)
- Ivy
- Jungle Vines

New Plant Types:
- Dry
- Vine
- Weeping (pair)
- Twisting (pair)
- Solid Twisting (pair)
- Kelp
- Kelp Tree
- Floating (fresh/salt)
- Tall Water (fresh/salt)
- Water (fresh/salt)
- Water Grass (fresh/salt)

New Features (not configured)
- Loose Rocks
- Water-Land Patch
- Random Patch Density
- Emergent Patch
- Water Patch
- Hanging Vines
- Twisting Vines
- Kelp
- Kelp Tree
- Coral Claw / Mushroom / Tree
- Cave Vegetation
- Cave Patch
- Vines
- Ice Cave

New Tags:
- plant
- sea_bush_plantable_on
- creeping_plantable_on
- kelp_tree
- kelp_flower
- kelp_branch
- wall_corals
- corals


resources/data.py
```diff
@@ -54,23 +54,42 @@ def generate(rm: ResourceManager):
             'beach_sand_color': rock_data.beach_sand_color
         })
 
-        rm.block_tag('minecraft:base_stone_overworld', 'tfc:rock/raw/%s' % rock)  # used by vanilla, provided for consistiency
+        rm.block_tag('creeping_plantable_on', 'tfc:rock/raw/%s' % rock)
```
alcatrazEscapee: What might be better for some of these tags, is to have generic tags for stuff like `stone`, `gravel`, etc. (or using forge/minecraft tags if they exist, or should do), and then referencing those tags indirectly rather than having to include every rock block variant for each one of these tags. Should be better overall for out of the box compatability.

resources/data.py
```diff
@@ -54,23 +54,42 @@ def generate(rm: ResourceManager):
...
+        rm.block_tag('sea_bush_plantable_on', 'tfc:sand/%s' % color)
+        rm.block_tag('minecraft:sand', 'tfc:sand/%s' % color)
+
     # Tags
     rm.item_tag('forge:ingots/cast_iron', 'minecraft:iron_ingot')
     rm.block_tag('tree_grows_on', 'minecraft:grass_block', '#forge:dirt', '#tfc:grass')
     rm.block_tag('supports_landslide', 'minecraft:grass_path')
     rm.block_tag('bush_plantable_on', 'minecraft:grass_block', '#forge:dirt', '#tfc:grass')
+    rm.block_tag('sea_bush_plantable_on', '#forge:dirt')
+    rm.block_tag('creeping_plantable_on', 'minecraft:grass_block', '#tfc:grass', '#minecraft:base_stone_overworld')
```
alcatrazEscapee: See above - if this includes `#minecraft:base_stone_overworld`, which includes all rock types, there's no need to explicitly include all rock types in the `creeping_plantable_on` tag.

resources/world_gen.py
```diff
@@ -140,25 +165,25 @@ def generate(rm: ResourceManager):
...
-            forest_config(240, 450, -9, 11, 'oak', False),
-            forest_config(180, 470, 20, 32, 'palm', False),
-            forest_config(60, 270, -18, -4, 'pine', True),
-            forest_config(140, 310, 8, 31, 'rosewood', False),
-            forest_config(250, 420, -14, 2, 'sequoia', True),
-            forest_config(110, 320, -17, 1, 'spruce', True),
-            forest_config(230, 480, 15, 29, 'sycamore', True),
-            forest_config(10, 220, -13, 9, 'white_cedar', True),
-            forest_config(330, 500, 11, 32, 'willow', True),
+            forest_config(30, 210, 17, 32, 'tfc:wood/wood/acacia', 'tfc:wood/leaves/acacia', 'tfc:wood/twig/acacia', 'tfc:wood/fallen_leaves/acacia', 'acacia', True),
```
alcatrazEscapee: This seems unnecessary. Can this not be inferred by the `forest_config` function given the wood type?

resources/world_gen.py
```diff
@@ -253,76 +278,153 @@ def generate(rm: ResourceManager):
...
     }))
 
-    # Plants
+    rm.feature('cave_vegetation', wg.configure_decorated(wg.configure('tfc:cave_vegetation', {
+        'blocks': [{
+            'stone': ['tfc:rock/raw/%s' % rock],
+            'ore': [{'block': 'tfc:rock/mossy_cobble/%s' % rock, 'weight': 8},
+                    {'block': 'tfc:rock/cobble/%s' % rock, 'weight': 2}]
+        } for rock in ROCKS.keys()]
+    }), decorate_climate_target(24, 310, 8), air_mask(0.01, 130), decorate_range(64, 130)))
```
alcatrazEscapee, line 288: This is more of a note for later than an actual change request, but this needs to be changed to implement (what are essentially) cave biomes. We'll likely not have actual biomes, but still need to separate them rather than "all caves in this climate are lush". Add a `# todo: decorate using noise and/or cave biomes in 1.17` would be good for now.

resources/world_gen.py
```diff
@@ -253,76 +278,153 @@ def generate(rm: ResourceManager):
...
+    rm.feature(('plant', 'reindeer_lichen_cover'), wg.configure_decorated(plant_feature('tfc:plant/reindeer_lichen[age=1,stage=1,up=false,down=true,north=false,east=false,west=false,south=false]', 'minecraft:simple_block_placer', 1, 6), decorate_climate_target(0, 250, 7, True), air_mask(0.002, 130), decorate_range(64, 130)))
+
+    # todo: convert the creeping plant blocks to use the target climate thing natively in the spreadsheet rather than here
+
+    rm.feature('bamboo', wg.configure_decorated(wg.configure('minecraft:bamboo', {'probability': 0.2}), decorate_chance(30), decorate_climate_target(23, 400, 5, True), ('minecraft:count_noise_biased', {
+        'noise_to_count_ratio': 160,
+        'noise_factor': 80.0,
+        'noise_offset': 0.3
+    }), 'minecraft:square', 'minecraft:heightmap_world_surface'))
+    # todo: mixin to bamboo to make it not grow insanely fast / not able to insta grow
```
alcatrazEscapee: Two things:
1. Forge actually has an event for crop growth which is fired for bamboo, which we can probably subscribe to in order to reduce growth speed (configurable ofc) by cancelling at a random chance.
2. Bone meal, in my in-dev crops branch is much simpler to deal with at the source - simply by disabling any and all bonemeal uses (and adding specific handling for adding nutrients, for example to farmland). This is I think the better option, as there's not really any room for insta-grow magic powder anywhere in TFC.

Basically no need here.

resources/world_gen.py
```diff
@@ -253,76 +278,153 @@ def generate(rm: ResourceManager):
...
+        "noise_factor": 400.0,
+        "noise_offset": 0.0
+    }), 'minecraft:square', 'minecraft:top_solid_heightmap'))
+
+    # Groundcover
+    sand = [wg.block_state('tfc:sand/%s' % color) for color in SAND_BLOCK_TYPES]
+    gravel = [wg.block_state('tfc:rock/gravel/%s' % rock) for rock in ROCKS.keys()]
+    raw = [wg.block_state('tfc:rock/raw/%s[supported=false]' % rock) for rock in ROCKS.keys()]
+    grass = [wg.block_state('tfc:grass/%s[north=false,south=false,east=false,west=false,snowy=false]' % soil) for soil in SOIL_BLOCK_VARIANTS]
+    # todo: add these to the deep oceans?
```
alcatrazEscapee: Discuss

resources/world_gen.py
```diff
@@ -374,12 +476,16 @@ def surface_builder(rm: ResourceManager, name: str, surface_builder):
...
+def forest_config(min_rain: float, max_rain: float, min_temp: float, max_temp: float, log: str, leaves: str, twig: str, fallen_leaves: str, tree: str, old_growth: bool):
     cfg = {
         'min_rain': min_rain,
         'max_rain': max_rain,
         'min_temp': min_temp,
         'max_temp': max_temp,
+        'log': log,
+        'leaves': leaves,
+        'twig': twig,
+        'fallen_leaves': fallen_leaves,
```
alcatrazEscapee: This should be able to just string replace here with `tree`, i.e. `'tfc:fallen_leaves/%s' % tree`

resources/world_gen.py
```diff
@@ -253,76 +278,153 @@ def generate(rm: ResourceManager):
...
+    # todo: convert the creeping plant blocks to use the target climate thing natively in the spreadsheet rather than here
+
+    rm.feature('bamboo', wg.configure_decorated(wg.configure('minecraft:bamboo', {'probability': 0.2}), decorate_chance(30), decorate_climate_target(23, 400, 5, True), ('minecraft:count_noise_biased', {
+        'noise_to_count_ratio': 160,
+        'noise_factor': 80.0,
+        'noise_offset': 0.3
+    }), 'minecraft:square', 'minecraft:heightmap_world_surface'))
+    # todo: mixin to bamboo to make it not grow insanely fast / not able to insta grow
+
+    rm.feature('coral', wg.configure_decorated(wg.configure('minecraft:simple_random_selector', {'features': [wg.configure('tfc:coral_tree'), wg.configure('tfc:coral_mushroom'), wg.configure('tfc:coral_claw')]}), decorate_climate_target(25, 300, 5), ('minecraft:count_noise_biased', {
```
alcatrazEscapee, line 394: Another thing that needs to be adjusted: coral placement needs to be a bit better than just noise (which, in warmer oceans, tends to scatter it *everywhere*. Ideally I'd like to have biome-level distinctions for both fringing reefs and barrier reefs. `todo` would be good here for whenever I or someone else gets to that.

src/main/java/net/dries007/tfc/client/ClientEventHandler.java
```diff
@@ -106,7 +112,10 @@ public static void registerColorHandlerBlocks(ColorHandlerEvent.Block event)
         registry.register((state, worldIn, pos, tintIndex) -> TFCColors.getGrassColor(pos, tintIndex), TFCBlocks.SOIL.get(SoilBlockType.CLAY_GRASS).values().stream().map(RegistryObject::get).toArray(Block[]::new));
         registry.register((state, worldIn, pos, tintIndex) -> TFCColors.getGrassColor(pos, tintIndex), TFCBlocks.PEAT_GRASS.get());
         // Plants
-        registry.register((state, worldIn, pos, tintIndex) -> TFCColors.getGrassColor(pos, tintIndex), TFCBlocks.PLANTS.values().stream().map(RegistryObject::get).toArray(Block[]::new));
+        Block[] leafyPlants = Stream.of(Plant.values()).filter(Plant::isLeafColored).map(p -> TFCBlocks.PLANTS.get(p).get()).toArray(Block[]::new);
+        Block[] grassyPlants = Stream.of(Plant.values()).filter(p -> !p.isLeafColored()).map(p -> TFCBlocks.PLANTS.get(p).get()).toArray(Block[]::new);
+        registry.register((state, worldIn, pos, tintIndex) -> TFCColors.getSeasonalFoliageColor(state, pos, tintIndex, Wood.Default.HICKORY.getFallFoliageCoords()), leafyPlants);
```
alcatrazEscapee: Having this delegate to hickory specifically is a bit weird imo, either move that to a static reference for plants or add `Plant#getFallFoliageCoords()`

src/main/java/net/dries007/tfc/common/blocks/CalciteBlock.java
```diff
@@ -92,7 +93,7 @@ public boolean canSurvive(BlockState state, IWorldReader worldIn, BlockPos pos)
     {
         BlockPos abovePos = pos.above();
         BlockState aboveState = worldIn.getBlockState(abovePos);
-        return (aboveState.getBlock() == this && !aboveState.getValue(TIP)) || aboveState.isFaceSturdy(worldIn, abovePos, Direction.DOWN);
+        return (aboveState.getBlock() == this && !aboveState.getValue(TIP)) || aboveState.is(BlockTags.BASE_STONE_OVERWORLD) || aboveState.is(BlockTags.ICE);
```
alcatrazEscapee: This should probably stay with the validity check here (i.e. "is face sturdy") for calcite. If ice does not pass that check, it should probably be adjusted just for ice (e.g. calcite should not survive on ice. This also applies if we want to make icicles accumulate in winter storms (*which would be bloody cool!*). If this causes issues during world gen that should be handled by the feature imo.

src/main/java/net/dries007/tfc/common/blocks/TFCBlockStateProperties.java
```diff
@@ -38,6 +38,8 @@
 
     public static final FluidProperty WATER = FluidProperty.create("fluid", Stream.of(Fluids.EMPTY, Fluids.WATER, TFCFluids.SALT_WATER));
     public static final FluidProperty WATER_AND_LAVA = FluidProperty.create("fluid", Stream.of(Fluids.EMPTY, Fluids.WATER, TFCFluids.SALT_WATER, Fluids.LAVA));
+    public static final FluidProperty SALTWATER = FluidProperty.create("fluid", Stream.of(Fluids.EMPTY, TFCFluids.SALT_WATER));
+    public static final FluidProperty FRESHWATER = FluidProperty.create("fluid", Stream.of(Fluids.EMPTY, Fluids.WATER));
```
alcatrazEscapee: `SALT_WATER` and `FRESH_WATER`

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -72,6 +74,7 @@
 
     public static final RegistryObject<SnowPileBlock> SNOW_PILE = register("snow_pile", () -> new SnowPileBlock(new ForgeBlockProperties(Properties.copy(Blocks.SNOW)).tileEntity(SnowPileTileEntity::new)), EARTH);
     public static final RegistryObject<CalciteBlock> CALCITE = register("calcite", () -> new CalciteBlock(Properties.of(Material.GLASS).noDrops().instabreak().sound(SoundType.GLASS)));
+    public static final RegistryObject<CalciteBlock> ICICLE = register("icicle", () -> new CalciteBlock(Properties.of(Material.ICE).noDrops().instabreak().sound(SoundType.GLASS).noOcclusion()));
```
alcatrazEscapee: Should rename `CalciteBlock` then, given it's not just for calcite. `ThinSpikeBlock` imo.

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -137,15 +139,27 @@
...
+            register("coral/" + color.toString().toLowerCase() + "_" + type.toString().toLowerCase(), type.create(color), block -> type.createBlockItem(block, new Item.Properties().tab(FLORA)), type.needsItem())
+        )
     );
 
+    public static final RegistryObject<Block> SEA_PICKLE = register("sea_pickle", () -> new TFCSeaPickleBlock(AbstractBlock.Properties.of(Material.WATER_PLANT, MaterialColor.COLOR_GREEN)
+        .lightLevel((state) -> 3 + 3 * state.getValue(SeaPickleBlock.PICKLES)).sound(SoundType.SLIME_BLOCK).noOcclusion()), FLORA);
+    // see the registration for pickles in Blocks for how this should work (it should not light if it's not waterlogged)
+    // but since keys for FluidProperty can only be obtained by querying a fluid (even if it's Fluids.EMPTY)
+    // rather than how Waterlogged is a boolean
+    // it's not readily accessible at runtime in the way we want.
```
alcatrazEscapee: This doesn't make sense. It should be fine to add a static method in `TFCSeaPickleBlock`, i.e.
```java
public static void isDead(BlockState state) {
    FluidProperty property = ((TFCSeaPickleBlock)state.getBlock()).getFluidProperty();
    return state.getValue(property) == property.keyFor(Fluids.EMPTY);
}
```

src/main/java/net/dries007/tfc/common/blocks/plant/BodyPlantBlock.java
```diff
@@ -0,0 +1,49 @@
...
+    protected AbstractTopPlantBlock getHeadBlock()
+    {
+        return (AbstractTopPlantBlock) headBlock.get();
+    }
+
+    @Override
+    public boolean isValidBonemealTarget(IBlockReader worldIn, BlockPos pos, BlockState state, boolean isClient)
+    {
+        return false;
+    }
```
alcatrazEscapee, line 36: There's ~4 methods for bonemeal, as per `IGrowable`. For consistency if we override one in the negative, we should no-op or return false on all. (Despite the aforementioned comment about indev bonemeal handling.

src/main/java/net/dries007/tfc/common/blocks/plant/BodyPlantBlock.java
```diff
@@ -0,0 +1,49 @@
...
+{
+    private final Supplier<? extends Block> headBlock;
+
+    public BodyPlantBlock(AbstractBlock.Properties properties, Supplier<? extends Block> headBlock, VoxelShape shape, Direction direction)
+    {
+        super(properties, direction, shape, false);
+        this.headBlock = headBlock;
+    }
+
+    protected AbstractTopPlantBlock getHeadBlock()
```
alcatrazEscapee, line 27: Missing `@Override`

src/main/java/net/dries007/tfc/common/blocks/plant/FloatingWaterPlantBlock.java
```diff
@@ -38,12 +51,26 @@ protected FloatingWaterPlantBlock(Properties properties)
...
+        }
     }
 
     @Override
     public VoxelShape getShape(BlockState state, IBlockReader worldIn, BlockPos pos, ISelectionContext context)
     {
         return SHAPE;
     }
+
+    public abstract Fluid getFavoriteFluid();
```
alcatrazEscapee: 1. Change this to a predicate method on a given fluid (i.e. `isValidFluid(FluidState fluid)`)
2. In the implementation of this method, check the saved fluid against the passed in fluid.

I also don't see the case why this has to be a abstract method implemented by anon. subclass (i.e. `getPlant()` exists for a very specific call path which happens during the `Block` super constructor). Is there a reason this can't just be a property on the class?

src/main/java/net/dries007/tfc/common/blocks/plant/KelpTreeBlock.java
```diff
@@ -0,0 +1,167 @@
...
+                return fluid;
+            }
+        };
+    }
+
+    protected KelpTreeBlock(AbstractBlock.Properties builder)
+    {
+        super(0.3125F, builder);
+        registerDefaultState(stateDefinition.any().setValue(NORTH, Boolean.FALSE).setValue(EAST, Boolean.FALSE).setValue(SOUTH, Boolean.FALSE).setValue(WEST, Boolean.FALSE).setValue(UP, Boolean.FALSE).setValue(DOWN, Boolean.FALSE)
+        .setValue(getFluidProperty(), getFluidProperty().keyFor(Fluids.EMPTY)));
```
alcatrazEscapee: This is a weird newline

src/main/java/net/dries007/tfc/common/blocks/plant/KelpTreeBlock.java
```diff
@@ -0,0 +1,167 @@
...
+        Block eastBlock = world.getBlockState(pos.east()).getBlock();
+        Block southBlock = world.getBlockState(pos.south()).getBlock();
+        Block westBlock = world.getBlockState(pos.west()).getBlock();
+        return this.defaultBlockState()
+            .setValue(DOWN, downBlock.is(TFCTags.Blocks.KELP_TREE) || downBlock.is(TFCTags.Blocks.SEA_BUSH_PLANTABLE_ON))
+            .setValue(UP, upBlock.is(TFCTags.Blocks.KELP_TREE))
+            .setValue(NORTH, northBlock.is(TFCTags.Blocks.KELP_TREE))
+            .setValue(EAST, eastBlock.is(TFCTags.Blocks.KELP_TREE))
+            .setValue(SOUTH, southBlock.is(TFCTags.Blocks.KELP_TREE))
+            .setValue(WEST, westBlock.is(TFCTags.Blocks.KELP_TREE));
```
alcatrazEscapee: This could be a `for (Direction direction : Direction.values())` loop and it would be cleaner imo.

src/main/java/net/dries007/tfc/common/blocks/plant/KelpTreeBlock.java
```diff
@@ -0,0 +1,167 @@
...
+    {
+        updateFluid(worldIn, state, pos);
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public boolean canSurvive(BlockState state, IWorldReader worldIn, BlockPos pos)
+    {
+        BlockState belowState = worldIn.getBlockState(pos.below());
+        //boolean attachedVertically = !KelpTreeFlowerBlock.isEmptyWaterBlock(worldIn, pos.above()) && !KelpTreeFlowerBlock.isEmptyWaterBlock(worldIn, pos.below());
```
alcatrazEscapee: Leaving block commented code is bad. Either there's a good explanation and reason for keeping this here (in which case also include said comment), or remove the commented code.

src/main/java/net/dries007/tfc/common/blocks/plant/KelpTreeBlock.java
```diff
@@ -0,0 +1,167 @@
...
+        builder.add(NORTH, EAST, SOUTH, WEST, UP, DOWN);
+    }
+
+    @Override
+    public FluidProperty getFluidProperty()
+    {
+        return getFluidPropertyAbstract();
+    }
+
+    public abstract FluidProperty getFluidPropertyAbstract();
```
alcatrazEscapee: This doesn't make any sense, `getFluidProperty` already is abstract as it's provided by the interface implementation, and this method name is bad. Can just use `getFluidProperty` and remove all uses of `getFluidPropertyAbstract`

src/main/java/net/dries007/tfc/common/blocks/plant/KelpTreeFlowerBlock.java
```diff
@@ -0,0 +1,355 @@
...
+                    else
+                    {
+                        this.placeDeadFlower(worldIn, pos);
+                    }
+                }
+                else
+                {
+                    this.placeDeadFlower(worldIn, pos);
+                }
+                net.minecraftforge.common.ForgeHooks.onCropsGrowPost(worldIn, pos, state);
```
alcatrazEscapee: No need for a FQN here

src/main/java/net/dries007/tfc/common/blocks/plant/KelpTreeFlowerBlock.java
```diff
@@ -0,0 +1,355 @@
...
+
+    public void generatePlant(IWorld worldIn, BlockPos pos, Random rand, int maxHorizontalDistance, Fluid fluid)
+    {
+        if (!getFluidProperty().canContain(fluid))
+            return;
+        setBodyBlockWithFluid(worldIn, pos, fluid);
+        growTreeRecursive(worldIn, pos, rand, pos, maxHorizontalDistance, 0, fluid);
+    }
+
+    public void growTreeRecursive(IWorld worldIn, BlockPos branchPos, Random rand, BlockPos originalBranchPos, int maxHorizontalDistance, int iterations, Fluid fluid)
```
alcatrazEscapee, line 270: Just taking a guess this looks like mojang copy paste - if so, always good to provide a javadoc with an `@link` to the copy pasta'd code so we can check if/when it changes across versions, and cross reference what it's supposed to be.

src/main/java/net/dries007/tfc/common/blocks/plant/KelpTreeFlowerBlock.java
```diff
@@ -0,0 +1,355 @@
...
+
+    @Override
+    public FluidProperty getFluidProperty()
+    {
+        return getFluidPropertyAbstract();
+    }
+
+    public abstract FluidProperty getFluidPropertyAbstract();
+
+    public abstract Supplier<? extends Block> getBodyBlockAbstract();
```
alcatrazEscapee: Neither of these should need to exist, they can just be class level properties.

The hack with abstract properties in `PlantBlock` subclasses (and fluid properties) is that the `Block` constructor assembles block states, which needs references to all properties. This odd combination of `getFluidProperty` and `getFluidPropertyAbstract` is just not nessecary, and I see no use of `getBodyBlockAbstract` that would also predicate this.

Also calling methods `Abstract` is just a bad name overall - the method is never going to be abstract when it's implemented.

src/main/java/net/dries007/tfc/common/blocks/plant/Plant.java
```diff
@@ -72,23 +93,43 @@
...
+    TURTLE_GRASS(BlockType.GRASS_WATER, 0.9F, new int[] {3, 0, 0, 0, 0, 1, 2, 2, 2, 2, 2, 2}),
     VRIESEA(BlockType.EPIPHYTE, 0.8F, new int[] {1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0}),
-    WATER_CANNA(BlockType.FLOATING, 0.8F, new int[] {0, 0, 0, 0, 0, 1, 2, 3, 3, 3, 3, 0}),
-    WATER_LILY(BlockType.FLOATING, 0.8F, new int[] {5, 5, 6, 0, 1, 2, 2, 2, 2, 3, 4, 5}),
-    YUCCA(BlockType.STANDARD, 0.8F, new int[] {0, 0, 0, 0, 0, 1, 2, 2, 2, 2, 2, 3});
+    WATER_CANNA(BlockType.FLOATING_FRESH, 0.8F, new int[] {0, 0, 0, 0, 0, 1, 2, 3, 3, 3, 3, 0}),
+    WATER_LILY(BlockType.FLOATING_FRESH, 0.8F, new int[] {5, 5, 6, 0, 1, 2, 2, 2, 2, 3, 4, 5}),
+    YUCCA(BlockType.DRY, 0.8F, new int[] {0, 0, 0, 0, 0, 1, 2, 2, 2, 2, 2, 3}),
+
+    //not data driven. I added an unused int array because NPE. naming convention is inherited from vanilla
```
alcatrazEscapee: Be much better to just add a `@Nullable` marker and null check in the constructor. This also ties into the fact that there's a number of plants which have a stage property which isn't used - I have this marked in my dev-env as a `//todo`, as we shouldn't be putting extra properties (or properties with unused values) if possible.

src/main/java/net/dries007/tfc/common/blocks/plant/TFCKelpBlock.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    protected AbstractTopPlantBlock getHeadBlock()
+    {
+        return (AbstractTopPlantBlock) getHeadBlockAbstract().get();
+    }
+
+    public abstract FluidProperty getFluidPropertyAbstract();
+
+    public abstract Supplier<? extends Block> getHeadBlockAbstract();
```
alcatrazEscapee: Again, this is a weird pattern and neither of these should need to exist here.

src/main/java/net/dries007/tfc/common/blocks/plant/TFCKelpTopBlock.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        return null;
+    }
+
+    //Dear Alcatraz; I'm not sure what this does
+    //I'm pretty sure this is where it converts the top to the body block so I'm putting the bit of extra handling here for that
+    //Basically the Abstract tall plant system believes that we're _not_ waterlogged (so that we can use our own fluid logging)
+    //It takes 'waterlogged' as a boolean in the constructor which I hardcode to being false in the superclass
+    //So I have to recreate their handling at the top level essentially
+    @Override
```
alcatrazEscapee: I can answer specific questions about `updateShape` in discord. But generally, this updates a specific block w.r.t. a given side which has just been updated. Such as fences connecting to adjacent blocks.

The fluid ticks are odd, shouldn't be scheduling fluid ticks for fluids that don't exist at the current pos (can do `state.getFluidState().getType()` for the fluid at.)

alcatrazEscapee (changes requested):
Didn't get all the way through since a couple comments began to get repetitive. Basic feedback:

- There's a couple weird patterns `getFooAbstract` that shouldn't be needed. Either override the source method (i.e. `getFluidProperty`) in the case where the super constructor needs access to the value prior to subclass constructor class member initialization. Otherwise just make it a class level member.
- Copied mojang code should have a javadoc with `{@link` to where it was sourced so we can maintain it later
- A couple odds and ends with data gen, generally looks fine tho.

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
alcatrazEscapee: Can also remove the unneeded comment here as well then.

PR #1660, opened 2020-12-26
Underground assets rework / vein update

eerussianguy:
Still todo is stuff thats geologically dependent on hot springs / fissures. However this is a pretty good base for what we want. The only thing that isn't obtainable right now is Sapphires.

I know the python stuff is a little garbage but I was not sure how to do it without blowing up the for loop. So feel free to do it your own way


alcatrazEscapee:
Superseded by #1724

PR #1671, opened 2021-01-22
Eight bugfixes

### LINKED ISSUE #556: Comparators cannot detect barrel item or fluid values.
That's it really, comparators don't light up when a barrel has either items or a fluid in it, this isn't necessary but it would be a nice addition.

### LINKED ISSUE #689: Way to move plants outside of the zone they are allowed to live in.
Many people like to collect plants from places they have gone to decorate with them.
The current system in the game doesn't allow this to work anywhere. This could be crafting them with some special soil to get a "resilient plant" or something. Pots could work as well but wouldn't be ideal.

An example of what people have been doing in the test server:

### LINKED ISSUE #1331: TFCT: Metal items should heat faster in a hot crucible with enough liquid
If an item is completely surrounded with hot metal, it will heat faster.

I think this would be believable and useful.

 - TFC Version: build 140

### LINKED ISSUE #1586: Ingots shift clicked onto an anvil cannot be placed as ingot piles.
**Describe the bug**
Make sure to include:
Nothing
Ingots shift clicked onto an anvil cannot be placed back down into ingot piles.

**To Reproduce**
Self explanatory

**Meta Info**
 - TFC Version: 1.12.2-1.7.4.162

### LINKED ISSUE #1618: Support beams placement check can go through solid blocks
Bit of an odd placement case:

X/Y cross section:
```
SB..TS
S....S
S....S
```
Where `S` is a support beam, `B` is a solid block. Placing a support beam at `T` will place a full support beam, despite the placement should be being blocked by the block at `B`.

It only occurs if the block `B` is directly adjacent to the other support column. Causing a block update near the placed supports will cause them to break.

### LINKED ISSUE #1629: Allow dying unfinished small vessels in barrels
Almost any dying can be done either in barrels or in a crafting table. However, small vessels can only be dyed by crafting, not in a barrel. For uniformity, it would make sense if they could be dyed in a barrel.

**Meta Info**
 - TFC Version: 1.7.8.166
 - This is an enhancement for TFC alone

### LINKED ISSUE #1637: Lead and steel lamps have identical textures
The texture files for lead lamps and steel lamps are byte-for-byte identical. This is probably unintentional.

**Meta Info**
 - TFC Version: 1.7.9.167

### LINKED ISSUE #1668: When using bellows you can keep a forge going indefinite
**Describe the bug**
Make sure to include:
when a forge is out of fuel bellows would not allow it to gain heat without re-lighting with more fuel.

If the forge does not cool down completely you can keep the forge going at least at lower temperatures sufficient to melt metals like bismuth and zinc or cook food. So long as the forge does not cool down completely this effect is indefinite it would seem. I have not looked into you your code but it feels like the bellows is updating the forge thinking it is still lit because there is some temperature to it and not checking if there is actually fuel for it to use to do so, sorry if that last bit was unwelcome.

3. If it was a crash, upload the full log file (`latest.log`)
no crash occurs

**To Reproduce**
1. light forge.
2. allow fuel to burn out and down so temperature is dropping.
3. use bellows at any point bellow what i assume on the forge;s temperature gauge to be either bright red or the least dark red.
4. so long as bellows are used temperature will remain at this level.
5. if the forge is allowed to cool fully then effect does not occur.

**Meta Info**
 - TFC Version: 0.1.7.9.167
 journey map, waila and waila havestability, not enough items,  on version bug was discovered. bug does occur without these mods present.
  none

eerussianguy:
Fixes #1637 Lead and steel lamps are identical
Fixes #1629 Small vessels can be dyed in barrels
Fixes #1668 Nerfed bellows to not be an infinite heat source
Fixes #1331 Hot crucibles will heat items much faster than cold ones
Fixes #689  Added working flower pots
Fixes #1586 Added tooltip for partially forged items
Fixes #1618 Support beam placement checks in correct order
Fixes #556 barrels now have comparator signal when sealed

Bonus: Barrel GUI will catch names that are too long and move them


src/main/java/net/dries007/tfc/api/types/Plant.java
```diff
@@ -276,6 +276,16 @@ public int getAgeForWorldgen(Random rand, float temp)
         return rand.nextInt(Math.max(1, Math.min(Math.round(2.5f + ((temp - minGrowthTemp) / minGrowthTemp)), 4)));
     }
 
+    public boolean canBePotted()
+    {
+        return (isCrossModel() && !(oreDictName.isPresent() && oreDictName.get().equals("reed"))) || plantType == PlantType.CACTUS || plantType == PlantType.CREEPING || plantType == PlantType.TALL_PLANT;
```
Rongmario: Would rather have `(isCrossModel() && !(oreDictName.isPresent() && oreDictName.get().equals("reed")))` after the simple equality checks due to short circuiting.

src/main/java/net/dries007/tfc/client/model/ModelFlowerPot.java
```diff
@@ -0,0 +1,121 @@
...
+    @Override
+    public boolean isBuiltInRenderer()
+    {
+        return false;
+    }
+
+    @Override
+    public TextureAtlasSprite getParticleTexture()
+    {
+        return Minecraft.getMinecraft().getTextureMapBlocks().getAtlasSprite("minecraft:blocks/hardened_clay");
```
Rongmario: Cache this

src/main/java/net/dries007/tfc/client/model/ModelFlowerPot.java
```diff
@@ -0,0 +1,121 @@
...
+import net.dries007.tfc.objects.blocks.plants.BlockPlantTFC;
+
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+
+@MethodsReturnNonnullByDefault
+public class ModelFlowerPot implements IBakedModel
+{
+    private static final IModel FLOWER_POT_BLANK = ModelLoaderRegistry.getModelOrMissing(new ResourceLocation(MOD_ID, "block/flower_pot_cross"));
+    private Block CACHED_BLOCK = null; // to avoid reloading models in certain cases
+    private List<BakedQuad> CACHED_QUADS = null;
```
alcatrazEscapee: should be `lowerCamelCase`

src/main/java/net/dries007/tfc/client/model/ModelFlowerPot.java
```diff
@@ -0,0 +1,121 @@
...
+        if (state instanceof IExtendedBlockState && state.getBlock() instanceof BlockFlowerPotTFC)
+        {
+            IExtendedBlockState extendedBlockState = (IExtendedBlockState) state;
+            IBlockState flower = extendedBlockState.getValue(BlockFlowerPotTFC.FLOWER);
+            Block block = flower.getBlock();
+            if (block instanceof BlockPlantTFC && ((BlockPlantTFC) block).getPlant().isCrossModel())
+            {
+                // if we think that it's just a cross model, let's do it the easy way.
+                BlockModelShapes shapes = Minecraft.getMinecraft().getBlockRendererDispatcher().getBlockModelShapes();
+                TextureAtlasSprite flowerAsset = shapes.getModelForState(flower).getQuads(flower, null, rand).get(0).getSprite();
```
Rongmario: Would `Minecraft.getMinecraft().getBlockRendererDispatcher().getBlockModelShapes().getTexture()` be better?

src/main/java/net/dries007/tfc/objects/te/TEFlowerPotTFC.java
```diff
@@ -0,0 +1,65 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.nbt.NBTUtil;
+
+import net.dries007.tfc.util.Helpers;
+
+@ParametersAreNonnullByDefault
+public class TEFlowerPotTFC extends TEBase
+{
+    public IBlockState state;
```
alcatrazEscapee: Should be private with accessor + mutator if nessecary.

src/main/java/net/dries007/tfc/client/model/ModelFlowerPot.java
```diff
@@ -0,0 +1,121 @@
...
+                Map<String, String> sprites = new HashMap<>();
+                sprites.put("plant", flowerAsset.getIconName());
+                return bake(FLOWER_POT_BLANK.retexture(ImmutableMap.copyOf(sprites))).getQuads(state, side, rand);
+            }
+            else if (CACHED_BLOCK == block && CACHED_QUADS != null) // custom potted stuff should not be state sensitive since they don't inherit block state json
+            {
+                return CACHED_QUADS;
+            }
```
Rongmario: Instead of making a new `HashMap` and doing `ImmutableMap.copyOf`, just use `ImmutableMap.of()`.

This type of caching is fine for a simple baked model like this I suppose.

src/main/java/net/dries007/tfc/objects/blocks/BlockFlowerPotTFC.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+
+    @Override
+    public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World world, BlockPos pos, EntityPlayer player)
+    {
+        TEFlowerPotTFC te = Helpers.getTE(world, pos, TEFlowerPotTFC.class);
+        if (te != null)
+        {
+            return new ItemStack(Item.getItemFromBlock(te.state.getBlock()));
```
Rongmario: `ItemStack` allows `Block` passed through.

src/main/java/net/dries007/tfc/objects/blocks/BlockFlowerPotTFC.java
```diff
@@ -0,0 +1,195 @@
...
+    {
+        return BlockRenderLayer.CUTOUT;
+    }
+
+    @Override
+    protected BlockStateContainer createBlockState()
+    {
+        IProperty[] listedProperties = new IProperty[] {};
+        IUnlistedProperty[] unlistedProperties = new IUnlistedProperty[] {FLOWER};
+        return new ExtendedBlockState(this, listedProperties, unlistedProperties);
```
Rongmario: There's a builder pattern for this: `new BlockStateContainer.Builder(this).add(FLOWER).build()`

alcatrazEscapee (changes requested):
Minor issues. One major point: the flower pot implementation is a mess (maybe a justifiable mess, I'm not entirely sure yet)

- We should still (if at all possible) be using the vanilla flower pot item. Having our own unique one is just messy.
- The entire baked model / extended property stuff seems unnecessary to me, is there a reason it can't just be normal models per block? That is a much more future proof design.
- Placement of TFC stuff into a vanilla flower pot should convert upon placement (we can intercept that via event). When dropped, it should drop vanilla flower pot.

src/main/java/net/dries007/tfc/objects/blocks/wood/BlockSupport.java
```diff
@@ -477,16 +477,16 @@ private int getHorizontalDistance(EnumFacing face, IBlockAccess worldIn, BlockPo
         int distance = -1;
         for (int i = 0; i < 5; i++)
         {
+            if (!(worldIn.getBlockState(pos.offset(face, i)).getBlock() instanceof BlockSupport) && !worldIn.isAirBlock(pos.offset(face, i)))
```
Rongmario: Save `pos.offset(face, i)` as a local variable and use it in the statements

src/main/java/net/dries007/tfc/objects/te/TEFlowerPotTFC.java
```diff
@@ -0,0 +1,65 @@
...
+    {
+        state = stateIn;
+        markForSync();
+    }
+
+    public void dump()
+    {
+        if (hasWorld() && !world.isRemote)
+        {
+            Helpers.spawnItemStack(world, pos, new ItemStack(Item.getItemFromBlock(state.getBlock())));
```
Rongmario: As mentioned earlier, `ItemStack` can have `Block` passed through.

Rongmario (changes requested):

Rongmario (approved):

PR #1673, opened 2021-01-26
Made EntityFallingBlockTFC more optimized and friendlier

### LINKED ISSUE #973: Releasing a gravity-affected block with a sticky piston causes unexpected behavior / crashes
Version 125.

There's lots of different ways to reproduce this, one of which I showed here:
[streamable link](https://streamable.com/uo3zym)

Essentially releasing falling blocks with a piston causes the game to freak out and freeze. It doesn't produce a crash report because Windows ends up asking you to close the game manually because it froze.

Simplest form of the bug is when you have a dirt sitting on a block held by a sticky piston. Retract the piston to remove the block from underneath the dirt and the dirt will freeze in place. This doesn't cause any lag. Once you get to ~3 blocks doing the same thing, the game will break.

Rongmario:
- It is now a subclass of `EntityFallingBlock`, but still handles falling same (?) way as the previous implementation.
- Majority of checking the block below is moved to `fall`, which is a method that gets called when the entity just touched the block below.
- This saves a lot of `BlockPos` checking where it was done `onUpdate` (every tick).
- `IFallingBlock` checks `BlockFalling#canFallThrough` too, as well.
- `getName` is overriden, so falling block entities have names...
- AT: `protected net.minecraft.entity.item.EntityFallingBlock field_175132_d # fallState`
- Lang entry: `entity.falling_block.name`
- Most importantly, this fixes #973!


Rongmario:
I'll re-open when the PR is ready.

PR #1678, opened 2021-02-04
Overhaul Falling Block System

### LINKED ISSUE #1302: Falling blocks sometimes fail to load from NBT.
TFC: Somethingorother(142)

https://pastebin.com/rayL2CHu

Rongmario:
## Centralization
- All the preparation, logic and registration are now located in the class: `net.dries007.tfc.api.util.FallingBlockManager`
- CraftTweaker support for it is in: `net.dries007.tfc.compat.crafttweaker.CTFallingBlockManager`
- Firing of logic is now done on `NeighborNotifyEvent`, `EntityPlaceEvent`, `ExplosionEvent.Detonate`, `BreakEvent`. Whereas before it used to be fired in the Block's instance methods: `neighborChanged`, `onBlockAdded`, `onExplosionDestroy`, `onPlayerDestroy` respectively.

## Logic
- There are no big logic changes with regards to the falling blocks system. Nothing that the player-end should know anyway.
- #973 is fixed: this is originally addressed in PR #1673 but that has now since been abandoned for this PR. The fix was to match position, motion modification logic with vanilla's `EntityFallingBlock`.
- TileEntities and its `NBTTagCompound` are now kept properly, even fixing vanilla's logic in that regard (vanilla never utilizes it!)
- `EntityFallingBlockTFC` now holds a common/server-side only field: `currentSpecification`, more on that later. This need not to be synced to client-side.
- A client-side only field has also been added, `renderState`, this is for the `getBlock` getter; it now points to that on the client-side, to give `RenderFallingBlock` a good idea of what the current look of the block should be.
- These changes are fine with backwards compatibility. Existing `EntityFallingBlockTFC` that has been saved should carry on falling properly. Given they are still registered as a fallable state, otherwise they'd be `Entity#setDead`
- Fixes #1302 

## Optimization
- More significant use of short-circuiting in if statements that require a lot of different calls to checks.
- Matched our `EntityFallingBlockTFC` more or less to vanilla's `EntityFallingBlock` with our own logic concerning TFC falling logic and cave-ins/collapsing of blocks. This means:

1. Removal of `DataParameter<Optional<IBlockState>>` (now synced on-demand with the help of `IEntityAdditionalSpawnData`). 
2. Removal of `failedBreakCheck` which was redundant.
3. Removal of some motion updates as they triggered #973

- Streamlining of some checks within `onUpdate`

- `getFallablePos` hugely (citation needed) optimized in terms of checks, from 2 `Arrays.stream()` and allocations, it has been reduced to a smaller 4 elements `ArrayList` allocation, a boolean `hasFoundSideSupport` local variable, and a simple for loop that does what the 2 previous `Arrays::stream` would have done.
- `checkFalling` now uses `MutableBlockPos` when `BlockFalling::fallInstantly` is true (@ worldgen), and this relieves BlockPos creations and allocations. 
- Less calls of `world#getBlockState` which can be hefty during worldgen.

## Usages
- `FallingBlockManager` contains register and other modification helper methods, some more can be included to remove existing materials, but those cannot be exposed to CraftTweaker as they do not have appropriate `Material` exposure
- The class also contains various check helpers, they're great for API usage and I may javadoc them soon:tm:
- Liberal usages of functional interfaces, they're easily usable in CraftTweaker and is a blessing in Java 8 for various functional things I implemented for falling blocks.
- `Specification` is a data holder for a spec sheet (if you will) of falling blocks. They can be used and should be encouraged to be re-used when possible. If a similar spec is needed, there's a copy constructor which would do nicely. Say you want to a different `resultingState`, copy, and `setResultingState`. In vanilla TFC, I utilized this for raw and smooth rocks as they have their own respective cobble resulting states.


src/main/java/net/dries007/tfc/objects/blocks/BlocksTFC.java
```diff
@@ -402,6 +403,65 @@ else if (x.getType() != Rock.Type.SPIKE && x.getType() != Rock.Type.ANVIL)
...
+                    Rock.Type nonGrassType = type.getNonGrassVersion();
+                    if (nonGrassType != type)
+                    {
+                        BlockRockVariant.get(rock, type).getBlockState().getValidStates().forEach(s -> {
+                            FallingBlockManager.getSpecification(s).setResultingState(BlockRockVariant.get(rock, nonGrassType).getDefaultState())
+                        });
+                    }
+                }
+
+                 */
```
alcatrazEscapee: Remove

src/main/java/net/dries007/tfc/objects/blocks/stone/BlockRockVariantConnected.java
```diff
@@ -160,7 +158,8 @@ public boolean checkFalling(World worldIn, BlockPos pos, IBlockState state)
                 {
                     worldIn.setBlockState(pos1, BlockRockVariant.get(rock, type.getNonGrassVersion()).getDefaultState());
                 }
-                worldIn.spawnEntity(new EntityFallingBlockTFC(worldIn, pos1, this, worldIn.getBlockState(pos1)));
+                // worldIn.spawnEntity(new EntityFallingBlockTFC(worldIn, pos1, this, worldIn.getBlockState(pos1)));
```
alcatrazEscapee: Remove

src/main/java/net/dries007/tfc/api/types/Rock.java
```diff
@@ -125,44 +128,58 @@ public Item create(RockCategory category)
...
-        private final FallingBlockType gravType;
+        @Deprecated private final FallingBlockType gravType;
+        @Nullable private final FallingBlockManager.Specification fallingSpecification;
 
-        Type(Material material, FallingBlockType gravType, boolean isGrass)
+        @Deprecated
+        Type(Material material, @Deprecated FallingBlockType gravType, boolean isGrass)
+        {
+            this(material, gravType, isGrass, null);
+        }
```
alcatrazEscapee: There's no real reason to deprecate a private enum constructor, this can just be removed.

src/main/java/net/dries007/tfc/api/types/Rock.java
```diff
@@ -125,44 +128,58 @@ public Item create(RockCategory category)
...
             this.isGrass = isGrass;
+            this.fallingSpecification = fallingSpecification;
         }
 
         public boolean canFall()
         {
             return gravType != NO_FALL;
         }
 
+        @Deprecated
```
alcatrazEscapee: If you're going to deprecate something, add a comment pointing to what should be used instead.

src/main/java/net/dries007/tfc/api/util/FallingBlockManager.java
```diff
@@ -0,0 +1,530 @@
...
+        public void endFall(World world, BlockPos pos)
+        {
+            if (endFallCallback != null)
+            {
+                endFallCallback.endFall(world, pos);
+            }
+        }
+
+        @FunctionalInterface
+        public interface FallDropsProvider
```
alcatrazEscapee: Interfaces should have `I` prefix as per forge/minecraft convention.

src/main/java/net/dries007/tfc/compat/crafttweaker/CTFallingBlockManager.java
```diff
@@ -0,0 +1,283 @@
+package net.dries007.tfc.compat.crafttweaker;
```
alcatrazEscapee, line 6: Lovely, but afaik there's no documentation anywhere for this. A PR to the docs repo would be appreciated for all this new stuff. Also helps explain what this is all doing.

src/main/java/net/dries007/tfc/objects/entity/EntityFallingBlockTFC.java
```diff
@@ -1,108 +1,119 @@
-/*
- * Work under Copyright. Licensed under the EUPL.
- * See the project README.md and LICENSE.txt for more information.
- */
```
alcatrazEscapee: This shouldn't be removed, might need to check elsewhere for consistent license headers.

src/main/java/net/dries007/tfc/util/ICollapsableBlock.java
```diff
@@ -1,121 +0,0 @@
-/*
```
alcatrazEscapee, line 1: So this is a better reference for the earlier deprecation. Personally, in the 1.12 code base, I would prefer we just don't bother deprecating things at all. There's such an extent of things that *could feasibly be* used by other mods (for instance... this interface), and we don't particularly have an official public API, asides from a package with things that addons/other mods may want to use.

To add to that, we're not planning to transition away from deprecated functionality to using new stuff. 1.16 is a completely rewrite of the falling block mechanics (and much of what's here has already been done.) So I think this is a bit pointless.

alcatrazEscapee (changes requested):
Haven't tested - assuming you have and can verify that existing falling block mechanics work as previous. Some cleanup required. Docs for the added CT stuff would be nice.

src/main/java/net/dries007/tfc/compat/crafttweaker/CTFallingBlockManager.java
Rongmario: Writing up as we speak

src/main/java/net/dries007/tfc/objects/entity/EntityFallingBlockTFC.java
Rongmario: I think for some reason my IDE deleted that; I might have to open a separate PR to amend existing classes with no license headers also.

src/main/java/net/dries007/tfc/util/ICollapsableBlock.java
Rongmario: Sure. The API portion is `FallingBlockManager` now, blocks can do what they could do before and slightly more there instead of implementing via interfaces.

src/main/java/net/dries007/tfc/api/types/Rock.java
Rongmario: Decided to just strip it.

alcatrazEscapee (commented):
All I'm waiting for this is a docs PR with the added CT stuff.

PR #1679, opened 2021-02-05
Optimized pre-definition of ItemSizeHandlers

Rongmario:
- A normal TFC dev env normally gets you

- `11,612 | 272.16 KB | net.dries007.tfc.api.capability.size.ItemSizeHandler`
- Which is 11k~ instances of `ItemSizeHandler` at about 272K RAM usage

After this commit:

- `8 | 192 bytes | net.dries007.tfc.api.capability.size.ItemSizeHandler`
- Which is 8 instances of `ItemSizeHandler` at around 192B

- This was only doable because the object holds 3 defined fields, 2 enum values and 1 boolean. The backing cache `EnumMap<Size, EnumMap<Weight, ItemSizeHandler[]>>` also is pretty efficient thanks to Java. This will obviously not work for custom `IItemSize` implementations, which is fine.

### DIFF
--- a/src/main/java/net/dries007/tfc/api/capability/size/CapabilityItemSize.java
+++ b/src/main/java/net/dries007/tfc/api/capability/size/CapabilityItemSize.java
@@ -36,20 +36,20 @@ public final class CapabilityItemSize
     public static void preInit()
     {
         // Register the capability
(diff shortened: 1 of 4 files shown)


alcatrazEscapee:
654ea3b33f1b3975d8296af1071206586bd7067a

PR #1703, opened 2021-02-24
Count all logs of same type when looking at log pile

sguest:
Currently log piles only show the contents of the first slot in the HWYLA/TOP tooltip, meaning they will show a maximum of 4. This will allow them to show a full count.

For log piles with multiple types of wood, only a count of logs matching the first slot is shown, any other logs in the other slots are not reflected in the tooltip. I'm not sure how else to handle this since `IWailaBlock` only supports a single ItemStack for the icon.

### DIFF
--- a/src/main/java/net/dries007/tfc/compat/waila/HwylaPluginTFC.java
+++ b/src/main/java/net/dries007/tfc/compat/waila/HwylaPluginTFC.java
@@ -34,6 +34,7 @@ public class HwylaPluginTFC implements IWailaPlugin
         new HwylaBlockInterface(new InfoProvider()),
         new HwylaBlockInterface(new TreeProvider()),
         new HwylaBlockInterface(new IngotPileProvider()),
+        new HwylaBlockInterface(new LogPileProvider()),
         new HwylaBlockInterface(new QuernProvider())
     );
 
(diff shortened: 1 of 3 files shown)


Rongmario:
HWYLA and TOP both have their own inventory detection mechanism and custom tooltip, it is not for us to handle it ourselves.

sguest:
> 
> 
> HWYLA and TOP both have their own inventory detection mechanism and custom tooltip, it is not for us to handle it ourselves.

But the default mechanism does a poor job handling log piles since it only uses the first slot. Doesn't a custom override make sense in such a case to better show all the logs in the pile?

Before changes:
TOP-before

After changes:
TOP-after

Rongmario:
That's not the default, try pressing shift and see what I mean.

sguest:
Interesting, I didn't know TOP did that (I've been mostly using HYWLA for my game, but TOP was a better choice for taking a screenshot).

However, it doesn't look like HWYLA will allow you to see the full count even if you're sneaking. Also for log piles specifically it still just feels wrong to only show 1 stack on the non-sneaking tooltip. So personally I still think this is a good change, but it's good to at least know the option is there with TOP without any changes.

Rongmario:
HWLYA's inventory tooltip exists in code but I think its either half implemented, or bugged.

eerussianguy:
I'm going to merge this, but next time please apply TFC code style. Today I will do it for you on merge.

sguest:
Apologies, I thought for something this small I could just format it by hand, next time I'll make sure I get the auto formatting working.

eerussianguy:
Yeah the big deal is just bracket placement. The rest is sort of ephemeral

PR #1716, opened 2021-03-03
Add units information to small vessels

Charnuz:
This should reduce the amount of math being done/accidentally making unknown metal when making alloys. Hopefully it will make them slightly easier to use.
image
Currently it only displays when everything in the vessel has metal units. Went for one decimal place for situations where rounding would round up to be a valid alloy, but the actual percentage would be slightly below what is required. There could be config settings for this but I'm not sure what they'd be.


eerussianguy:
This looks great. I might only enable it for advanced tooltips (f3+H) because some players might feel like they want to do the math themself as is the classic way.

Charnuz:
Thats a good idea, perhaps a config option to allow it to display without advanced tooltips on?

eerussianguy:
I'd be fine with that but here's what Alc/Clay would say: The config option is pressing the key. Weird to have a config for a config.

Charnuz:
Ah yeah, was thinking more of people who would want that information without having the rest of the advanced tooltip text. Looks like the tooltip info should be moved from the IItemHeat.addHeatInfo() method since that can't determine if advanced tooltips are on? Might need to be handled by ClientEvents.onItemTooltip()

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemSmallVessel.java
```diff
@@ -282,16 +288,41 @@ public void addHeatInfo(@Nonnull ItemStack stack, @Nonnull List<String> text)
             else
             {
                 boolean hasContent = false;
+                Map<Metal, Integer> materials = new HashMap<>();
```
alcatrazEscapee, line 291: This could be a fastutil `Object2IntMap` / `Object2IntOpenHashMap`

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemSmallVessel.java
```diff
@@ -282,16 +288,41 @@ public void addHeatInfo(@Nonnull ItemStack stack, @Nonnull List<String> text)
...
                         hasContent = true;
                     }
                 }
 
-                if (!hasContent)
+                if (hasContent)
+                {
+                    if (onlySmeltables)
+                    {
+                        int textPosition = (int) super.stacks.stream().filter(itemstack -> !ItemStack.EMPTY.equals(itemstack)).count() + 1;
```
alcatrazEscapee, line 315: What is `super.stacks` doing here this looks very confusing to me? (This is not a criticism I actually don't understand what's going on here)

alcatrazEscapee (approved):
Minor inconsequential performance nitpick and a question, otherwise looks fine.

eerussianguy:
I will take care of these considerations locally. Thanks for your contribution!

src/main/java/net/dries007/tfc/objects/items/ceramics/ItemSmallVessel.java
Charnuz: That gets the count of non-empty item stacks so we can place the tooltip right after the contents text, since that can be between 0-4 and super.stacks.size() always returns 4.

Now that I think about it, it probably should be `super.stacks.stream().filter(slot->!slot.isEmpty()).count()` instead.

PR #1727, opened 2021-03-07
Bulk PR but cleaned up (?)

eerussianguy:
Created a new branch cause resetting was giving me issues (my cllient is a little sucky)


src/main/java/net/dries007/tfc/ForgeEventHandler.java
```diff
@@ -397,4 +405,97 @@ else if (originalBlock == Blocks.BASALT)
             event.setNewState(TFCBlocks.ROCK_BLOCKS.get(Rock.Default.BASALT).get(Rock.BlockType.HARDENED).get().defaultBlockState());
         }
     }
+
+    @SubscribeEvent
+    public static void onEntityItemPickup(EntityItemPickupEvent event)
+    {
+        ItemStack stack = event.getItem().getItem();
+        if (stack.getItem() == TFCItems.TORCH.get())
```
alcatrazEscapee: afaik there's no reason we can't write torch handling to handle both vanilla torches and ours, so this could be a tag check. The key (`"burnTicks"`) should also reference a static constant somewhere, preferably with a doc comment as the discoverability of this is pretty low.

src/main/java/net/dries007/tfc/StartFireEvent.java
```diff
@@ -0,0 +1,82 @@
+package net.dries007.tfc;
```
alcatrazEscapee: no, nothing extraneous in the root package, stick this in `util.events`

src/main/java/net/dries007/tfc/TFCEventFactory.java
```diff
@@ -0,0 +1,31 @@
+package net.dries007.tfc;
```
alcatrazEscapee: This class is unnecessary. Forge has one because it's patched into vanilla code.

src/main/java/net/dries007/tfc/TFCEventFactory.java
```diff
@@ -0,0 +1,31 @@
...
+import net.minecraft.world.World;
+import net.minecraftforge.common.MinecraftForge;
+
+public class TFCEventFactory
+{
+    /**
+     * This starts a fire on the passed in face if the event is not intercepted. See {@link StartFireEvent} for usage considerations.
+     * @return Whether a fire block was placed or not
+     */
+    public static boolean startFire(World world, BlockPos pos, BlockState state, Direction direction, @Nullable PlayerEntity player, @Nullable ItemStack stack)
```
alcatrazEscapee: Static method in `StartFireEvent`, `StartFireEvent` can have private constructor.

src/main/java/net/dries007/tfc/client/TFCKeybindings.java
```diff
@@ -0,0 +1,32 @@
...
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+import static net.dries007.tfc.TerraFirmaCraft.MOD_NAME;
+
+@Mod.EventBusSubscriber(modid = MOD_ID, bus = Mod.EventBusSubscriber.Bus.FORGE, value = Dist.CLIENT)
+public class TFCKeybindings
+{
+    public static final KeyBinding PLACE_BLOCK = new KeyBinding("tfc.key.place_block", KeyConflictContext.IN_GAME, InputMappings.Type.KEYSYM, GLFW.GLFW_KEY_V, MOD_NAME);
+
+    @SubscribeEvent
+    public static void onKeyEvent(InputEvent.KeyInputEvent event)
```
alcatrazEscapee: Should just be a method in `ClientForgeEventHandler` No need for the class to be marked as an event bus subscriber

src/main/java/net/dries007/tfc/client/TFCKeybindings.java
```diff
@@ -0,0 +1,32 @@
...
+import net.minecraftforge.fml.network.PacketDistributor;
+
+import net.dries007.tfc.network.PacketHandler;
+import net.dries007.tfc.network.PlaceBlockSpecialPacket;
+
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+import static net.dries007.tfc.TerraFirmaCraft.MOD_NAME;
+
+@Mod.EventBusSubscriber(modid = MOD_ID, bus = Mod.EventBusSubscriber.Bus.FORGE, value = Dist.CLIENT)
+public class TFCKeybindings
```
alcatrazEscapee:
```suggestion
public class TFCKeyBindings
```

src/main/java/net/dries007/tfc/client/TFCSoundTypes.java
```diff
@@ -0,0 +1,8 @@
+package net.dries007.tfc.client;
+
+import net.minecraftforge.common.util.ForgeSoundType;
+
+public class TFCSoundTypes
```
alcatrazEscapee: Can just be in `TFCSounds`, no reason for them to be different as we won't have too many

src/main/java/net/dries007/tfc/client/particle/BubbleParticle.java
```diff
@@ -0,0 +1,66 @@
...
+            zd *= 0.85F;
+        }
+    }
+
+    public IParticleRenderType getRenderType()
+    {
+        return IParticleRenderType.PARTICLE_SHEET_OPAQUE;
+    }
+
+    @OnlyIn(Dist.CLIENT)
```
alcatrazEscapee: No `OnlyIn`

src/main/java/net/dries007/tfc/client/particle/TFCParticleTypes.java
```diff
@@ -0,0 +1,16 @@
...
+
+import net.minecraft.particles.BasicParticleType;
+import net.minecraft.particles.ParticleType;
+import net.minecraftforge.fml.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+
+public class TFCParticleTypes
```
alcatrazEscapee:
```suggestion
public class TFCParticles
```

src/main/java/net/dries007/tfc/client/render/GrillTileEntityRenderer.java
```diff
@@ -0,0 +1,66 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.vector.Vector3f;
+import net.minecraftforge.items.CapabilityItemHandler;
+
+import com.mojang.blaze3d.matrix.MatrixStack;
+import net.dries007.tfc.common.tileentity.GrillTileEntity;
+
+import static net.dries007.tfc.common.tileentity.PotTileEntity.SLOT_EXTRA_INPUT_END;
+import static net.dries007.tfc.common.tileentity.PotTileEntity.SLOT_EXTRA_INPUT_START;
```
alcatrazEscapee: It's quite odd to import these from the pot in the grill TER

src/main/java/net/dries007/tfc/client/render/PitKilnTileEntityRenderer.java
```diff
@@ -0,0 +1,65 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.vector.Vector3f;
+
+import net.minecraftforge.items.CapabilityItemHandler;
+
+import com.mojang.blaze3d.matrix.MatrixStack;
+import net.dries007.tfc.common.blocks.PitKilnBlock;
+import net.dries007.tfc.common.tileentity.PitKilnTileEntity;
+
+@ParametersAreNonnullByDefault
```
alcatrazEscapee:
```suggestion
```
No need, should be taken care of by the package level annotations

src/main/java/net/dries007/tfc/client/render/GrillTileEntityRenderer.java
```diff
@@ -0,0 +1,66 @@
...
+
+@ParametersAreNonnullByDefault
```
alcatrazEscapee:
```suggestion
```
No need, should be taken care of by the package level annotations

src/main/java/net/dries007/tfc/client/render/PlacedItemTileEntityRenderer.java
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.client.renderer.tileentity.TileEntityRendererDispatcher;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.vector.Vector3f;
+
+import net.minecraftforge.items.CapabilityItemHandler;
+
+import com.mojang.blaze3d.matrix.MatrixStack;
+import net.dries007.tfc.common.tileentity.PlacedItemTileEntity;
+
+@ParametersAreNonnullByDefault
```
alcatrazEscapee:
```suggestion
```
No need, should be taken care of by the package level annotations.

Going to stop marking these for now, that's why the package level annotations exist, IDE warnings means you're missing them.

src/main/java/net/dries007/tfc/client/screen/FirepitScreen.java
```diff
@@ -0,0 +1,39 @@
...
+import net.dries007.tfc.common.capabilities.heat.Heat;
+import net.dries007.tfc.common.container.FirepitContainer;
+import net.dries007.tfc.common.tileentity.FirepitTileEntity;
+
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+
+public class FirepitScreen extends TFCContainerScreen<FirepitContainer>
+{
+    private static final ResourceLocation FIREPIT = new ResourceLocation(MOD_ID, "textures/gui/fire_pit.png");
+    private final FirepitTileEntity tile;
```
alcatrazEscapee: Should be in a superclass screen `TileEntityScreen<T extends InventoryTileEntity, C extends TileEntityContainer<T>> extends TFCContainerScreen<C>`

src/main/java/net/dries007/tfc/client/screen/FirepitScreen.java
```diff
@@ -0,0 +1,39 @@
...
+        super.renderBg(matrixStack, partialTicks, mouseX, mouseY);
+        int temp = (int) (51 * (int) (tile.getFields()[FirepitTileEntity.FIELD_TEMP]) / Heat.maxVisibleTemperature());
+        if (temp > 0)
+            blit(matrixStack, leftPos + 30, topPos + 66 - Math.min(51, temp), 176, 0, 15, 5);
+    }
+
+    @Override
+    protected void renderLabels(MatrixStack matrixStack, int x, int y)
+    {
+        // disable inventory labels cause they don't fit
```
alcatrazEscapee: We should really just fix this tbh. Across the board, make the gui taller if need be.

src/main/java/net/dries007/tfc/client/screen/FirepitScreen.java
```diff
@@ -0,0 +1,39 @@
...
+    {
+        super(container, playerInventory, name, FIREPIT);
+        tile = container.getTileEntity();
+    }
+
+    @Override
+    protected void renderBg(MatrixStack matrixStack, float partialTicks, int mouseX, int mouseY)
+    {
+        super.renderBg(matrixStack, partialTicks, mouseX, mouseY);
+        int temp = (int) (51 * (int) (tile.getFields()[FirepitTileEntity.FIELD_TEMP]) / Heat.maxVisibleTemperature());
```
alcatrazEscapee: This is unnecessary, it's what `getSyncableData` is for. `tile.getSyncableData().get(id)`

src/main/java/net/dries007/tfc/client/screen/GrillScreen.java
```diff
@@ -0,0 +1,40 @@
...
+import net.dries007.tfc.common.container.GrillContainer;
+import net.dries007.tfc.common.tileentity.FirepitTileEntity;
+import net.dries007.tfc.common.tileentity.GrillTileEntity;
+
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+
+public class GrillScreen extends TFCContainerScreen<GrillContainer>
+{
+    private static final ResourceLocation BACKGROUND = new ResourceLocation(MOD_ID, "textures/gui/fire_pit_grill.png");
+    private final GrillTileEntity tile;
```
alcatrazEscapee: Same, TE reference should be in a superclass screen.

src/main/java/net/dries007/tfc/client/screen/PotScreen.java
```diff
@@ -0,0 +1,48 @@
+package net.dries007.tfc.client.screen;
```
alcatrazEscapee, line 7: Same notes here as other screens

src/main/java/net/dries007/tfc/client/screen/TFCContainerScreen.java
```diff
@@ -14,10 +14,14 @@
...
 import com.mojang.blaze3d.systems.RenderSystem;
+import net.dries007.tfc.common.tileentity.TFCTileEntity;
 
-public abstract class TFCContainerScreen<C extends Container> extends ContainerScreen<C>
+import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+
+public class TFCContainerScreen<C extends Container> extends ContainerScreen<C>
 {
     protected final ResourceLocation texture;
+    protected static final ResourceLocation SMALL_INV = new ResourceLocation(MOD_ID, "textures/gui/small_inventory.png");
```
alcatrazEscapee: As a general rule for ordering things:

- static before instance
- fields, ctors, methods.
- most access (public) -> least access (private)
- final before non final

IDEA can do this for you with the "rearrange code" refactor.

src/main/java/net/dries007/tfc/common/blocks/BurningLogPileBlock.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    public static void tryLightLogPile(World world, BlockPos pos)
+    {
+        LogPileTileEntity pile = Helpers.getTileEntity(world, pos, LogPileTileEntity.class);
+        if (pile != null)
+        {
+            int logs = pile.countLogs();
+            pile.clearContent(); // avoid dumping when onRemove is called
+            pile.setRemoved();
```
alcatrazEscapee: Should not be calling this manually.

src/main/java/net/dries007/tfc/common/blocks/BurningLogPileBlock.java
```diff
@@ -0,0 +1,121 @@
...
+                newPile.light(logs);
+                tryLightNearby(world, pos);
+            }
+        }
+    }
+
+    private static void tryLightNearby(World world, BlockPos pos)
+    {
+        if (world.isClientSide()) return;
+        for (Direction side : Direction.values())
```
alcatrazEscapee: See: note in discord about `Direction.values()`

src/main/java/net/dries007/tfc/common/blocks/LogPileBlock.java
```diff
@@ -0,0 +1,127 @@
...
+        InventoryTileEntity te = Helpers.getTileEntity(world, pos, InventoryTileEntity.class);
+        if (te != null)
+            te.onBreak();
+        super.onRemove(state, world, pos, newState, isMoving);
+    }
+
+    @Override
+    protected void createBlockStateDefinition(StateContainer.Builder<Block, BlockState> builder)
+    {
+        builder.add(AXIS);
```
alcatrazEscapee: Doesn't matter in this case, but for future reference, this should always be implemented as a super class call:
```suggestion
        super.createBlockStateDefinition(builder.add(AXIS));
```

that way subclasses can pick up superclass properties without having to explicitly declare them

src/main/java/net/dries007/tfc/common/blocks/PitKilnBlock.java
```diff
@@ -0,0 +1,207 @@
...
+    public static final int LOG_START = 8;
+    public static final int LIT = 16;
+
+    protected static final VoxelShape[] SHAPE_BY_LAYER = new VoxelShape[] {
+        ONE_SIXTEENTH, ONE_EIGHTH, THREE_SIXTEENTHS, ONE_FOURTH,
+        FIVE_SIXTEENTHS, THREE_EIGHTHS, SEVEN_SIXTEENTHS, ONE_HALF, // end straw
+        THREE_FOURTHS, THREE_FOURTHS, THREE_FOURTHS, THREE_FOURTHS,
+        VoxelShapes.block(), VoxelShapes.block(), VoxelShapes.block(), VoxelShapes.block(), // end logs
+        VoxelShapes.block() // LIT
+    };
```
alcatrazEscapee:
```suggestion
    public static final VoxelShape[] SHAPES = Util.make(new VoxelShape[16], shapes -> {
        for (int i = 0; i < 8; i++)
        {
            shapes[i] = Block.box(0, 0, 0, 16, 1 + i, 16);
        }
        final VoxelShape half = Block.box(0, 0, 0, 16, 8, 16);
        final VoxelShape threeQuarters = Block.box(0, 0, 0, 16, 12, 16);
        for (int i = 0; i < 4; i++)
        {
            shapes[8 + i] = VoxelShapes.or(half, Block.box(0, 8, 0, 16, 12, 4));
            shapes[12 + i] = VoxelShapes.or(threeQuarters, Block.box(0, 12, 0, 16, 16, 4));
        }
    });
```

Double check the math + orientations of the "log" shapes but something like that should be cleaner and more accurate.

alcatrazEscapee (commented):
First Pass

src/main/java/net/dries007/tfc/ForgeEventHandler.java
eerussianguy: This is actually dead code. I was not able to get the tags to persist properly so I went with a slightly stranger solution that I think is workable. Might be something to revisit later

src/main/java/net/dries007/tfc/client/screen/FirepitScreen.java
```diff
@@ -0,0 +1,33 @@
...
+        super(container, playerInventory, name, FIREPIT);
+        inventoryLabelY += 20;
+        imageHeight += 20;
+    }
+
+    @Override
+    protected void renderBg(MatrixStack matrixStack, float partialTicks, int mouseX, int mouseY)
+    {
+        super.renderBg(matrixStack, partialTicks, mouseX, mouseY);
+        int temp = (int) (51 * ((FirepitTileEntity) tile).getSyncableData().get(0) / Heat.maxVisibleTemperature());
```
alcatrazEscapee: The constant `0` should be referencing a field ID from `FirepitTileEntity` Otherwise it's impossible to know what this is syncing, or the uses if it needs changing. Same everywhere else `getSyncableData()` is used not en-masse

src/main/java/net/dries007/tfc/common/blocks/GroundcoverBlock.java
```diff
@@ -63,7 +63,8 @@ public static GroundcoverBlock looseOre(Properties properties)
     }
 
     private final VoxelShape shape;
-    @Nullable private final Supplier<? extends Item> pickBlock;
+    @Nullable
+    private final Supplier<? extends Item> pickBlock;
```
alcatrazEscapee, line 66: Note to self (and sort of you): reimport code style and see if IDEA still makes this change. I thought I fixed it last time to not force single line annotations, might not of worked though. imo I want to allow it and not force anything either way.

src/main/java/net/dries007/tfc/common/blocks/BurningLogPileBlock.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public void randomTick(BlockState state, ServerWorld world, BlockPos pos, Random rand)
+    {
+        tryLightNearby(world, pos);
+    }
+
+    @Override
+    public void animateTick(BlockState stateIn, World worldIn, BlockPos pos, Random rand)
```
alcatrazEscapee: Missing `@OnlyIn(Dist.CLIENT)` as it overrides a method with the same annotation. (Forge disagrees that we should include this here as 'oh no it's a scary coremod'. There's documented cases of not including it breaking reflection, to which the response was 'so what'. But we're already a big scary coremod so I'd rather not leave the fragile thing there.)

IDEA has an option to configure annotations to copy from superclass methods when overriding, suggest adding `OnlyIn` as one of them.

It's also a good reminder that this is a clientside method

src/main/java/net/dries007/tfc/common/blocks/CharcoalPileBlock.java
```diff
@@ -0,0 +1,89 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.RayTraceResult;
+import net.minecraft.world.IBlockReader;
+import net.minecraft.world.IWorld;
+import net.minecraft.world.IWorldReader;
+import net.minecraft.world.World;
+
+import net.dries007.tfc.client.TFCSounds;
+
+public class CharcoalPileBlock extends SnowBlock
```
alcatrazEscapee: Don't override snow block, this is asking for missed override methods (and in fact, it is, and should not inherit `canBeReplaced` or `randomTick`. Also we mixin `SnowBlock`, none of which we want as part of charcoal. Neither should `instanceof SnowBlock` be true for charcoal. Just duplicate the shape + property spec.

src/main/java/net/dries007/tfc/common/blocks/CharcoalPileBlock.java
```diff
@@ -0,0 +1,89 @@
...
+    {
+        return new ItemStack(Items.CHARCOAL);
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public SoundType getSoundType(BlockState state)
+    {
+        return TFCSounds.CHARCOAL;
+    }
```
alcatrazEscapee: Can this not just be passed in via the block properties?

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -9,15 +9,18 @@
 import java.util.Map;
```
alcatrazEscapee, line 9: fwiw, this class (along with `TFCItems` is organized by item group. Please verify that it still is.

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -26,6 +29,11 @@
 import net.minecraftforge.registries.ForgeRegistries;
 
 import net.dries007.tfc.common.TFCItemGroup;
+import net.dries007.tfc.common.blocks.berry_bush.*;
+import net.dries007.tfc.common.blocks.devices.FirepitBlock;
+import net.dries007.tfc.common.blocks.devices.GrillBlock;
+import net.dries007.tfc.common.blocks.devices.PotBlock;
+import net.dries007.tfc.common.blocks.fruit_tree.*;
```
alcatrazEscapee: No `_` in package names, fix wherever it was used.

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -151,23 +169,81 @@
...
+    );
+    public static final Map<BerryBush.Default, RegistryObject<Block>> STATIONARY_BUSHES = Helpers.mapOfKeys(BerryBush.Default.class, BerryBush.Default::isStationary, berry ->
+        register("berry_bush/" + berry.name().toLowerCase() + "_bush", () -> new StationaryBerryBushBlock(new ForgeBlockProperties(Properties.of(Material.LEAVES).strength(0.6f).noOcclusion().randomTicks().sound(SoundType.SWEET_BERRY_BUSH)).tileEntity(BerryBushTileEntity::new).flammable(60, 30), berry.getBush()), FLORA)
+    );
 
+    public static final Map<BerryBush.Default, RegistryObject<Block>> WATERLOGGED_BUSHES = Helpers.mapOfKeys(BerryBush.Default.class, BerryBush.Default::isWaterlogged, berry ->
+        register("berry_bush/" + berry.name().toLowerCase() + "_bush", () -> new WaterloggedBerryBushBlock(new ForgeBlockProperties(Properties.of(Material.LEAVES).strength(0.6f).noOcclusion().randomTicks().sound(SoundType.SWEET_BERRY_BUSH)).tileEntity(BerryBushTileEntity::new).flammable(60, 30), berry.getBush()), FLORA)
+    );
```
alcatrazEscapee: Problems:
- these maps are sparse, meaning you need to know where the blocks are before iterating
- you've made it difficult to iterate over all berry bushes without seriously convoluted code

Fix:
- collect into one map, using either a `create`  + switch factory method or a `Type` enum to have different constructors (i.e. see coral)

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -151,23 +169,81 @@
...
         .lightLevel((state) -> TFCSeaPickleBlock.isDead(state) ? 0 : 3 + 3 * state.getValue(SeaPickleBlock.PICKLES)).sound(SoundType.SLIME_BLOCK).noOcclusion()), FLORA);
+    public static final RegistryObject<Block> BURNING_LOG_PILE = register("burning_log_pile", () -> new BurningLogPileBlock(new ForgeBlockProperties(AbstractBlock.Properties.of(Material.WOOD).randomTicks().strength(0.6F).sound(SoundType.WOOD)).flammable(60, 30).tileEntity(BurningLogPileTileEntity::new)));
+    public static final RegistryObject<Block> LOG_PILE = register("log_pile", () -> new LogPileBlock(new ForgeBlockProperties(AbstractBlock.Properties.of(Material.WOOD).strength(0.6F).sound(SoundType.WOOD)).flammable(60, 30).tileEntity(LogPileTileEntity::new)));
+    public static final Map<DyeColor, RegistryObject<Block>> RAW_ALABASTER = Helpers.mapOfKeys(DyeColor.class, color ->
+        register(("alabaster/stained/" + color.getName()).toLowerCase() + "_raw_alabaster", () -> new Block(AbstractBlock.Properties.of(Material.STONE).requiresCorrectToolForDrops().strength(1.0F, 6.0F)), DECORATIONS)
+    );
```
alcatrazEscapee: This ordering is both very confusing, messy, and also almost surely not item group order.

src/main/java/net/dries007/tfc/common/blocks/berry_bush/AbstractBerryBushBlock.java
```diff
@@ -0,0 +1,240 @@
...
+    @Override
+    protected void createBlockStateDefinition(StateContainer.Builder<Block, BlockState> builder)
+    {
+        builder.add(LIFECYCLE, STAGE);
+    }
+
+    public BerryBush getBush()
+    {
+        return bush;
+    }
```
alcatrazEscapee: This is a pointless type object?

src/main/java/net/dries007/tfc/common/blocks/berry_bush/BerryBush.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.function.Supplier;
+
+import net.minecraft.item.Item;
+import net.minecraft.item.Items;
+
+import net.dries007.tfc.common.items.TFCItems;
+
+import static net.dries007.tfc.common.blocks.berry_bush.AbstractBerryBushBlock.Lifecycle.*;
+
+public class BerryBush
```
alcatrazEscapee: This needs some redesign
- Noop bush is confusing
- There's no json here, so no reason for a `T` / `T.Default` pattern
- Having both `Berry` and `BerryBush` is redundant
- `BerryBush.Default` is just a passthrough for `BerryBush` which is not an enum for some reason
- Nowhere do blocks actually use `getBush` meaning the type object doesn't need to be held on in the block, and can just be used to instantiate with req'd properties - i.e. see soil block variants. They don't hold on to their type objects.

src/main/java/net/dries007/tfc/common/blocks/LogPileBlock.java
```diff
@@ -0,0 +1,129 @@
...
+    public void onRemove(BlockState state, World world, BlockPos pos, BlockState newState, boolean isMoving)
+    {
+        InventoryTileEntity te = Helpers.getTileEntity(world, pos, InventoryTileEntity.class);
+        if (state.hasTileEntity() && (!state.is(newState.getBlock()) || !newState.hasTileEntity()))
+        {
+            if (te != null)
+                te.onBreak();
+            world.removeBlockEntity(pos);
+        }
+    }
```
alcatrazEscapee: There has to be a better way of handling this than copy pasting the same vanilla method across every single TE, and not delegating to super at all.

src/main/java/net/dries007/tfc/common/blocks/PitKilnBlock.java
```diff
@@ -0,0 +1,197 @@
...
+    public static final int STRAW_END = 7;
+    public static final int LOG_START = 8;
+    public static final int LIT = 16;
+    public static final VoxelShape[] SHAPE_BY_LAYER = Util.make(new VoxelShape[17], shapes -> {
+        for (int i = 0; i < 8; i++)
+            shapes[i] = Block.box(0.0D, 0.0D, 0.0D, 16.0D, i + 1, 16.0D);
+        for (int i = 0; i < 4; i++)
+            shapes[8 + i] = VoxelShapes.or(shapes[7 + i], Block.box(4 * i, 8.0D, 0.0D, 4 * (i + 1), 12.0D, 16.0D));
+        for (int i = 0; i < 4; i++)
+            shapes[12 + i] = VoxelShapes.or(shapes[11 + i], Block.box(4 * i, 12.0D, 0.0D, 4 * (i + 1), 16.0D, 16.0D));
```
alcatrazEscapee: stop skipping on the `{ }` me no likey

src/main/java/net/dries007/tfc/common/blocks/PitKilnBlock.java
```diff
@@ -0,0 +1,197 @@
...
+
+    public PitKilnBlock(ForgeBlockProperties properties)
+    {
+        super(properties.properties());
+        this.properties = properties;
+        registerDefaultState(getStateDefinition().any().setValue(STAGE, 0));
+    }
+
+    @Override
+    public void animateTick(BlockState stateIn, World worldIn, BlockPos pos, Random rand)
```
alcatrazEscapee: Missing `OnlyIn`

src/main/java/net/dries007/tfc/common/blocks/PitKilnBlock.java
```diff
@@ -0,0 +1,197 @@
...
+            {
+                worldIn.addAlwaysVisibleParticle(ParticleTypes.CAMPFIRE_SIGNAL_SMOKE, x, y, z, 0, 0.1f + rand.nextFloat() / 8, 0);
+            }
+        }
+    }
+
+    @Override
+    protected void createBlockStateDefinition(StateContainer.Builder<Block, BlockState> builder)
+    {
+        builder.add(STAGE);
```
alcatrazEscapee: Missing `super` call

alcatrazEscapee (changes requested):
Some more work. Still haven't properly taken a look at the new berry bush or fruit tree mechanics, just going through looking for organizational + common errors.

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
eerussianguy: I was really hoping to get away with this

src/main/java/net/dries007/tfc/common/blocks/LogPileBlock.java
eerussianguy: This was actually a big bug that I was chasing a while. The problem is essentially that you only want to drop the inventory when you've actually broken the block (not when it's being manipulated in some way). This is the most natural place for that.

Other option is to call te.Remove() a bunch of times / clear the inventory a lot

src/main/java/net/dries007/tfc/common/blocks/berry_bush/BerryBush.java
eerussianguy: Berry is just a temp enum for me to register berry items. It will eventually be deleted when we start registering food.

The noop bit is essentially a fiction I created for inheritance purposes. Nearly everything fruit tree and berry bush related is a subclass of AbstractBerryBushBlock. So you need to be able to sometimes halt the previous logic

src/main/java/net/dries007/tfc/common/blocks/CharcoalPileBlock.java
eerussianguy: discord

src/main/java/net/dries007/tfc/common/blocks/CharcoalPileBlock.java
eerussianguy: We love object oriented don't we folks 👍

src/main/java/net/dries007/tfc/common/blocks/PlacedItemBlock.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public VoxelShape getCollisionShape(BlockState state, IBlockReader worldIn, BlockPos pos, ISelectionContext context)
+    {
+        return VoxelShapes.empty();
+    }
+
+    public boolean isBlank(BlockState state)
```
alcatrazEscapee: Rather call this `isEmpty`

src/main/java/net/dries007/tfc/common/blocks/TFCTorchBlock.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public ActionResultType use(BlockState state, World world, BlockPos pos, PlayerEntity player, Hand hand, BlockRayTraceResult result)
+    {
+        if (!world.isClientSide())
+        {
+            ItemStack held = player.getItemInHand(hand);
+            if (held.getItem().is(Tags.Items.RODS_WOODEN))
```
alcatrazEscapee: Add an item tag: `tfc:can_be_lit_on_torch`. Then reference `forge:rods/wooden` as the only entry.
- No need for a config option here for this mechanic
- Torch creation is now customizable beyond on/off

src/main/java/net/dries007/tfc/common/blocks/TFCWallTorchBlock.java
```diff
@@ -0,0 +1,82 @@
...
+import net.minecraft.world.World;
+import net.minecraft.world.server.ServerWorld;
+import net.minecraftforge.common.Tags;
+import net.minecraftforge.items.ItemHandlerHelper;
+
+import net.dries007.tfc.common.tileentity.TickCounterTileEntity;
+import net.dries007.tfc.config.TFCConfig;
+import net.dries007.tfc.util.Helpers;
+
+public class TFCWallTorchBlock extends WallTorchBlock implements IForgeBlockProperties
```
alcatrazEscapee, line 31: This is a great example of unfortunate sideways inheritance (i.e. we'd like to not have to duplicate these new methods across both `TorchBlock` and `WallTorchBlock`. *However*, there's a nice trick we can do to make it a little less painful (and avoid having to copy paste)

```java
public ActionResultType use(BlockState state, World world, BlockPos pos, PlayerEntity player, Hand hand, BlockRayTraceResult result) {
    return TFCBlocks.TORCH.get().use(state, world, pos, player, hand, result);
}
```

That, or, extract to static methods on `TFCTorchBlock` and make both places delegate there. In any case, means we don't copy paste the same methods across two different classes. That's what methods are for :P

src/main/java/net/dries007/tfc/common/blocks/devices/FirepitBlock.java
```diff
@@ -0,0 +1,264 @@
...
+    );
+
+    public static boolean canSurvive(IWorldReader world, BlockPos pos)
+    {
+        return world.getBlockState(pos.below()).isFaceSturdy(world, pos, Direction.UP);
+    }
+
+    protected static void tryExtinguish(World world, BlockPos pos, BlockState state)
+    {
+        Helpers.playSound(world, pos, SoundEvents.FIRE_EXTINGUISH);
```
alcatrazEscapee: This should really be in the null check

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -472,4 +489,25 @@ public static void destroyBlockAndDropBlocksManually(World worldIn, BlockPos pos
             worldIn.setBlock(pos, fluidstate.createLegacyBlock(), 3, 512);
         }
     }
+
+    public static double offset(Random rand)
```
alcatrazEscapee:
```suggestion
    public static double fastGaussian(Random rand)
```

src/main/java/net/dries007/tfc/network/PlaceBlockSpecialPacket.java
```diff
@@ -0,0 +1,85 @@
...
+
+    PlaceBlockSpecialPacket(PacketBuffer buffer)
+    {
+
+    }
+
+    void encode(PacketBuffer buffer)
+    {
+
+    }
```
alcatrazEscapee: This is just wasted space.

src/main/java/net/dries007/tfc/network/PlaceBlockSpecialPacket.java
```diff
@@ -0,0 +1,85 @@
...
+
+    void handle(Supplier<NetworkEvent.Context> contextSupplier)
+    {
+        NetworkEvent.Context context = contextSupplier.get();
+
+        ServerPlayerEntity player = context.getSender();
+        if (player != null)
+        {
+            World world = player.getLevel();
+            PlayerController mode = Minecraft.getInstance().gameMode;
```
alcatrazEscapee: Client only code on server only execution

src/main/java/net/dries007/tfc/network/PacketHandler.java
```diff
@@ -36,5 +36,6 @@ public static void init()
         CHANNEL.registerMessage(id++, ChunkUnwatchPacket.class, ChunkUnwatchPacket::encode, ChunkUnwatchPacket::new, ChunkUnwatchPacket::handle);
         CHANNEL.registerMessage(id++, CalendarUpdatePacket.class, CalendarUpdatePacket::encode, CalendarUpdatePacket::new, CalendarUpdatePacket::handle);
         CHANNEL.registerMessage(id++, SwitchInventoryTabPacket.class, SwitchInventoryTabPacket::encode, SwitchInventoryTabPacket::new, SwitchInventoryTabPacket::handle);
+        CHANNEL.registerMessage(id++, PlaceBlockSpecialPacket.class, PlaceBlockSpecialPacket::encode, PlaceBlockSpecialPacket::new, PlaceBlockSpecialPacket::handle);
```
alcatrazEscapee:
```suggestion
        CHANNEL.registerMessage(id++, PlaceBlockSpecialPacket.class, buffer -> {}, buffer -> new PlaceBlockSpecialPacket(), PlaceBlockSpecialPacket::handle);
```

src/main/java/net/dries007/tfc/config/ServerConfig.java
```diff
@@ -44,8 +46,15 @@
     // Blocks - Cobblestone
     public final Cache.Boolean enableMossyRockSpreading;
     public final Cache.Int mossyRockSpreadRate;
+    // Blocks - Torch
+    public final Cache.Int torchTime;
+    // Blocks - Charcoal Pit
+    public final Cache.Int charcoalTicks;
+    // Blocks - Pit Kiln
+    public final Cache.Int pitKilnTicks;
```
alcatrazEscapee: This is marvelously inconsistent (call them all `ticks`)

src/main/java/net/dries007/tfc/common/types/Rock.java
```diff
@@ -117,11 +117,11 @@ public boolean isNaturallyGenerating()
     {
         RAW((rock, self) -> new Block(Block.Properties.of(Material.STONE).sound(SoundType.STONE).strength(2, 10).harvestLevel(0).harvestTool(ToolType.PICKAXE)), true),
         HARDENED((rock, self) -> new Block(Block.Properties.of(Material.STONE).sound(SoundType.STONE).strength(2, 10).harvestLevel(0).harvestTool(ToolType.PICKAXE)), false),
-        SMOOTH((rock, self) -> new Block(Block.Properties.of(Material.STONE).sound(SoundType.STONE).strength(1.5f, 10).harvestLevel(0).harvestTool(ToolType.PICKAXE)), true),
+        SMOOTH((rock, self) -> new Block(Block.Properties.of(Material.STONE).sound(SoundType.BONE_BLOCK).strength(1.5f, 10).harvestLevel(0).harvestTool(ToolType.PICKAXE)), true),
```
alcatrazEscapee: As lovely as the bone block sound is - and believe me, I do enjoy it - this makes no sense.

src/main/java/net/dries007/tfc/common/types/Metal.java
```diff
@@ -156,6 +156,11 @@ public boolean hasUtilities()
             return utility;
         }
 
+        public boolean hasRods()
+        {
+            return this == WROUGHT_IRON || this == GOLD || this == STEEL;
+        }
+
```
alcatrazEscapee: Is there any reason we can't classify rods in the exact same manner as, i.e. sheets? (just keep them in part?) The only reason they weren't added like in 1.12 was because they were an afterthought just for rails. Having them generically would be simpler, and just more overall useful for potential packmakers / recipes.

src/main/java/net/dries007/tfc/common/types/Fuel.java
```diff
@@ -0,0 +1,68 @@
...
+    private final boolean isBloomeryFuel;
+
+    public Fuel(ResourceLocation id, JsonObject obj)
+    {
+        this.id = id;
+        this.ingredient = CraftingHelper.getIngredient(JSONUtils.getAsJsonObject(obj, "ingredient"));
+        this.amount = JSONUtils.getAsInt(obj, "amount");
+        this.temperature = JSONUtils.getAsFloat(obj, "temperature");
+        this.isForgeFuel = JSONUtils.getAsBoolean(obj, "isForgeFuel", false);
+        this.isBloomeryFuel = JSONUtils.getAsBoolean(obj, "isBloomeryFuel", false);
```
alcatrazEscapee: json fields should all be `lower_snake_case` (Although I commented this before then later suggesting removing them outright, so just keep in mind for future reference.)

src/main/java/net/dries007/tfc/common/types/FuelManager.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        Fuel fuel = get(stack);
+        return fuel != null && fuel.isBloomeryFuel();
+    }
+
+    public static boolean isFirepitFuel(ItemStack stack)
+    {
+        Fuel fuel = get(stack);
+        return fuel != null && !fuel.isBloomeryFuel() && !fuel.isForgeFuel();
+    }
```
alcatrazEscapee: There's a number of things here:
- these convenience methods are rather asking for a definition of a `Optional<Fuel> get(ItemStack)`
- fire pit fuels being inferred from lack of bloomery and forge fuel is annoyingly restrictive
- This would be much better if there were just tags for valid fuels in specific devices which restrict them. For instance, `tfc:firepit_fuels`, `tfc:bloomery_fuels`, and `tfc:blast_furnace_fuels`.

src/main/java/net/dries007/tfc/common/types/Fuel.java
```diff
@@ -0,0 +1,68 @@
...
+    private final int amount;
+    private final float temperature;
+    private final boolean isForgeFuel;
+    private final boolean isBloomeryFuel;
+
+    public Fuel(ResourceLocation id, JsonObject obj)
+    {
+        this.id = id;
+        this.ingredient = CraftingHelper.getIngredient(JSONUtils.getAsJsonObject(obj, "ingredient"));
+        this.amount = JSONUtils.getAsInt(obj, "amount");
```
alcatrazEscapee: Can we call this `duration` instead because my first thought seeing `amount` was "I don't think any of our devices support consuming fuel in stacks"

src/main/java/net/dries007/tfc/common/tileentity/TickCounterTileEntity.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public long getTicksSinceUpdate()
+    {
+        return Calendars.SERVER.getTicks() - lastUpdateTick;
+    }
+
+    public void resetCounter()
+    {
+        lastUpdateTick = Calendars.SERVER.getTicks();
```
alcatrazEscapee, line 36: any modification needs to call `markDirty` (in this case, since there's no comparator change, `markDirtyFast` is better)

src/main/java/net/dries007/tfc/common/tileentity/TFCTileEntities.java
```diff
@@ -26,8 +29,24 @@
 {
     public static final DeferredRegister<TileEntityType<?>> TILE_ENTITIES = DeferredRegister.create(ForgeRegistries.TILE_ENTITIES, MOD_ID);
 
-    public static final RegistryObject<TileEntityType<FarmlandTileEntity>> FARMLAND = register("farmland", FarmlandTileEntity::new, TFCBlocks.SOIL.get(SoilBlockType.FARMLAND).values());
+    public static final Collection<RegistryObject<Block>> SAPLING_LIST = new ArrayList<>(TFCBlocks.WOODS.size());
+    static
+    {
+        TFCBlocks.WOODS.forEach((key, value) -> SAPLING_LIST.add(value.get(Wood.BlockType.SAPLING)));
+    }
```
alcatrazEscapee: This weird little business here could be
```java
TFCBlocks.WOODS.values().stream().map(map -> map.get(Wood.BlockType.SAPLING))
```
and declared inline, and not public static final

src/main/java/net/dries007/tfc/common/tileentity/TFCTileEntities.java
```diff
@@ -26,8 +29,24 @@
...
+
+    public static final RegistryObject<TileEntityType<FarmlandTileEntity>> FARMLAND = register("farmland", FarmlandTileEntity::new, TFCBlocks.SOIL.get(SoilBlockType.FARMLAND).values().stream());
     public static final RegistryObject<TileEntityType<SnowPileTileEntity>> SNOW_PILE = register("snow_pile", SnowPileTileEntity::new, TFCBlocks.SNOW_PILE);
+    public static final RegistryObject<TileEntityType<FirepitTileEntity>> FIREPIT = register("firepit", FirepitTileEntity::new, TFCBlocks.FIREPIT);
+    public static final RegistryObject<TileEntityType<GrillTileEntity>> GRILL = register("grill", GrillTileEntity::new, TFCBlocks.GRILL);
+    public static final RegistryObject<TileEntityType<PotTileEntity>> POT = register("pot", PotTileEntity::new, TFCBlocks.POT);
+    public static final RegistryObject<TileEntityType<TickCounterTileEntity>> TICK_COUNTER = register("tick_counter", TickCounterTileEntity::new, Stream.of(SAPLING_LIST, TFCBlocks.FRUIT_TREE_SAPLINGS.values(), TFCBlocks.FRUIT_TREE_GROWING_BRANCHES.values(), Arrays.asList(TFCBlocks.TORCH, TFCBlocks.WALL_TORCH, TFCBlocks.DEAD_BERRY_BUSH, TFCBlocks.DEAD_CANE, TFCBlocks.BANANA_SAPLING)).flatMap(Collection::stream));
```
alcatrazEscapee, line 39: Instead of
`Stream.of(...).flatMap(Collection::stream)`
In cases where you have to make calls to `Arrays.asList` or `Collections.singletonList` which look really ungangly in stream calls, you can either 1. extract the single elements out of the `flatMap` call, or use `Stream.concat` to join individual elements into the stream. Or stream the elements and ignore the flat map completely.

src/main/java/net/dries007/tfc/common/recipes/SimplePotRecipe.java
```diff
@@ -0,0 +1,239 @@
...
+
+import com.mojang.serialization.JsonOps;
+import net.dries007.tfc.util.Helpers;
+
+//todo: much of the functionality is here, just needs to be finished
+public class SimplePotRecipe implements IPotRecipe
+{
+    protected final ResourceLocation id;
+    protected final HashSet<Ingredient> inputItems;
+    protected final NonNullList<ItemStack> outputItems;
```
alcatrazEscapee, line 33: Is there any reason why we use `NonNullList` here? I don't particularly like it as it's missing helpful functionality of `ArrayList` or `LinkedList`, and the copy into the latter kind of suggests that.

src/main/java/net/dries007/tfc/common/recipes/SimplePotRecipe.java
```diff
@@ -0,0 +1,239 @@
...
+
+    @Override
+    public boolean matches(FluidInventoryRecipeWrapper wrapper, World worldIn)
+    {
+        if (!wrapper.getInputFluid().isFluidEqual(inputFluid)) return false;
+
+        HashSet<Ingredient> notFound = new HashSet<>(inputItems);
+        notFound.removeIf(Ingredient::isEmpty); // clear the clutter out
+
+        HashSet<ItemStack> stacks = new HashSet<>();
```
alcatrazEscapee: Don't use a hash set for item stacks they don't implement either `equals` or `hashCode` in any meaningful way.

src/main/java/net/dries007/tfc/common/recipes/SimplePotRecipe.java
```diff
@@ -0,0 +1,239 @@
...
+            for (Ingredient i : notFound)
+            {
+                if (i.test(stack))
+                {
+                    notFound.remove(i);
+                    break; // removeIf not used so we can handle duplicates
+                }
+            }
+        }
+        return notFound.isEmpty();
```
alcatrazEscapee: Two ingredients that match the same item stack will fail here I think. This needs to properly assign each stack to a matching ingredient. Although that quickly becomes a non trivial implementation of parsing a [unique perfect matching from a half graph](https://en.wikipedia.org/wiki/Half_graph#Matching), which... might be overboard? but consider:

spicy soup:
- 4 food ingredients
- one spicy food

that would fail under both the single-stack-matches-multiple-ingredients issue, and would potentially fail with an incorrectly labeled graph matching.

Alternatively, we restrict to one ingredient and perform a linear check on all ingredients.

src/main/java/net/dries007/tfc/common/recipes/IItemRecipe.java
```diff
@@ -0,0 +1,12 @@
+package net.dries007.tfc.common.recipes;
+
+import net.minecraft.world.World;
+
+public interface IItemRecipe extends ISimpleRecipe<ItemStackRecipeWrapper>
```
alcatrazEscapee, line 11: This seems pointless? `matches` will need to be overridden by subclasses always.

src/main/java/net/dries007/tfc/common/recipes/HeatingRecipe.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public Collection<Item> getValidItems()
+    {
+        return Arrays.stream(this.ingredient.getItems()).map(ItemStack::getItem).collect(Collectors.toSet());
+    }
+
+    public static class Serializer extends RecipeSerializer<HeatingRecipe>
+    {
+        private final HeatingRecipe.Serializer.Factory<HeatingRecipe> factory;
```
alcatrazEscapee: There's no point having this factory pattern for a recipe without subclasses (or indeed, serializer subclasses.) Just instantiate the serializer as a singleton with the recipe constructor.

src/main/java/net/dries007/tfc/common/recipes/HeatingRecipe.java
```diff
@@ -0,0 +1,130 @@
...
+import net.minecraft.network.PacketBuffer;
+import net.minecraft.util.JSONUtils;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.world.World;
+
+import net.dries007.tfc.common.capabilities.heat.HeatCapability;
+import net.dries007.tfc.common.capabilities.heat.IHeat;
+import net.dries007.tfc.util.collections.IndirectHashCollection;
+
+public class HeatingRecipe extends SimpleItemRecipe
```
alcatrazEscapee, line 30: Heat recipes can produce either items or fluids, this chosen to be unimplemented or forgotten?

src/main/java/net/dries007/tfc/common/tileentity/PotTileEntity.java
```diff
@@ -0,0 +1,199 @@
...
+    private static final ITextComponent NAME = new TranslationTextComponent(MOD_ID + ".tile_entity.pot");
+
+    public static final int SLOT_EXTRA_INPUT_START = 4;
+    public static final int SLOT_EXTRA_INPUT_END = 8;
+
+    private IPotRecipe.Output output;
+    private IPotRecipe cachedPotRecipe;
+    private int boilingTicks;
+    protected FluidTank tank;
+    protected final LazyOptional<IFluidHandler> fluidCapability;
```
alcatrazEscapee, line 49: Like the small vessel in 1.12, imo the best way to approach this would be to have a single joint capability implementation which holds both fluids and items. It gets rid of the need for multiple capabilities (as we can cast to `PotTileEntity.IPotCapability` with `.filter(...).map(...)` in `LazyOptional` declarations), and removes the need for `FluidInventoryRecipeWrapper` entirely - you just pass in the pot capability interface, within a `RecipeWrapper` subclass that's generic on the interface type. That makes the wrappers simpler, as the recipe itself becomes `PotRecipe extends SimpleRecipe<PotCapability>`

alcatrazEscapee (changes requested):
No gigantic mega "throw all things in" prs again please 😔

src/main/java/net/dries007/tfc/common/blocks/LogPileBlock.java
alcatrazEscapee: Yeah, but we can still avoid the copy-paste + fix the lack of super call.
- Add `class DeviceBlock extends Block`, as for most inventory tile entity blocks they will just be extending `Block` rather than basing off vanilla functionality. Override the method there, but,
  - Call `super` rather than removing the block entity yourself
  - The only necessary check there is the `state.is(newState.getBlock())` for our intents and purposes
  - Delegate to `InventoryTileEntity#onRemove` (renamed from `onBreak` as it's a passthrough)
- Throw the basic `IForgeProperties` / `ForgeBlockProperties` impl. onto `DeviceBlock`

src/main/java/net/dries007/tfc/config/ServerConfig.java
eerussianguy: you can have multiple variables with the same name?

eerussianguy: Oh I guess torch time is wrong.

src/main/java/net/dries007/tfc/common/tileentity/TickCounterTileEntity.java
eerussianguy: Calling markDirty here actually breaks stuff. Because for some reason the TE is not fully created if you place it and then call resetCounter() right afterwards. It will send 1 invalid packet before becoming normal again. I will have to talk with you about this at some point

src/main/java/net/dries007/tfc/common/tileentity/TFCTileEntities.java
eerussianguy: I tried for about 20 minutes to figure out how to get streams to work based on what you said and did not get anywhere, so i will have to leave it for now

src/main/java/net/dries007/tfc/common/tileentity/InventoryTileEntity.java
```diff
@@ -0,0 +1,138 @@
...
+    protected final LazyOptional<IItemHandler> inventoryCapability;
+    protected ITextComponent customName, defaultName;
+
+    public InventoryTileEntity(TileEntityType<?> type, int inventorySlots, ITextComponent defaultName)
+    {
+        super(type);
+
+        this.inventory = new ItemStackHandlerCallback(this, inventorySlots);
+        this.inventoryCapability = LazyOptional.of(() -> InventoryTileEntity.this.inventory);
+        this.defaultName = defaultName;
```
alcatrazEscapee, line 44:
```suggestion
        this(type, new ItemStackHandlerCallback(this, inventorySlots), defaultName);
```

src/main/java/net/dries007/tfc/common/tileentity/InventoryTileEntity.java
```diff
@@ -0,0 +1,138 @@
...
+import net.minecraftforge.common.util.LazyOptional;
+import net.minecraftforge.items.CapabilityItemHandler;
+import net.minecraftforge.items.IItemHandler;
+import net.minecraftforge.items.ItemStackHandler;
+
+import net.dries007.tfc.common.container.ISlotCallback;
+import net.dries007.tfc.common.container.ItemStackHandlerCallback;
+import net.dries007.tfc.util.Helpers;
+
+public abstract class InventoryTileEntity extends TFCTileEntity implements INamedContainerProvider, ISlotCallback, IClearable
```
alcatrazEscapee: This shouldn't implement `INamedContainerProvider` because it's reasonable to have an inventory without a gui (*or so I keep hearing about* :P )

src/main/java/net/dries007/tfc/common/tileentity/InventoryTileEntity.java
```diff
@@ -0,0 +1,138 @@
...
+    public void onReplaced()
+    {
+        inventoryCapability.invalidate();
+    }
+
+    @Override
+    public Container createMenu(int p_createMenu_1_, PlayerInventory p_createMenu_2_, PlayerEntity p_createMenu_3_)
+    {
+        return null;
+    }
```
alcatrazEscapee: Likewise this here is waiting for bugs - if it *was* to be implemented, the class should be abstract and implemented by the subclass. Here it will always be an unneeded override.

src/main/java/net/dries007/tfc/common/tileentity/InventoryTileEntity.java
```diff
@@ -0,0 +1,138 @@
...
+
+    public boolean canInteractWith(PlayerEntity player)
+    {
+        return true;
+    }
+
+    public void onReplaced()
+    {
+        inventoryCapability.invalidate();
+    }
```
alcatrazEscapee: Not called from anywhere, invalidate should be done in `onRemove`

src/main/java/net/dries007/tfc/common/tileentity/PotTileEntity.java
```diff
@@ -0,0 +1,199 @@
...
+        boilingTicks = 0;
+        tank = new FluidTank(1000); //todo: predicate whitelist? custom tank impl?
+        fluidCapability = LazyOptional.of(() -> tank);
+    }
+
+    @Override
+    public void load(BlockState state, CompoundNBT nbt)
+    {
+        tank.readFromNBT(nbt.getCompound("tank"));
+        if (nbt.hasUUID("output"))
```
alcatrazEscapee:
```suggestion
        if (nbt.contains("output"))
```
wrong method?

alcatrazEscapee (commented):
A couple more minor things, one which looks like a bug with pot output saving?

src/main/java/net/dries007/tfc/common/tileentity/InventoryTileEntity.java
eerussianguy: bad

alcatrazEscapee: You could make this work but it would require another constructor with a `Function<InventoryTileEntity, IItemHandlerModifiable>` so w/e.

PR #1732, opened 2021-03-10
full HUD implementation

eerussianguy:


src/main/java/net/dries007/tfc/client/screen/overlay/PlayerDataOverlay.java
```diff
@@ -0,0 +1,200 @@
...
+
+    @SuppressWarnings("deprecation")
+    @SubscribeEvent
+    public static void render(RenderGameOverlayEvent.Pre event)
+    {
+        Minecraft mc = Minecraft.getInstance();
+        IngameGui gui = mc.gui;
+        ClientPlayerEntity clientPlayer = mc.player;
+        if (clientPlayer == null) return;
+        PlayerEntity player = clientPlayer.inventory.player;
```
alcatrazEscapee, line 45: This is entirely redundant, `clientPlayer == player` will always be true

src/main/java/net/dries007/tfc/client/screen/overlay/PlayerDataOverlay.java
```diff
@@ -0,0 +1,200 @@
...
+        int armorRowHeight = healthRowHeight - 10;
+        int mid = guiScaledWidth / 2;
+
+        FontRenderer fontRenderer = mc.font;
+        RenderSystem.enableBlend();
+        RenderSystem.color4f(1.0f, 1.0f, 1.0f, 1.0f);
+        TextureManager textureManager = mc.getTextureManager();
+        textureManager.bind(ICONS);
+
+        if (!clientPlayer.abilities.instabuild)
```
alcatrazEscapee, line 77: I *think* this should be checking `PlayerEntity#isCreative` rather than `abilities.instabuild`

src/main/java/net/dries007/tfc/util/config/HealthDisplayFormat.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /**
+     * Shows this text in config instead of the enum name
+     */
+    @Override
+    public String toString()
+    {
+        return name;
+    }
```
alcatrazEscapee, line 26: No, don't do this.

src/main/java/net/dries007/tfc/config/ClientConfig.java
```diff
@@ -24,17 +25,34 @@
...
             "1. (Requires a world restart) Cloud height is moved from 160 -> 210",
             "2. The 'horizon height' (where the fog changes from sky to black) is moved from 63 -> 96"
         ).define("assumeTFCWorld", true));
+
+        innerBuilder.pop().push("display");
+
+        useVanillaHunger = wrap(builder.apply("useVanillaHunger").comment("Render vanilla hunger over the hotbar?").define("useVanillaHunger", false));
+        useVanillaHealth = wrap(builder.apply("useVanillaHunger").comment("Render vanilla health bar over the hotbar?").define("useVanillaHealth", false));
+        hideThirstBar = wrap(builder.apply("hideThirstBar").comment("Hide the TFC thirst bar?").define("hideThirstBar", false));
+        healthDisplayFormat = wrap(builder.apply("healthDisplayFormat").comment("Health display format. Options: TFC, VANILLA, TFC_CURRENT_HEALTH, VANILLA_CURRENT_HEALTH").define("healthDisplayFormat", HealthDisplayFormat.VANILLA));
```
alcatrazEscapee, line 54: The options are already commented as it's an enum. However, it would be useful to provide info about what they actually do. i.e.
```
.comment(
    "TFC - Show health as 750 / 1000",
    "VANILLA - Show health as 15 / 20",
// ... etc.
```

It should also be using `defineEnum` instead of `define` and I don't believe this option would actually work as is.

alcatrazEscapee (changes requested):

PR #1740, opened 2021-03-20
Fixes #1736 after recommendation by Alcatraz 

### LINKED ISSUE #1736: prospecting false negatives showing lack of randomness
As previously discussed, I've experienced a Z-axis bias in positive and false negative propicking. 

For the blocks at x=-111, y=144, and z from 300 to 320, I saw false negative propick results on EVERY block.
For the adjacent blocks at x=-112, y=144, and z from 300 to 320, I saw VL vein sample propick results on EVERY block.

Here's a screenshot of the area.

 - TFC Version: build 175
 - TNFC modpack, 12.2.62.1

Have not reproduced in TFC alone. Recently heard from one other TNFC player who experienced something similar with 17 blocks in a row. They didn't specify which axis. 
Link to their message:
https://discord.com/channels/104658769123753984/104658769123753984/820879646995513344

Gaelmare:
also after 1.16 version of SharedSeedRandomMixin


Gaelmare:
Tested at same location as the random world screenshot in #1736, results look much better, even in y-axis.

White wool=positive
Nether Bricks=false negative

PR #1749, opened 2021-04-06
Query tool classes to determine if item should be on the tool rack

Rongmario:
- Better way of determining if an item is a tool than checking for an oredict name `tool` as its not the correct convention.


alcatrazEscapee:
Ore dict is end user editable. Tool classes are not. Feel free to add tool classes as an automatic but leave the ore dict there.

PR #1752, opened 2021-04-12
Add Support Beams

eerussianguy:
still needs more thorough bug catching / tests / place logic?


src/main/java/net/dries007/tfc/common/blocks/wood/VerticalSupportBlock.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        VoxelShape shape = getMiddleShape();
+        for (Direction d : Direction.Plane.HORIZONTAL)
+        {
+            if (state.getValue(PROPERTY_BY_DIRECTION.get(d)))
+            {
+                shape = VoxelShapes.or(shape, shapeByDirection(d));
+            }
+        }
+        return shape;
```
alcatrazEscapee: Since `VoxelShape` unions can be relatively expensive (and `getShape` called a lot), it's good practice to cache these, either from property or the state. For example, see `WallBlock#makeShapes` in vanilla - it builds a `Map<BlockState, VoxelShape>` and then that is used in this method to query.

Additionally, because of this, `getMiddleShape` need not be an overridable method, it can become a constructor parameter since this is it's only practical usage.

alcatrazEscapee (changes requested):
Just one minor note about `VoxelShape` caches, otherwise looks good

PR #1762, opened 2021-05-07
pipe veins + dikes

eerussianguy:


src/main/java/net/dries007/tfc/world/feature/vein/DikeVeinFeature.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.MutableBoundingBox;
+import net.minecraft.world.ISeedReader;
+
+import com.mojang.serialization.Codec;
+import net.dries007.tfc.common.types.Rock;
+import net.dries007.tfc.world.chunkdata.ChunkData;
+import net.dries007.tfc.world.chunkdata.ChunkDataProvider;
+
+public class DikeVeinFeature extends PipeVeinFeature
```
alcatrazEscapee: Basically I don't see why this needs to be a thing at all. My vision of dikes was for them to just unrelated rocks - not necessarily the bottom rock, and this hardcoding in any case is rather gross.

src/main/java/net/dries007/tfc/world/feature/vein/PipeVeinConfig.java
```diff
@@ -14,23 +14,55 @@
 {
     public static final Codec<PipeVeinConfig> CODEC = RecordCodecBuilder.create(instance -> instance.group(
         VeinConfig.MAP_CODEC.forGetter(c -> c),
-        Codecs.POSITIVE_INT.optionalFieldOf("radius", 3).forGetter(PipeVeinConfig::getRadius)
+        Codecs.POSITIVE_INT.optionalFieldOf("radius", 3).forGetter(PipeVeinConfig::getRadius),
+        Codecs.POSITIVE_INT.optionalFieldOf("minSkew", 0).forGetter(PipeVeinConfig::getMinSkew),
+        Codecs.POSITIVE_INT.optionalFieldOf("maxSkew", 0).forGetter(PipeVeinConfig::getMaxSkew),
+        Codecs.POSITIVE_INT.optionalFieldOf("minSlant", 0).forGetter(PipeVeinConfig::getMinSlant),
+        Codecs.POSITIVE_INT.optionalFieldOf("maxSlant", 0).forGetter(PipeVeinConfig::getMaxSlant)
```
alcatrazEscapee: Serialization fields should be `snake_case`

src/main/java/net/dries007/tfc/world/feature/vein/PipeVeinFeature.java
```diff
@@ -8,36 +8,63 @@
...
     }
 
     @Override
-    protected float getChanceToGenerate(int x, int y, int z, Vein vein, PipeVeinConfig config)
+    protected float getChanceToGenerate(int x, int y, int z, PipeVein vein, PipeVeinConfig config)
     {
-        if (Math.abs(y) < config.getSize() && (x * x) + (z * z) < config.getRadius() * config.getRadius())
+        final double yScaled = (double) y / config.getSize();
+        x += vein.skew * MathHelper.cos(vein.angle) * yScaled;
+        z += vein.skew * MathHelper.sin(vein.angle) * yScaled;
```
alcatrazEscapee: These can be precomputed on the vein itself - i.e. from `vein.angle` and `vein.skew`, compute `vein.skewX` and `vein.skewZ`. Since this will be called a lot more often.

src/main/java/net/dries007/tfc/world/feature/vein/PipeVeinFeature.java
```diff
@@ -8,36 +8,63 @@
...
-        return new Vein(defaultPos(chunkX, chunkZ, random, config));
+        final int yFlipper;
+        final float angle;
+        final int skew;
+        final int slant;
+
+        PipeVein(BlockPos pos, Random random, PipeVeinConfig config)
+        {
+            super(pos);
+            this.yFlipper = random.nextBoolean() ? 1 : -1;
```
alcatrazEscapee: This should
a. be called `sign` not `yFlipper` :P
b. be a field in json - `sign`, which is a float between 0 and 1. Then `vein.sign = type.sign < random.nextFloat() ? -1 : 1`

- If `sign = 0`, then it will always set to 1
- If `sign = 1`, then it will always set to -1
- Anywhere between is a linear probability between the two.

Preferably document so we know what the probability is actually of - i.e. which cone shape.

src/main/java/net/dries007/tfc/world/feature/vein/VeinFeature.java
```diff
@@ -105,7 +105,7 @@ protected void place(ISeedReader world, Random random, int blockX, int blockZ, V
     }
 
     @Nullable
-    protected BlockState getStateToGenerate(BlockState stoneState, Random random, C config)
+    protected BlockState getStateToGenerate(BlockState stoneState, Random random, C config, ISeedReader world, BlockPos pos)
```
alcatrazEscapee: Minor nitpick, but almost everywhere else parameters are in the order world, pos, state. (IDEA can refactor this with a single dropdown menu), it's a bit weird to stick them on the end (given it's not a binary compat issue)

alcatrazEscapee (changes requested):

alcatrazEscapee:
dd533f78a29f0d0ecd2a0a8154c6607f5938fe91

PR #1766, opened 2021-05-15
Update zh_cn.lang

HeartyYF:
Seemingly 1.12 comes to an end of new features, and now I wonder if there's something that can turn this .lang into a json that can be used in 1.16.


eerussianguy:
There is a way to convert, but everything is localized differently in 1.16 so nothing would match, unfortunately.

alcatrazEscapee:
There is! There's even a [handy online tool](https://www.tterrag.com/lang2json/) you can use!

_**But**_

The 1.16 lang file has a great many changes, especially in the bulk naming department, and so I would very hesitate to do any localization by hand, or straight copying from 1.12. An ideal solution, in my eyes, would be to adapt / co-opt the existing python infrastructure, although I'm not sure how feasable (or usable) that would be.

In any case, something to think about, but not pertinent at the moment.

PR #1769, opened 2021-05-22
Fixes #1761

### LINKED ISSUE #1761: Milking animals with a full inventory deletes the bucket stack
**Describe the bug**
If you have a full inventory, including offhand slot, and you milk an animal using a >1 stack of buckets, it deletes the bucket stack.
Milking with a single bucket works as intended.
TFC wooden buckets, TNFC ultimate buckets, and vanilla buckets are all affected the same way.

I uploaded a video demonstrating the bug here:
https://youtu.be/LuObRpEvYds

**To Reproduce**
1. Get a stack of at least 2 buckets and a familiarized cow.
2. Fill every inventory slot including your offhand with items.
3. Milk the cow.
4. The stack of buckets disappears. A bucket of milk is not given.

**Meta Info**
 - TFC Version: 1.7.18.176
 - Both TechNodeFirmaCraft and a clean install of just Forge+TFC.

Charnuz:
This does make it so players in creative milk the animal without getting a bucket of milk again, can fix that if wanted


eerussianguy:
Just to make sure, you tested with multiple and single stacks of buckets? Need to make sure all edge cases are covered in survival mode

Charnuz:
Yes, with multiple it drops the milk bucket on the ground, and single gets replaced with the milk bucket

PR #1787, opened 2021-06-10
Fix #1771 (For real this time)

### LINKED ISSUE #1771: Bellows influence is capped at +300, despite air ticks setting to +600
Since bf04559f67b737f1e42402cc0907be9b61be98db, despite bellows setting air temperature to +600:

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/0029a016f486f342704a24b6890e2b890f9311e3/src/main/java/net/dries007/tfc/objects/te/TECharcoalForge.java#L80

it's being artificially capped here (and in all other devices) to +300:

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/0029a016f486f342704a24b6890e2b890f9311e3/src/main/java/net/dries007/tfc/objects/te/TECharcoalForge.java#L151

due to the fix for #1668.

Charnuz:
could make this a static value somewhere, let me know if thats preferred

### DIFF
--- a/src/main/java/net/dries007/tfc/ConfigTFC.java
+++ b/src/main/java/net/dries007/tfc/ConfigTFC.java
@@ -727,6 +727,9 @@ public final class ConfigTFC
         @Config.LangKey("config." + MOD_ID + ".devices.goldpan")
         public static final GoldPanCFG GOLD_PAN = new GoldPanCFG();
 
+        @Config.Comment("Bellows")
+        @Config.LangKey("config." + MOD_ID + ".devices.bellows")
+        public static final BellowsCFG BELLOWS = new BellowsCFG();
 
         public static final class TemperatureCFG
         {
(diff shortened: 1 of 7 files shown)


eerussianguy:
So we are on our third or so PR trying to tackle bellows interaction. This probably is the last one but to be honest there is a good solution to making this standard:

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/a2303ef14d67bc347327fca0e738cf08f0c043ac/src/main/java/net/dries007/tfc/objects/te/TEBlastFurnace.java#L268-L273

This bit of code only actually updates the `temperature` variable. What I would do is add a method in `CapabilityItemHeat` that takes in the temp, air ticks, max air ticks, etc, and move that common logic there.

Charnuz:
https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/a2303ef14d67bc347327fca0e738cf08f0c043ac/src/main/java/net/dries007/tfc/objects/te/TEBlastFurnace.java#L59 Was there a reason that `airTicks` was a long here? Seems unneeded.

alcatrazEscapee:
Not particularly.

Charnuz:
Added a config for max air ticks, if thats something we dont want exposed to the player I can just scrap that.

PR #1794, opened 2021-07-14
Fix animal respawn

### LINKED ISSUE #1793: Creatures in the RespawnableCreatures config do not respawn or spawn from eggs
Creatures in the RespawnableCreatures config do not respawn ever. Nor do they spawn when a spawn egg is clicked in compatible conditions. 

Make sure to include:
1. What did you expect to happen? Bears continue to terrorize the server
2. What actually happened instead (i.e. what was the bug) The server becomes much less terrifying

**To Reproduce**
1. Do 'Play and kill all the bears'
2. Then do 'Wait until the heat death of the universe for new bears to spawn.

This does work in 1.7.11.169 and prior. Since the .170 version (which added @DisasterMoo [Livestock respawning](https://github.com/TerraFirmaCraft/TerraFirmaCraft/pull/1665/files)) it's been broken.

**Meta Info**
 - TFC Version: TerraFirmaCraft-MC1.12.2-1.7.19.177
 - Were any other mods included? Can you reproduce it without these other mods? Done in bare TFC instance.

AnodeCathode:
Part of the issue is that @DisasterMoo 's changes to domestic animal respawn was nuking all the predators. So I've just changed the check to ensure it's only nuking old livestock. It's a start. Partially fixes #1793

### DIFF
--- a/src/main/java/net/dries007/tfc/objects/entity/animal/EntityAnimalTFC.java
+++ b/src/main/java/net/dries007/tfc/objects/entity/animal/EntityAnimalTFC.java
@@ -34,6 +34,7 @@ import net.minecraftforge.oredict.OreDictionary;
 
 import net.dries007.tfc.Constants;
 import net.dries007.tfc.api.types.IAnimalTFC;
+import net.dries007.tfc.api.types.ILivestock;
 import net.dries007.tfc.api.types.IPredator;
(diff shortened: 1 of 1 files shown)


eerussianguy:
#1793  edit: oh duh you linked it already

PR #1800, opened 2021-07-18
Fixed small Logpile placing bug

neumann-lukas:
Fixed a small bug that you cant place a log pile by shift-right clicking on a full one. This is Important for the construction of charcoal pits.


eerussianguy:
Can you verify that you can place regular logs on the ground

neumann-lukas:
Yes, sorry if I explained it badly, on the ground everything worked ok before, but if you do shift-right click on another FULL logpile, it should place another one like in 1.12. Otherwise you cant fill charcoal pits, because the are no blocks other then logpiles to click on.

PR #1801, opened 2021-07-18
Add the charcoal forge

eerussianguy:


src/main/java/net/dries007/tfc/client/screen/CharcoalForgeScreen.java
```diff
@@ -0,0 +1,41 @@
...
+        super(container, playerInventory, name, FORGE);
+        inventoryLabelY += 20;
+        imageHeight += 20;
+    }
+
+    @Override
+    protected void renderBg(MatrixStack matrixStack, float partialTicks, int mouseX, int mouseY)
+    {
+        super.renderBg(matrixStack, partialTicks, mouseX, mouseY);
+        int temp = (int) (51 * ((CharcoalForgeTileEntity) tile).getSyncableData().get(CharcoalForgeTileEntity.DATA_SLOT_TEMPERATURE) / Heat.maxVisibleTemperature());
```
alcatrazEscapee: This cast shouldn't be necessary. Because the variable here:

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/cc08a576ad03949220f89d78084ee4de0b3a5ad4/src/main/java/net/dries007/tfc/client/screen/TileEntityScreen.java#L18

is typed wrong. That should be `T`, which should then make this cast redundant.

src/main/java/net/dries007/tfc/common/blocks/devices/CharcoalForgeBlock.java
```diff
@@ -0,0 +1,170 @@
...
+    public BlockState updateShape(BlockState state, Direction facing, BlockState facingState, IWorld world, BlockPos currentPos, BlockPos facingPos)
+    {
+        return state.getValue(HEAT) > 0 && !isValid(world, currentPos) ? state.setValue(HEAT, 0) : state;
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public ActionResultType use(BlockState state, World world, BlockPos pos, PlayerEntity player, Hand hand, BlockRayTraceResult result)
+    {
+        if (player instanceof ServerPlayerEntity && !player.isShiftKeyDown())
```
alcatrazEscapee: the `ServerPlayerEntity` should be moved inside. effectively:
```java
if (/* conditions that can be checked on both sides, and may fail */) {
    if (/* server side */) {
        // server side specific actions
    }
    return SUCCESS; // return success on both sides
}
return PASS; // return pass or fail if something didn't, or couldn't happen.
```

I'm also fairly convinced you don't need the shift check at all because shifting bypasses the block interaction by default. Meaning this can just be a TE -> open gui on server -> always return success chain.

src/main/java/net/dries007/tfc/common/tileentity/FirepitTileEntity.java
```diff
@@ -35,6 +36,20 @@
 
     private static final ITextComponent NAME = new TranslationTextComponent(MOD_ID + ".tile_entity.firepit");
 
+    public static FluidStack mergeOutputFluidIntoSlot(IItemHandlerModifiable inventory, FluidStack fluidStack, float temperature, int slot)
```
alcatrazEscapee: This could probably be moved to `Helpers` then.

src/main/java/net/dries007/tfc/common/tileentity/AbstractFirepitTileEntity.java
```diff
@@ -56,6 +57,42 @@ public static void convertTo(IWorld world, BlockPos pos, BlockState state, Abstr
         }
     }
 
+    public static Triple<Integer, Float, Long> consumeFuelForTicks(long deltaPlayerTicks, IItemHandlerModifiable inventory, int burnTicks, float burnTemperature, int slotStart, int slotEnd)
```
alcatrazEscapee: This is gross but w/e. Although probably could move to `Helpers` with our other inventory helper functions.

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/common/blocks/devices/CharcoalForgeBlock.java
```diff
@@ -0,0 +1,170 @@
...
+                .match(new BlockPos(2, 0, 0), skyMatcher)
+                .match(new BlockPos(-1, 0, 0), skyMatcher)
+                .match(new BlockPos(-2, 0, 0), skyMatcher)
+            )
+            // Underneath
+            .match(new BlockPos(1, 0, 0), isValidSide)
+            .match(new BlockPos(-1, 0, 0), isValidSide)
+            .match(new BlockPos(0, 0, 1), isValidSide)
+            .match(new BlockPos(0, 0, -1), isValidSide)
+            .match(new BlockPos(0, -1, 0), isValidSide);
```
alcatrazEscapee: I am *very tempted* to suggest rewriting these like such:
```java
BlockPos origin = BlockPos.ZERO;

// later
.match(origin.east())
.match(origin.down())
.match(origin.east(2))
```
etc.

I think that might look a bit cleaner and be more future proof for if we ever need to edit / understand these values. Thoughts?

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -557,6 +571,57 @@ public static boolean spawnItem(World world, BlockPos pos, ItemStack stack)
         return perfectMatchDet(matrices, size);
     }
 
+    public static Triple<Integer, Float, Long> consumeFuelForTicks(long deltaPlayerTicks, IItemHandlerModifiable inventory, int burnTicks, float burnTemperature, int slotStart, int slotEnd)
```
alcatrazEscapee, line 587: This needs at least a doc comment saying what it does (i.e. what the spec is) and what on earth it is returning. (god how I wish this could return a record :pensive:

PR #1802, opened 2021-07-19
add scraping implementation

eerussianguy:


src/main/java/net/dries007/tfc/client/render/ScrapingTileEntityRenderer.java
```diff
@@ -0,0 +1,71 @@
...
+                drawTiles(buffer, matrixStack, scraped, positions, 1, combinedLight, combinedOverlay);
+            }
+        });
+    }
+
+    private void drawTiles(IRenderTypeBuffer buffer, MatrixStack matrixStack, ResourceLocation texture, short positions, int condition, int combinedLight, int combinedOverlay)
+    {
+        Matrix4f mat = matrixStack.last().pose();
+        IVertexBuilder builder = buffer.getBuffer(RenderType.cutout());
+        //noinspection deprecation
```
alcatrazEscapee: use a `@SuppressWarnings` on the method as that stops the actual java / gradle compile warning.

src/main/java/net/dries007/tfc/common/recipes/ScrapingRecipe.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+        return null;
+    }
+
+    public ScrapingRecipe(ResourceLocation id, Ingredient ingredient, ItemStack result)
+    {
+        super(id, ingredient, result);
+        if (result.getCount() != 1)
+        {
+            throw new IllegalArgumentException("Scraping recipes only support a stack size of 1");
```
alcatrazEscapee: Why? Like in theory we could no?

src/main/java/net/dries007/tfc/common/TFCTags.java
```diff
@@ -91,6 +92,8 @@
         public static final ITag.INamedTag<Item> FORGE_FUEL = create("forge_fuel");
         public static final ITag.INamedTag<Item> HANDSTONE = create("handstone");
         public static final ITag.INamedTag<Item> FLUXSTONE = create("fluxstone");
+        public static final ITag.INamedTag<Item> SCRAPABLE_ITEM = create("scrapable_item");
```
alcatrazEscapee: I'd take the `item` off here and just have `scrapable`

src/main/java/net/dries007/tfc/util/InteractionManager.java
```diff
@@ -234,6 +237,34 @@ public static void setup()
...
+                {
+                    level.setBlockAndUpdate(abovePos, TFCBlocks.SCRAPING.get().defaultBlockState());
+                    final ScrapingTileEntity te = Helpers.getTileEntity(level, abovePos, ScrapingTileEntity.class);
+                    if (te != null)
+                    {
+                        return Helpers.getCapability(te, CapabilityItemHandler.ITEM_HANDLER_CAPABILITY).map(cap -> {
+                            if (!level.isClientSide)
+                            {
+                                player.setItemInHand(context.getHand(), cap.insertItem(0, stack, false));
+                                te.cachedItem = recipe.getResultItem().copy();
```
alcatrazEscapee: This should use `assemble`, and not perform a copy() (because `assemble` does.) `getResultItem` is for display purposes (generally). the same wrapper can be re-used from earlier in this method.

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/util/InteractionManager.java
eerussianguy: cachedItem is specifically for display purposes only. assemble is called when the TE realizes it's complete

src/main/java/net/dries007/tfc/common/recipes/ScrapingRecipe.java
eerussianguy: Makes the insertion logic easier to have a stacksize limit. But I'll handle it

PR #1803, opened 2021-07-21
Knapping + some recipes

eerussianguy:


src/main/java/net/dries007/tfc/client/screen/KnappingScreen.java
```diff
@@ -0,0 +1,110 @@
...
+    {
+        if (clickType == 0)
+        {
+            mouseClicked(x, y, clickType);
+        }
+        return super.mouseDragged(x, y, clickType, dragX, dragY);
+    }
+
+    @Override
+    public boolean mouseClicked(double x, double y, int clickType)
```
alcatrazEscapee, line 95: This seems unnecessary, does the super class not handle this? Since it will call `AbstractButton#onClick` which delegates eventually to the `Button#IPressable` in the ctor of button?

src/main/java/net/dries007/tfc/client/screen/button/KnappingButton.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public KnappingButton(int id, int x, int y, int width, int height, ResourceLocation texture, SoundEvent sound)
+    {
+        super(x, y, width, height, StringTextComponent.EMPTY, KnappingButton::doPress);
+        this.id = id;
+        this.texture = texture;
+        this.sound = sound;
+    }
+
+    public static void doPress(Button widget)
```
alcatrazEscapee: Rather than having this static bouncer method, nonspecific type, and `onPress` -> `doPress` delegation, it's simpler just to pass `button -> {}` in the constructor, and then override `Button#onPress`

src/main/java/net/dries007/tfc/client/screen/button/KnappingButton.java
```diff
@@ -0,0 +1,75 @@
...
+    {
+        if (visible)
+        {
+            Minecraft.getInstance().getTextureManager().bind(texture);
+            isHovered = mouseX >= x && mouseY >= y && mouseX < x + width && mouseY < y + height;
+
+            matrixStack.pushPose();
+            blit(matrixStack, x, y, 0, 0, 16, 16, 16, 16);
+            matrixStack.popPose();
+            onDrag(mouseX, mouseY, 0, 0); // ?
```
alcatrazEscapee: Is this copied from 1.12 code or by looking at 1.16 buttons, i.e. the player inventory tab ones? I'd be careful about doing any of that.

src/main/java/net/dries007/tfc/client/screen/button/KnappingButton.java
```diff
@@ -0,0 +1,75 @@
...
+    public void renderButton(MatrixStack matrixStack, int mouseX, int mouseY, float partialTicks)
+    {
+        if (visible)
+        {
+            Minecraft.getInstance().getTextureManager().bind(texture);
+            isHovered = mouseX >= x && mouseY >= y && mouseX < x + width && mouseY < y + height;
+
+            matrixStack.pushPose();
+            blit(matrixStack, x, y, 0, 0, 16, 16, 16, 16);
+            matrixStack.popPose();
```
alcatrazEscapee: Don't need push + pop if you're not mutating the state I believe

src/main/java/net/dries007/tfc/common/container/IButtonHandler.java
```diff
@@ -0,0 +1,23 @@
...
+
+import javax.annotation.Nullable;
+
+import net.minecraft.nbt.CompoundNBT;
+
+public interface IButtonHandler
+{
+    /**
+     * An interface for containers that need to receive button presses from a client-side GUI
+     * If you implement this interface you should also use {todo} to send update packets from the GUI
```
alcatrazEscapee: Packet reference?

src/main/java/net/dries007/tfc/common/container/KnappingContainer.java
```diff
@@ -0,0 +1,199 @@
...
+
+public class KnappingContainer extends ItemStackContainer implements IButtonHandler, IInventoryNoop
+{
+    public final SimpleCraftMatrix matrix;
+    private final int amountToConsume;
+    public final boolean usesDisabledTex;
+    private final boolean needsKnife;
+    private final boolean consumeAfterComplete;
+    public final SoundEvent sound;
+    public boolean requiresReset;
```
alcatrazEscapee: Encapsulate public non final variables, final ones *ehh* it's preferred esp. if they field is mutable.

src/main/java/net/dries007/tfc/common/container/KnappingContainer.java
```diff
@@ -0,0 +1,199 @@
...
+        this.sound = sound;
+
+        matrix = new SimpleCraftMatrix();
+        hasBeenModified = false;
+        requiresReset = false;
+        hasConsumedIngredient = false;
+        stackCopy = this.stack.copy();
+    }
+
+    private static ItemStack getItemForKnapping(PlayerInventory inv)
```
alcatrazEscapee: Does this not break if you open a knapping gui with a non-empty mainhand but the knapping item in offhand?

src/main/java/net/dries007/tfc/common/container/KnappingContainer.java
```diff
@@ -0,0 +1,199 @@
...
+                ItemHandlerHelper.giveItemToPlayer(player, stack);
+                consumeIngredientStackAfterComplete();
+            }
+        }
+        super.removed(player);
+    }
+
+    /**
+     * Used in client to check a slot state in the matrix
+     * JEI won't cause issues anymore see https://github.com/TerraFirmaCraft/TerraFirmaCraft/issues/718
```
alcatrazEscapee: I'd remove this issue reference since the code in question is basically being rewritten.

src/main/java/net/dries007/tfc/common/container/KnappingContainer.java
```diff
@@ -0,0 +1,199 @@
...
+    {
+        matrix.setAll(false);
+        requiresReset = true;
+        consumeIngredientStackAfterComplete();
+    }
+
+    @Nullable
+    private KnappingRecipe getMatchingRecipe(ServerWorld level)
+    {
+        return level.getRecipeManager().getRecipeFor(recipeType, this, level).orElse(null);
```
alcatrazEscapee, line 191: Ideally should use the (or a) knapping item cache. Indirect based on a packed version of a int key?

src/main/java/net/dries007/tfc/common/container/KnappingContainer.java
```diff
@@ -0,0 +1,199 @@
...
+import net.dries007.tfc.common.recipes.SimpleCraftMatrix;
+import net.dries007.tfc.common.recipes.KnappingRecipe;
+import net.dries007.tfc.util.Helpers;
+
+public class KnappingContainer extends ItemStackContainer implements IButtonHandler, IInventoryNoop
+{
+    public final SimpleCraftMatrix matrix;
+    private final int amountToConsume;
+    public final boolean usesDisabledTex;
+    private final boolean needsKnife;
```
alcatrazEscapee: This is really specific and not very useful to extensions. Rather than this, add a subclass of KnappingContainer used for leather knapping which overrides just `consumeIngredientStackAfterComplete`. That's better than declaring knives as an inherent part of knapping itself.

src/main/java/net/dries007/tfc/common/container/RunnableSlot.java
```diff
@@ -0,0 +1,39 @@
...
+package net.dries007.tfc.common.container;
+
+import javax.annotation.Nonnull;
+
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.item.ItemStack;
+import net.minecraftforge.items.IItemHandler;
+import net.minecraftforge.items.SlotItemHandler;
+
+public class RunnableSlot extends SlotItemHandler
```
alcatrazEscapee: This is unnecessary, the functionality exists in `CallbackSlot` already. Implement `ISlotCallback` on the container and override the methods `onTake` and `isItemValid`.

Also minding the fact that `RunnableSlot` is a horrible name in of itself.

src/main/java/net/dries007/tfc/common/recipes/RockKnappingRecipe.java
```diff
@@ -0,0 +1,78 @@
...
+import net.minecraft.util.JSONUtils;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.world.World;
+
+import net.dries007.tfc.common.TFCTags;
+import net.dries007.tfc.common.container.KnappingContainer;
+
+public class RockKnappingRecipe extends KnappingRecipe
+{
+    protected final Ingredient predicate;
```
alcatrazEscapee:
```suggestion
    protected final Ingredient ingredient;
```
for consistiency.

src/main/java/net/dries007/tfc/network/ScreenButtonPacket.java
```diff
@@ -0,0 +1,64 @@
...
+import net.dries007.tfc.common.container.IButtonHandler;
+
+public class ScreenButtonPacket
+{
+    private final int buttonID;
+    private final CompoundNBT extraNBT;
+
+    public ScreenButtonPacket(PacketBuffer buffer)
+    {
+        buttonID = buffer.readInt();
```
alcatrazEscapee: This could be replaced with a var int in all places

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -368,6 +368,17 @@ public static void damageCraftingItem(ItemStack stack, int amount)
...
+    @Nonnull
+    public static ItemStack consumeItem(ItemStack stack, int amount)
+    {
+        if (stack.getCount() <= amount)
+        {
+            return ItemStack.EMPTY;
+        }
+        stack.shrink(amount);
+        return stack;
+    }
```
alcatrazEscapee: - No need for `@Nonnull`
- This method shouldn't be necessary? the first bit is handled already by `shrink` and then all it's saving is `stack.shrink(amount); return stack` which seems a bit pedantic.

src/main/java/net/dries007/tfc/common/recipes/TypedRecipeSerializer.java
```diff
@@ -0,0 +1,15 @@
...
+ * You may obtain a copy of the Licence at:
+ * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ */
+
+package net.dries007.tfc.common.recipes;
+
+import net.minecraft.item.crafting.IRecipe;
+import net.minecraft.item.crafting.IRecipeType;
+
+public abstract class TypedRecipeSerializer<R extends IRecipe<?>> extends RecipeSerializer<R>
```
alcatrazEscapee, line 12: Is there a reason this class exists at all, and not just coded to `KnappingRecipe.Serializer` ?

src/main/java/net/dries007/tfc/common/recipes/SimpleCraftMatrix.java
```diff
@@ -0,0 +1,270 @@
...
+        short packed = buffer.readShort();
+        for (int i = 0; i < matrix.length; i++)
+            matrix[i] = (packed & (1 << i)) != 0;
+        return new SimpleCraftMatrix(outsideSlotRequired, matrix, width, height);
+    }
+
+    /**
+     * From {@link net.minecraft.item.crafting.ShapedRecipe}
+     */
+    public static String[] patternFromJson(JsonArray array)
```
alcatrazEscapee: Rather than deserializing `JsonArray` -> `String[]` -> `SimpleCraftMatrix` It'd be simpler just to have `fromJson` and return the matrix itself. i.e. a json object:
```json
{ "pattern": [...], "invert_empty_slots": true }
```

(as an aside I'm trying to think of better names for some of this stuff esp. since it's becoming serializable in json and yes I know I named them like two years ago yes they're crap)

src/main/java/net/dries007/tfc/common/recipes/SimpleCraftMatrix.java
```diff
@@ -0,0 +1,270 @@
...
+import com.google.gson.JsonSyntaxException;
+import net.minecraft.network.PacketBuffer;
+import net.minecraft.util.JSONUtils;
+
+import net.dries007.tfc.TerraFirmaCraft;
+
+/**
+ * A simple craft matrix for knapping / leather or clay working
+ *
+ * @author AlcatrazEscapee
```
alcatrazEscapee: Don't need an `@author` comment

src/main/java/net/dries007/tfc/common/recipes/SimpleCraftMatrix.java
```diff
@@ -0,0 +1,270 @@
...
+{
+    private static final int MAX_WIDTH = 5;
+    private static final int MAX_HEIGHT = 5;
+    private static final int MAX_AREA = MAX_WIDTH * MAX_HEIGHT;
+
+    private static void logMatrix(boolean[] matrix)
+    {
+        StringBuilder b = new StringBuilder();
+        for (boolean m : matrix) b.append(m ? "X" : " ");
+        TerraFirmaCraft.LOGGER.debug("Matrix: {" + b + "}");
```
alcatrazEscapee: Don't use the `TerraFirmaCraft` logger, have an instance per class if needed (or, because I doubt this method still needs to exist, yeet it)

src/main/java/net/dries007/tfc/common/recipes/SimpleCraftMatrix.java
```diff
@@ -0,0 +1,270 @@
...
+import net.minecraft.util.JSONUtils;
+
+import net.dries007.tfc.TerraFirmaCraft;
+
+/**
+ * A simple craft matrix for knapping / leather or clay working
+ *
+ * @author AlcatrazEscapee
+ */
+public class SimpleCraftMatrix
```
alcatrazEscapee: Thinking of a better name... how about `KnappingPattern` ?

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/client/screen/KnappingScreen.java
alcatrazEscapee: Whoops, I misread the purpose of this method.

src/main/java/net/dries007/tfc/common/container/KnappingContainer.java
eerussianguy: later

src/main/java/net/dries007/tfc/common/recipes/TypedRecipeSerializer.java
eerussianguy: Yes RockKnappingRecipe extends this instead of extending the default serializer. I had a lot of trouble with generic types and this was the compromise

src/main/java/net/dries007/tfc/common/recipes/SimpleCraftMatrix.java
eerussianguy: You added it yourself 😆

src/main/java/net/dries007/tfc/common/container/KnappingContainer.java
eerussianguy: Yes it breaks

Issue #1790, opened 2021-06-13
Barrels can be stacked like Russian nesting dolls


EminentMass:
Barrels can be stacked like Russian nesting dolls by placing them in the pour slot before sealing.
It could be changed so that it's impossible to seal a barrel while there is anything in the pour slot or just specifically other barrels.

TFC Version: 1.7.19.177

eerussianguy:
ad64e5c

PR #1807, opened 2021-07-31
Misc stuff & some bug fixes

### LINKED ISSUE #1792: Removing Ingots from a multi-block tall stack of Ingots may break the stack of Ingots.
**Describe the bug**
When the player removes an Ingot from a stack of Ingots and it reduces the number of block spaces taken up by the stack _(IE reduce size from 65 to 64, or 129 to 128 Ingots)_ the game will break/destroy the block-container for the stack of ingots at the point in which the player right-clicked it. It should break the top-most block-container (which no longer contains any Ingots and therefore should no longer exist), but it instead breaks the block the player right-clicked to remove the ingot - resulting in the game dumping 2+ stack of Ingot entities on the ground.

**To Reproduce**
Example 1:
1) Place down 65 Ingots of any type in a single pile/stack. (this creates a full block worth of the Ingots in a stack, and a single Ingot in the block space above it).
2) Right-click the bottom (full-block if Ingots) to remove an Ingot from the stack.
- What happens: You pick up 1 Ingot. The bottom stack breaks, dropping 64 Ingot entities on the ground.
- What I expect to happen: You pick up 1 Ingot.

Example 2:
1) Place down 129 Ingots of any type in a single pile/stack. (this creates two full blocks worth of the Ingots in a stack, and a single Ingot in the block space above those).
2) Right-click the middle block of ingots to remove an Ingot from the stack.
- What happens: You pick up 1 Ingot. The middle stack breaks, dropping 64 Ingot entities on top of the bottom stack of 64 Ingots.
- What I expect to happen: You pick up 1 Ingot.

Notes:
I technically noticed this while playing TNFC 1.12.2 several weeks ago, but this seemed like a TFC:TNG bug as it happened with the vanilla TFC ingots. I retested the bug today with just Forge mod loader and the latest release version of TFC:TNG I could find on the CurseForge website.

**Meta Info**
 - TFC Version: 1.7.19.177
 - Were any other mods included? - No
 - Can you reproduce it without these other mods? - Yes
 - If necessary, what other mods (including versions) MUST be present to experience the bug? - N/A
 
 ~LegionBreaker

Charnuz:
Fixes #1792, #1790, and #1586. Replaces the JEI hide scraping info with a new category, makes text from interacting with animals go to the actionbar, fires an event that allows control over the monster surface spawn denial, and a way to check if your plants are growing (for those people who dont use WAILA or the F3 menu)

The scraping category:
image


eerussianguy:
I am not really seeing the point of an event fired from inside an event. Especially since this is something that's fired for every spawn it adds overhead to the game. You can just listen to the spawn event with a lower or higher priority if that's something you want to control. You can un-cancel events just like you can cancel them.

eerussianguy:
Also, it's worth thinking about just using IHighlightHandler and drawing a colored box for the growing status just like in 1.7

Charnuz:
Ah yeah thats a good point, was hoping for a finer control over the mechanic but that'll do. Will look into that, should it be in addition or as a replacement?

eerussianguy:
I would do one or the other

eerussianguy:
Bump

Charnuz:
Sorry had my attention taken by other things, ill push an update today. I was thinking the highlighting would be better fit for if/when nutrients are added so we can add the highlighting for both at once

PR #1810, opened 2021-08-10
Remove predators and huntables before chunk regen

AnodeCathode:
Prevents slow overpopulation in chunks loaded every year. Particularly the chunks immediately surrounding a players base. Chunks loaded enough to trigger regen but not frequently enough to get protected chunk status slowly become overpopulated with predators.


eerussianguy (approved):
Seems fine

PR #1820, opened 2021-10-08
Implement boats, fix lagging porting issues around rendering

eerussianguy:


src/main/java/net/dries007/tfc/common/entities/TFCBoat.java
```diff
@@ -0,0 +1,76 @@
...
+import net.minecraft.world.entity.vehicle.Boat;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.level.Level;
+import net.minecraftforge.common.util.Constants;
+import net.minecraftforge.fmllegacy.network.NetworkHooks;
+
+import net.dries007.tfc.common.blocks.wood.Wood;
+import net.dries007.tfc.common.items.TFCItems;
+
+public class TFCBoat extends Boat
```
alcatrazEscapee, line 12: So, one thing is that this creates a lot of dependencies of `TFCBoat` on the `Wood` enum. Which is a bit of a pain for addons, if they want to add more wood types.

Is there a reason why we can't just have different entities? i.e. `acacia_boat`, `douglas_fir_boat`, etc.? Then that gets rid of 90% of `TFCBoatRenderer`, and most of the sync and save data of `TFCBoat`.

src/main/java/net/dries007/tfc/common/entities/TFCEntities.java
```diff
@@ -21,6 +21,7 @@
     public static final DeferredRegister<EntityType<?>> ENTITIES = DeferredRegister.create(ForgeRegistries.ENTITIES, MOD_ID);
 
     public static final RegistryObject<EntityType<TFCFallingBlockEntity>> FALLING_BLOCK = register("falling_block", EntityType.Builder.<TFCFallingBlockEntity>of(TFCFallingBlockEntity::new, MobCategory.MISC).sized(0.98f, 0.98f));
+    public static final RegistryObject<EntityType<TFCBoat>> BOAT = register("boat", EntityType.Builder.<TFCBoat>of(TFCBoat::new, MobCategory.MISC).sized(1.375F, 0.5625F).clientTrackingRange(10));
```
alcatrazEscapee: i.e. this could be a `Map<Wood, RegistryObject<EntityType<TFCBoat>>> BOATS`

src/main/java/net/dries007/tfc/common/items/TFCBoatItem.java
```diff
@@ -0,0 +1,98 @@
...
+import net.dries007.tfc.common.blocks.wood.Wood;
+import net.dries007.tfc.common.entities.TFCBoat;
+
+public class TFCBoatItem extends BoatItem
+{
+    private static final Predicate<Entity> ENTITY_PREDICATE = EntitySelector.NO_SPECTATORS.and(Entity::isPickable);
+
+    private final Wood wood;
+
+    public TFCBoatItem(Wood wood, Properties properties)
```
alcatrazEscapee: This is another problem, is that the constructor now only works for TFC variants. As opposed to, passing in a `Supplier<? extends EntityType<Boat>>` for example. Can also have a convenience constructor for internal use that accepts a `Wood` and then calls `this(...)` with `TFCEntities.BOATS.get(wood)`

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -90,6 +91,19 @@ public static ResourceLocation identifier(String name)
...
+     */
+    public static ModelLayerLocation modelIdentifier(String name, String part)
+    {
+        return new ModelLayerLocation(new ResourceLocation(MOD_ID, name), part);
+    }
+
+    public static ModelLayerLocation modelIdentifier(String name)
+    {
+        return modelIdentifier(name, "main");
+    }
```
alcatrazEscapee: These should both be in `ClientHelpers` since `ModelLayerLocation` is `@SideOnly(Dist.CLIENT)`

src/main/java/net/dries007/tfc/common/items/TFCBoatItem.java
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                }
+            }
+
+            if (hitresult.getType() == HitResult.Type.BLOCK)
+            {
+                // tfc start
+                TFCBoat boat = new TFCBoat(level, hitresult.getLocation().x, hitresult.getLocation().y, hitresult.getLocation().z);
+                boat.setWood(wood);
+                // tfc end
```
alcatrazEscapee: If this instead stores a field of a `Supplier<? extends EntityType<Boat>> boatType`, this can be replaced with

```java
Boat boat = boatType.get().create(level);
if (boat != null) {
    boat.setPos(hitresult.getLocation().x, hitresult.getLocation().y, hitresult.getLocation().z);
    // continue normally
```

src/main/java/net/dries007/tfc/common/items/TFCBoatItem.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private final Wood wood;
+
+    public TFCBoatItem(Wood wood, Properties properties)
+    {
+        super(Boat.Type.OAK, properties);
+        this.wood = wood;
+    }
+
+    public InteractionResultHolder<ItemStack> use(Level level, Player player, InteractionHand hand)
```
alcatrazEscapee, line 43: Missing a `@Override`. Would be useful to add a javadoc saying `copy pasta from {@link Boat#use}` just for future referencing of it.

alcatrazEscapee (changes requested):
If we have a entity type per wood type, instead of a `boat` entity with data defined wood, I think that both makes everything more extensible for addons, and removes any dependency on needing to pass a `Wood` instance around other than for initial registration.

The build seemed to fail because it couldn't download gradle... which is odd.

PR #1823, opened 2021-10-22
Finding Dory

eerussianguy:


resources/constants.py
```diff
@@ -443,9 +443,51 @@ def preset_vein(ore: str, vein_type: str, rocks: List[str], spoiler_ore: Optiona
...
+    'cod': spawner('tfc:cod', weight=10),
+    'pufferfish': spawner('tfc:pufferfish', max_count=2),
+    'tropical_fish': spawner('tfc:tropical_fish', weight=10, max_count=6),
+    'jellyfish': spawner('tfc:jellyfish', min_count=2, max_count=6)
+}
+
+"""OCEAN_CREATURES: Dict[str, Dict[str, Any]] = {
+    'orca': spawner('tfc:orca', min_count=1, max_count=3),
+    'dolphin': spawner('tfc:dolphin', min_count=1, max_count=3)
+}"""
```
alcatrazEscapee, line 478: These comments are todo or unused?

src/main/java/net/dries007/tfc/common/entities/Fauna.java
```diff
@@ -0,0 +1,107 @@
...
+    private final int chance;
+    private final int distanceBelowSeaLevel;
+    private final ClimateConfig climateConfig;
+    private final boolean solidGround;
+
+    @SuppressWarnings("unchecked")
+    public Fauna(ResourceLocation id, JsonObject json)
+    {
+        this.id = id;
+        entity = (EntityType<Mob>) ForgeRegistries.ENTITIES.getValue(new ResourceLocation(GsonHelper.getAsString(json, "entity")));
```
alcatrazEscapee:
```suggestion
        entity = (EntityType<Mob>) JsonHelpers.getRegistryEntry(json, "entity", ForgeRegistries.ENTITIES);
```

Similar for below. In addition, as a tip, you can call all vanilla `GsonHelper` methods through `JsonHelpers` as well, i.e. `JsonHelpers.getAsInt()` works fine.

src/main/java/net/dries007/tfc/common/entities/Fauna.java
```diff
@@ -0,0 +1,107 @@
...
+        return entity;
+    }
+
+    public SpawnPlacements.Type getType()
+    {
+        if (fluid != Fluids.EMPTY)//todo: serialize this...
+        {
+            return SpawnPlacements.Type.IN_WATER;
+        }
+        return SpawnPlacements.Type.ON_GROUND;
```
alcatrazEscapee: `JsonHelpers.getEnum` exists that should work for this.

src/main/java/net/dries007/tfc/common/entities/Fauna.java
```diff
@@ -0,0 +1,107 @@
...
+
+import com.mojang.serialization.JsonOps;
+import net.dries007.tfc.world.chunkdata.ChunkData;
+import net.dries007.tfc.world.decorator.ClimateConfig;
+
+/**
+ * A data driven way to make spawning conditions for animals player configurable.
+ * //todo: In 1.18, spawn predicates are set to change. We will have to adapt or throw this out.
+ */
+public class Fauna
```
alcatrazEscapee: So a couple things
- I worry this is a bit too specific to water spawning. i.e. `fluid`, `distanceBelowSeaLevel`, `solidGround`, etc. It might be better to have a type dispatch system?
- The other thing would be I'm not sure what you mean by spawn predicates - the goal here is to have additional end-user configurability of spawn checks, correct? What is vanilla changing about that?

src/main/java/net/dries007/tfc/common/entities/FaunaManager.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.google.gson.JsonObject;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.SpawnPlacements;
+import net.minecraft.world.level.levelgen.Heightmap;
+
+import net.dries007.tfc.mixin.accessor.SpawnPlacementsAccessor;
+import net.dries007.tfc.util.DataManager;
+
+public class FaunaManager extends DataManager<Fauna>
```
alcatrazEscapee: This looks like it should be a `RegisteredDataManager`, not a `DataManager`. Reason being:

- when you register a `SpawnPlacements`, you can register a "rules" instance that holds the `Supplier<Fauna>`, and resolves it at the correct time. Then, you don't need the accessor mixin, don't need the `postProcess`, etc. See `ClimateRanges`

```java
// this only needs to be called once, since the supplier will get reloaded automatically when resources reload
Supplier<Fauna> fishFauna = Fauna.MANAGER.register("fish");
SpawnPlacements.register(new FaunaSpawnPredicate(fishFauna));

class FaunaSpawnPredicate extends SpawnPlacements.SpawnPredicate<Mob> {
    FaunaSpawnPredicate(Supplier<Fauna> fauna) { ... }

    // do the predicate stuff, calling `fauna.get()`
}


```

src/main/java/net/dries007/tfc/common/entities/aquatic/IBucketableOceanFish.java
```diff
@@ -0,0 +1,14 @@
...
+ * Licensed under the EUPL, Version 1.2.
+ * You may obtain a copy of the Licence at:
+ * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ */
+
+package net.dries007.tfc.common.entities.aquatic;
+
+import net.minecraft.world.item.ItemStack;
+
+public interface IBucketableOceanFish
```
alcatrazEscapee: This apparently needs to be deleted?

src/main/java/net/dries007/tfc/common/entities/Fauna.java
```diff
@@ -0,0 +1,107 @@
...
+    private final int distanceBelowSeaLevel;
+    private final ClimateConfig climateConfig;
+    private final boolean solidGround;
+
+    @SuppressWarnings("unchecked")
+    public Fauna(ResourceLocation id, JsonObject json)
+    {
+        this.id = id;
+        entity = (EntityType<Mob>) ForgeRegistries.ENTITIES.getValue(new ResourceLocation(GsonHelper.getAsString(json, "entity")));
+        this.fluid = ForgeRegistries.FLUIDS.getValue(new ResourceLocation(GsonHelper.getAsString(json, "fluid", "minecraft:empty")));
```
alcatrazEscapee: Looking further at this - it seems salt water vs. fresh water is pretty hardcoded into our fishies. So having json exposed "fluid" wouldn't be particularly useful. So maybe it's best to just scrap this completely, and define either type-dispatched or just the spawn rule type?

src/main/java/net/dries007/tfc/mixin/accessor/SpawnPlacementsAccessor.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Map;
+
+import net.minecraft.world.entity.EntityType;
+import net.minecraft.world.entity.SpawnPlacements;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(SpawnPlacements.class)
+public interface SpawnPlacementsAccessor
```
alcatrazEscapee: With the tweak to fauna above, we don't need to re-register these so this should be fine to remove.

alcatrazEscapee (changes requested):
Looks quite good.

resources/constants.py
eerussianguy: todo, just makes my life easier later

resources/data.py
```diff
@@ -474,6 +481,58 @@ def block(block_type: str):
...
+def mob_loot(rm: ResourceManager, name_parts: utils.ResourceIdentifier, loot_pools: utils.Json):
+    res = utils.resource_location(rm.domain, name_parts)
+    rm.write((*rm.resource_dir, 'data', res.domain, 'loot_tables', 'entities', res.path), {
+        'type': 'minecraft:entity',
+        'pools': utils.loot_pool_list(loot_pools, 'entity')
+    })
+
+
+def climate_config(min_temp: Optional[float] = None, max_temp: Optional[float] = None, min_rain: Optional[float] = None, max_rain: Optional[float] = None, needs_forest: Optional[bool] = False, fuzzy: Optional[bool] = None) -> Dict[str, Any]:
+    return utils.del_none({
```
alcatrazEscapee: `del_none` shouldn't be necessary here I don't think, since it'll eventually be called by `utils.write`

PR #1826, opened 2021-11-10
Animal Husbandry

eerussianguy:
Draft because there is still a few things to be nailed down. Right now I am thinking of letting us get *just* pigs in the alpha, letting those get tested while we wait for the bulk animal adding to get in. Rather than bugfixing 20 entities at once


resources/data.py
```diff
@@ -534,6 +537,8 @@ def fauna(chance: int = None, distance_below_sea_level: int = None, climate: Dic
 
 
 def food_item(rm: ResourceManager, name_parts: utils.ResourceIdentifier, ingredient: utils.Json, category: Category, hunger: int, saturation: float, water: int, decay: float, fruit: Optional[float] = None, veg: Optional[float] = None, protein: Optional[float] = None, grain: Optional[float] = None, dairy: Optional[float] = None):
+    if type(ingredient) != utils.Json:
+        rm.item_tag('tfc:pig_food', ingredient)
```
alcatrazEscapee: This does not work how you think it does, and moreover, is just bad practice. Create a `tfc:foods` tag and let pig food reference that. Add all items explicitly to the food tag (a loop somewhere in the tags section).

you cannot `==` with types from the `typing` module. All the below return `True`:
```python
type('strings') != utils.Json
type({'dictionaries': 'with values'}) != utils.Json
type(None) != utils.Json
```
If you absolutely need to do runtime type checks, prefer `isinstance(obj, type)`, i.e. `isintance('thing', str)`. But you'll also find out that generic types cannot be type checked in this manner (since `utils.Json` means, "Any string, dictionary with string keys, or sequence of anything).

src/main/java/net/dries007/tfc/client/IngameOverlays.java
```diff
@@ -170,9 +173,58 @@ private static void renderHealthBar(LivingEntity entity, ForgeIngameGui gui, Pos
         gui.left_height += 10;
     }
 
+    public static void renderFamiliarity(RenderLivingEvent.Post<?, ?> event)
```
alcatrazEscapee: I think this should really be in `ClientForgeEventHandler`, since it's relation to everything else in `IngameOverlays` (everything using the explicit overlay system that is not rendered via events), doesn't really exist.

src/main/java/net/dries007/tfc/client/render/entity/TFCPigRenderer.java
```diff
@@ -0,0 +1,33 @@
...
+    private static final ResourceLocation OLD = ClientHelpers.animalTexture("pig_old");
+
+    public TFCPigRenderer(EntityRendererProvider.Context ctx)
+    {
+        super(ctx, new TFCPigModel(ctx.bakeLayer(ClientHelpers.modelIdentifier("pig"))), 0.7F);
+        //todo: saddle layer?
+    }
+
+    @Override
+    public ResourceLocation getTextureLocation(TFCPig pEntity)
```
alcatrazEscapee:
```suggestion
    public ResourceLocation getTextureLocation(TFCPig entity)
```
ewwwwwwww xP

src/main/java/net/dries007/tfc/common/entities/EntityHelpers.java
```diff
@@ -0,0 +1,133 @@
...
+import net.dries007.tfc.common.entities.land.TFCAnimalProperties;
+import net.dries007.tfc.common.entities.ai.TFCAvoidEntityGoal;
+import net.dries007.tfc.util.calendar.Calendars;
+
+public class EntityHelpers
+{
+    public static final EntityDataSerializer<Long> LONG_ENTITY_SERIALIZER = new EntityDataSerializer<>()
+    {
+        @Override
+        public void write(FriendlyByteBuf pBuffer, Long pValue)
```
alcatrazEscapee: Please no `p` parameters.

src/main/java/net/dries007/tfc/common/entities/EntityHelpers.java
```diff
@@ -0,0 +1,133 @@
...
+import net.minecraft.world.item.*;
+import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.world.level.Level;
+
+import net.dries007.tfc.common.entities.land.TFCAnimal;
+import net.dries007.tfc.common.entities.land.TFCAnimalProperties;
+import net.dries007.tfc.common.entities.ai.TFCAvoidEntityGoal;
+import net.dries007.tfc.util.calendar.Calendars;
+
+public class EntityHelpers
```
alcatrazEscapee: class can be `final`

src/main/java/net/dries007/tfc/common/entities/EntityHelpers.java
```diff
@@ -0,0 +1,133 @@
...
+{
+    public static final EntityDataSerializer<Long> LONG_ENTITY_SERIALIZER = new EntityDataSerializer<>()
```
alcatrazEscapee: I think the `ENTITY` is implicit, can just be `EntityHelpers.LONG_SERIALIZER`

src/main/java/net/dries007/tfc/common/entities/EntityHelpers.java
```diff
@@ -0,0 +1,133 @@
...
+            return pBuffer.readVarLong();
+        }
+        @Override
+        public Long copy(Long pValue)
+        {
+            return pValue;
+        }
+    };
+
+    public static void insertTFCAvoidGoal(PathfinderMob mob, GoalSelector selector, int priority)
```
alcatrazEscapee: The `TFC` is a bit redundant here (a while back we had a discussion on when to put "TFCFoo" and the gist was that unless it's shadowing a vanilla thing of the same name, leave it out). So, i.e. `replaceAvoidEntityGoal` makes a bit more sense.

src/main/java/net/dries007/tfc/common/entities/EntityHelpers.java
```diff
@@ -0,0 +1,133 @@
...
+    {
+        selector.getAvailableGoals().removeIf(wrapped -> wrapped.getGoal() instanceof AvoidEntityGoal);
+        selector.addGoal(priority, new TFCAvoidEntityGoal<>(mob, Player.class, 8.0F, 5.0D, 5.4D));
+    }
+
+    public static void insertCommonPreyAI(TFCAnimal animal, GoalSelector goalSelector)
```
alcatrazEscapee: This is also inconsistent since the previous method uses the `doSomethingGoal` but this is `doSomethingAI` when they look to be doing very similar tasks.

src/main/java/net/dries007/tfc/common/entities/land/Mammal.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private void setPregnantTime(long day)
+    {
+        entityData.set(PREGNANT_TIME, day);
+    }
+
+    @Nullable
+    @Override
+    public SpawnGroupData finalizeSpawn(ServerLevelAccessor pLevel, DifficultyInstance pDifficulty, MobSpawnType pReason, @Nullable SpawnGroupData pSpawnData, @Nullable CompoundTag pDataTag)
```
alcatrazEscapee: please no `p` 😩

src/main/java/net/dries007/tfc/common/entities/land/TFCAnimalProperties.java
```diff
@@ -0,0 +1,273 @@
...
+
+import net.minecraft.tags.Tag;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+
+import net.dries007.tfc.util.calendar.Calendars;
+import net.dries007.tfc.util.calendar.ICalendar;
+
+public interface TFCAnimalProperties
```
alcatrazEscapee, line 21: I don't like this name but can't think of anything better atm.

alcatrazEscapee (changes requested):
Some name nitpicking and some actual feedback.

One thing I think might be worth checking out, because I'm not exactly sure how it'll work at the moment. How does "uses" and "age" work for a pig? In the scenario where we kind of want "wild" animals to be untouched by time, per say, and any significant level of interaction with animals should cause time to move for them, what does that lead to?

For example (I'm putting "age" here in quotes because it's not a tick timer, it's a measure more akin to uses):
- raising familiarization increases "age"
- time spent loaded (i.e. ticking) while familiarized should raise age (?)
- time spent loaded (i.e. ticking) without any other conditions should raise age (?)
- getting pregnant or birthing children should raise age (?)

How do you envision this system working because it's not clear looking at this that anything "uses" a pig, per say.

eerussianguy:
#1839

PR #1827, opened 2021-11-14
misc bug fixes

eerussianguy:
- Meat and certain edible plants can be cooked
- Fixed emergent plants not generating (bad shared logic with other tall plants)
- Fixed floating plants not generating (redundant near_water placer, plus removed some redundant logic)
- Querns have block highlights again
- Log pile inventory should work now + some code cleanup there to new methods
- Closing the inventory after knapping properly updates the client
- Anvils and lamps no longer create black holes
- Fixed missing texture errors from crucibles
- steam particle now visible (I think low alpha values are getting culled for optimization reasons?)
- Theoretically fixed the handstone ejecting missingno particles

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/QuernBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/QuernBlockEntity.java
@@ -78,7 +78,7 @@ public class QuernBlockEntity extends InventoryBlockEntity<ItemStackHandler>
                     {
                         for (int i = 0; i < 15; i++)
                         {
-                            addParticle(level, pos, handstoneStack);
+                            addParticle(level, pos, undamagedHandstoneStack);
                         }
                     }
                 }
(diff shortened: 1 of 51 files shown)


src/main/java/net/dries007/tfc/common/blockentities/QuernBlockEntity.java
```diff
@@ -78,7 +78,7 @@ public static void clientTick(Level level, BlockPos pos, BlockState state, Quern
                     {
                         for (int i = 0; i < 15; i++)
                         {
-                            addParticle(level, pos, undamagedHandstoneStack);
+                            addParticle(level, pos, handstoneStack);
```
alcatrazEscapee: Is this intentional? the particle added will always be from the empty stack.

src/main/java/net/dries007/tfc/mixin/client/accessor/LevelRendererAccessor.java
```diff
@@ -0,0 +1,25 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Invoker;
+
+@Mixin(LevelRenderer.class)
+public interface LevelRendererAccessor
+{
+    @Invoker("renderShape")
+    static void invoke$renderShape(PoseStack stack, VertexConsumer consumer, VoxelShape shape, double x, double y, double z, float r, float g, float b, float a)
+    {
+
```
alcatrazEscapee:
```suggestion
throw new AssertionError();
```

alcatrazEscapee (changes requested):

PR #1849, opened 2022-01-08
Crops

eerussianguy:


src/main/java/net/dries007/tfc/util/climate/BiomeBasedClimateModel.java
```diff
@@ -17,10 +17,10 @@
 /**
  * A default climate model, for dimensions that are entirely biome determined (i.e. vanilla).
  */
-public class BiomeBasedClimateModel implements WorldGenClimateModel
+public class BiomeBasedClimateModel implements TimeInvariantClimateModel
 {
     @Override
-    public float getTemperature(LevelReader level, BlockPos pos, ChunkData data, long calendarTicks, int daysInMonth)
```
alcatrazEscapee, line 23: `getTemperature(LevelReader level, BlockPos pos)` should be overridden here, `getAverageTemperature` should not (it's already delegated to `getTemperature` from `TimeInvariantClimateModel`.

resources/world_gen.py
```diff
@@ -585,6 +585,29 @@ def generate(rm: ResourceManager):
...
+
+        res = utils.resource_location(rm.domain, name_parts)
+        patch_feature = res.join() + '_patch'
+        singular_feature = utils.resource_location(rm.domain, name_parts)
+        crop_features.append(patch_feature)
+
+        rm.configured_feature(patch_feature, 'minecraft:random_patch', {'tries': 4, 'xz_spread': 5, 'y_spread': 1, 'feature': singular_feature.join()})
+        rm.configured_feature(singular_feature, *feature)
+        rm.placed_feature(patch_feature, patch_feature)
+        rm.placed_feature(singular_feature, singular_feature, decorate_air_or_empty_fluid(), decorate_would_survive(name), decorate_heightmap('world_surface_wg'))
```
alcatrazEscapee: I think this is wrong - `decorate_heightmap` needs to be first here as otherwise those two won't pass. (if it does work that's in a way even more concerning)

resources/world_gen.py
```diff
@@ -585,6 +585,29 @@ def generate(rm: ResourceManager):
...
+
+    configured_placed_feature(rm, 'tfc:plant/wild_crops', 'minecraft:simple_random_selector', {'features': crop_features}, decorate_chance(40), decorate_square(), decorate_heightmap('world_surface_wg'), decorate_climate(min_rain=125, min_temp=-15), decorate_biome())
```
alcatrazEscapee: And this one should not have a heightmap decorator. (which explains why the above one actually works)

src/main/java/net/dries007/tfc/client/ClientEventHandler.java
```diff
@@ -336,13 +349,9 @@ private static boolean validateModelsAndTranslations()
...
         final TextureAtlasSprite missingParticle = missingModel.getParticleIcon();
 
-        final List<Block> missingModelErrors = blocksWithStateMatching(s -> s.getRenderShape() == RenderShape.MODEL && shaper.getBlockModel(s) == missingModel);
-        final List<Block> missingParticleErrors = blocksWithStateMatching(s -> !s.isAir() && shaper.getParticleIcon(s) == missingParticle);
         final List<Item> missingTranslationErrors = itemsWithStackMatching(s -> new TranslatableComponent(s.getDescriptionId()).getString().equals(s.getDescriptionId()));
 
-        return logValidationErrors("Blocks with missing models:", missingModelErrors, e -> LOGGER.error("  {}", e))
-            | logValidationErrors("Blocks with missing particles:", missingParticleErrors, e -> LOGGER.error("  {}", e))
-            | logValidationErrors("Items with missing translations:", missingTranslationErrors, e -> LOGGER.error("  {} ({})", e, e.getDescriptionId()));
+        return logValidationErrors("Items with missing translations:", missingTranslationErrors, e -> LOGGER.error("  {} ({})", e, e.getDescriptionId()));
```
alcatrazEscapee: Is this because they are broken somehow? I've spoken with Clay and found out that the translation one is not quite working and need to fix it but idk anything about the missing model or particle ones.

src/main/java/net/dries007/tfc/common/blockentities/FarmlandBlockEntity.java
```diff
@@ -9,23 +9,25 @@
 import java.util.List;
 import java.util.function.IntFunction;
 import javax.annotation.Nullable;
+import javax.swing.plaf.metal.MetalTheme;
```
alcatrazEscapee: eh?

src/main/java/net/dries007/tfc/common/blocks/crop/CropHelpers.java
```diff
@@ -0,0 +1,185 @@
...
+                });
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static void addNutrientParticles(ServerLevel level, BlockPos pos, Fertilizer fertilizer)
+    {
+        for (int i = 0; i < (int) (fertilizer.getNitrogen() * 10); i++)
```
alcatrazEscapee: Might be worth doing a clamp here? Since

- if someone adds a very high fertilizer we still want to limit particles to say 5 per color.
- if someone adds a fertilizer with 0.05, we still want to display particles (say min 1).

```java
final float n = fertilizer.getNitrogen();
n > 0 ? (int) Mth.clamp(n * 10, 5, 1) : 0
```

src/main/java/net/dries007/tfc/common/blocks/crop/DeadDoubleCropBlock.java
```diff
@@ -0,0 +1,69 @@
...
+        else
+        {
+            return belowState.is(this) && belowState.getValue(PART) == DoubleCropBlock.Part.BOTTOM;
+        }
+    }
+
+    @Override
+    public void addHoeOverlayInfo(Level level, BlockPos pos, BlockState state, List<Component> text, boolean isDebug)
+    {
+        super.addHoeOverlayInfo(level, state.getValue(PART) == DoubleCropBlock.Part.TOP ? pos.below() : pos, state, text, isDebug);
```
alcatrazEscapee: This technically breaks the contract of the given method, it should be passing in the value of `level.getBlockState(pos.below())` if it changes the position.

src/main/java/net/dries007/tfc/common/blocks/devices/TFCComposterBlock.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    protected void createBlockStateDefinition(StateDefinition.Builder<Block, BlockState> builder)
+    {
+        super.createBlockStateDefinition(builder.add(STAGE, TYPE));
+    }
+
+    @Override
+    public void setPlacedBy(Level level, BlockPos pos, BlockState state, @Nullable LivingEntity placer, ItemStack stack)
+    {
+        level.getBlockEntity(pos, TFCBlockEntities.COMPOSTER.get()).ifPresent(TickCounterBlockEntity::resetCounter);
```
alcatrazEscapee, line 110: Does this actually work? As per discussion on discord I thought `setPlacedBy` doesn't actually have the BE available?

PR #1888, opened 2022-03-07
More mob updates

eerussianguy:
Long long way to go. But this is a big deal so it needs to take it's own time.


src/main/java/net/dries007/tfc/ForgeEventHandler.java
```diff
@@ -637,6 +642,33 @@ public static void onPlayerTick(TickEvent.PlayerTickEvent event)
         {
             foodData.addThirst(TFCConfig.SERVER.thirstGainedFromDrinkingInTheRain.get().floatValue());
         }
+
+        // todo: PotionExpiryEvent is not called on the client due to bad design. So we have to fake it here. Needs a forge PR
```
alcatrazEscapee: Can we not send a packet from S -> C that just triggers the effect client side? Same as we do i.e. to detect when a player drinks (clientside event which sends C -> S to do logic)

resources/recipes.py
```diff
@@ -416,6 +428,9 @@ def plank(thing: str):
     alloy_recipe(rm, 'weak_blue_steel', 'weak_blue_steel', ('black_steel', 0.5, 0.55), ('steel', 0.2, 0.25), ('bismuth_bronze', 0.1, 0.15), ('sterling_silver', 0.1, 0.15))
     alloy_recipe(rm, 'weak_red_steel', 'weak_red_steel', ('black_steel', 0.5, 0.55), ('steel', 0.2, 0.25), ('brass', 0.1, 0.15), ('rose_gold', 0.1, 0.15))
 
+def disable_recipe(rm: ResourceManager, name_parts: utils.ResourceIdentifier):
```
alcatrazEscapee, line 431: We shouldn't need a custom recipe for this (and overriding with an explicit "noop" is clutter that should be avoided, since it pollutes the recipe lists). Does [this recipe removal](https://github.com/alcatrazEscapee/no-tree-punching/blob/1.18.x/src/main/resources/data/minecraft/recipes/acacia_planks.json) not work in 1.18?

src/main/java/net/dries007/tfc/common/entities/ai/BrainObjects.java
```diff
@@ -0,0 +1,60 @@
...
+import net.minecraft.world.entity.schedule.Schedule;
+import net.minecraft.world.entity.schedule.ScheduleBuilder;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+import com.mojang.serialization.Codec;
+import net.dries007.tfc.TerraFirmaCraft;
+
+public class BrainObjects
```
alcatrazEscapee: imo `TFCBrain` a. la. `TFCLoot`. Makes it clear it's TFC, not vanilla objects.

src/main/java/net/dries007/tfc/common/entities/ai/BrainObjects.java
```diff
@@ -0,0 +1,60 @@
...
+{
+    public static final DeferredRegister<Activity> ACTIVITIES = DeferredRegister.create(ForgeRegistries.ACTIVITIES, TerraFirmaCraft.MOD_ID);
+    public static final DeferredRegister<MemoryModuleType<?>> MEMORY_TYPES = DeferredRegister.create(ForgeRegistries.MEMORY_MODULE_TYPES, TerraFirmaCraft.MOD_ID);
+    public static final DeferredRegister<Schedule> SCHEDULES = DeferredRegister.create(ForgeRegistries.SCHEDULES, TerraFirmaCraft.MOD_ID);
+
+    public static final Activity HUNT = registerActivity("hunt");
+
+    public static final RegistryObject<Schedule> DIURNAL = registerSchedule("diurnal", newSchedule().changeActivityAt(0, HUNT).changeActivityAt(11000, Activity.REST)::build);
+    public static final RegistryObject<Schedule> NOCTURNAL = registerSchedule("nocturnal", newSchedule().changeActivityAt(0, Activity.REST).changeActivityAt(11000, HUNT)::build);
```
alcatrazEscapee: This looks convoluted. I *think* you should be able to rely on since Activity is sorted before Schedule, that you can unbox the activity registry object in the supplier. **But** only if you change it to invoke
```java
() -> newSchedule().foo()....build()
```
as that won't invoke `newSchedule()` until after.

I think this blows up in 1.18.2 anyway, because the actual instance needs to be created during register events (or so I've been told).

src/main/java/net/dries007/tfc/common/recipes/NoopRecipe.java
```diff
@@ -0,0 +1,94 @@
+/*
```
alcatrazEscapee: I don't think this is necessary, see earlier comment.

alcatrazEscapee (approved):

PR #1892, opened 2022-03-11
abstract all calls to tag matches

eerussianguy:


alcatrazEscapee (approved):
Fantastic work, and holy hell that's a lot more usages than I would've thought.

PR #1911, opened 2022-03-24
Random bug fixes

### LINKED ISSUE #1886: Quern handstone durability is not visible
With the removal of the quern gui, we have shown everything in world except the durability of the handstone. We should give the player some indication when the handstone is about to break.

### LINKED ISSUE #1910: Clay duplication bug [1.18.2 Alpha 2.0.12]
Clay is not used if stack is moved out of a slot by stacking with another partial stack of clay.

Reproduction:

1. Have two stacks of clay, such as both add to less than 64/full stack.
2. Enter knapping screen using one stack of clay.
3. Double click second stack to combine stacks.
4. Any valid clay recipe does not consume any resources, the inventory slot that original clay was in is locked during crafting.

https://user-images.githubusercontent.com/23376922/159898994-dff7a277-1f7a-40de-8780-884341ae0ce0.mp4

### LINKED ISSUE #1912: Heating issues [1.18.2 Alpha 2.0.12]
The gist of the issue is everything heats up and cools down too fast.

It's impossible to do vessel casting with pit kiln. the moment vessel is pulled out of a kiln it's solid metal.
Torches catch fire and food cooks in campfires instantly.
Forges get up to heat and cool down within seconds.

https://user-images.githubusercontent.com/23376922/160028635-385cbf5a-f70e-459e-b4df-1dbf3ab323ce.mp4

https://user-images.githubusercontent.com/23376922/160028684-251d296f-ee9d-4e53-9753-9d18c4e7ee68.mp4

https://user-images.githubusercontent.com/23376922/160028726-a2133501-1773-45c4-978e-d05fd60baf9e.mp4

Most heating should be scaled by factor of around 10 (between 8 and 15)

### LINKED ISSUE #1914: Log Piles placed against other Log Piles destroys one log [1.18.2 Alpha 2.0.13]
Log piles are created empty and waste one Log if placed against another log pile.

https://user-images.githubusercontent.com/23376922/160152106-e33f503d-dfbe-4aa3-8cab-af40d1a5fc06.mp4

There are also various issues immediately after loading into the game, among others: stack size for logs is 32, sticks cannot be placed in campfires. After reloading save by quitting to menu and entering a world again, those issues are resolved

eerussianguy:


src/main/java/net/dries007/tfc/common/container/ItemStackContainer.java
```diff
@@ -18,6 +20,8 @@
  */
 public class ItemStackContainer extends Container
 {
+    private static final List<ClickType> ILLEGAL_ITEM_CLICKS = List.of(ClickType.QUICK_MOVE, ClickType.PICKUP, ClickType.THROW, ClickType.SWAP);
```
alcatrazEscapee: If you are only going to do `.contains`, technically a `Set` is more efficient here. If `ClickType` is an `enum`, a `EnumSet` would also work

src/main/java/net/dries007/tfc/common/container/KnappingContainer.java
```diff
@@ -169,12 +169,7 @@ public boolean isItemValid(int slot, ItemStack stack)
     @Override
     protected boolean moveStack(ItemStack stack, int slotIndex)
     {
-        return switch (typeOf(slotIndex))
-            {
-                case CONTAINER -> !moveItemStackTo(stack, containerSlots, containerSlots + 36, true); // Hotbar first
-                case HOTBAR -> !moveItemStackTo(stack, containerSlots, containerSlots + 27, false);
-                case MAIN_INVENTORY -> !moveItemStackTo(stack, containerSlots + 27, containerSlots + 36, false);
-            };
+        return true; // no shift clicking
```
alcatrazEscapee: Can this not be modified to ignore the specified slot only? The slot ranges here are specific slots to transfer to. Ignoring shift clicks entirely seems a bit slapdash (and also a bit of a "remove the bug in the feature by removing the feature" fix)

src/main/java/net/dries007/tfc/common/container/ItemStackContainer.java
Rongmario: For future reference: If its just a tiny `List`, the performance of a `List#contains` matches or can even beat out a `Set#contains`. In this case `EnumSet` is the best option since `EnumSet#contains` just checks `(bit_vector & (1 << enum_ordinality)) != 0`

eerussianguy:
also, for changelog -- added missing wood bucket recipe

Issue #1917, opened 2022-03-27
Floating plants can render underwater


Charnuz:
Happens when the water underneath the plant adjusts its height to meet water coming from above. Vanilla lily pad for comparison
![image](https://user-images.githubusercontent.com/16440978/160277889-df290da1-1162-4897-86fc-8692142b896b.png)

eerussianguy:
Looked into this and there's no good way of fixing this. We cannot override getOffset anymore as that method moved to BlockStateBase, and i'd rather not do an expensive check anyway

eerussianguy:
Model needs to be higher.

PR #1932, opened 2022-04-11
Floating features model and blockstate tweaks 

Traister101:
- Changed template height from 1 unit to .25 units, this is the same as the vanilla lily pad and had no issues with the water rendering on top in my testing like seen in #1917. There is also a new template model template_floating_tinted_flower same as template_floating_tinted but allows for flowers
- Pistia height has been changed to it's lowest face at .25 units and it's highest face .5 units higher at .75 maintaining the same distance from before
- Enabled the yellow lily pad model and removed the imposter that took it's place
- Fiddled around with model weights of water lilys and lotus's


eerussianguy (approved):

eerussianguy:
This is good important work, thanks for this

src/main/resources/assets/tfc/models/block/plant/duckweed.json
```diff
@@ -1,7 +1,6 @@
 {
```
alcatrazEscapee, line 1: This is an error: just removing the autogeneration marker does not mean that file stops getting autogenerated. This should've been fixed in resource generation, otherwise if someone's being careless (like I just almost was), this would've been overwritten by the next resource generation)

cc @eerussianguy

PR #1938, opened 2022-04-18
fix server crash, implement game test framework

eerussianguy:


src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -781,6 +781,10 @@ public static boolean detectAssertionsEnabled()
      */
     public static boolean detectTestSourcesPresent()
```
alcatrazEscapee: This needs to be two seperate methods, as the name does not imply what it is doing

- `Helpers.detectBootstrapEnvironment()` - Both checks. This is an environment where we aren't launched through a normal server, and so stuff like the hack in `TFCLayers` needs to do something different, to avoid crashing. That's the current environment in which `/gradlew test` is ran through.
- `Helpers.detectTestSourcesPresent()` - Just if test sources are present, that's all.

.github/workflows/build.yml
```diff
@@ -48,6 +48,13 @@ jobs:
       env:
         VERSION: ${{ steps.version.outputs.result }}
         GRADLE_BUILD_ACTION_CACHE_DEBUG_ENABLED: true
+    - name: Game Test
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: runGameTestServer
+      env:
+        VERSION: indev-test
+        GRADLE_BUILD_ACTION_CACHE_DEBUG_ENABLED: true
```
alcatrazEscapee, line 57: Can you not append arguments to the previous action? This looks like it will repeat the build which would be a waste.

build.gradle
```diff
@@ -44,6 +44,14 @@ def envVersion = System.getenv('VERSION')
 version = envVersion == null ? 'indev' : envVersion
 java.toolchain.languageVersion = JavaLanguageVersion.of(17)
 
+
+sourceSets {
+    test
+    main {
+        runtimeClasspath += sourceSets.test.output
+    }
```
alcatrazEscapee, line 52: I'm very concerned by this, does this not dump the entire test outputs onto the class path in non test environments? i.e. `runClient` will have test sources present (no bueno)

PR #1945, opened 2022-04-25
Anvils

alcatrazEscapee:
Draft while various kinks and the rest of the functionality are worked out

Still missing / todo:
- [x] A lot of anvil plan / anvil screen functionality
- [x] Some specifics about the implementation of recipes, i.e. welding + anviling
- [x] Replacement for anvil skill bonuses (it's half there)
- [x] Data generation for recipes
- [x] Rendering BESR stuff (displaying items on anvil)


resources/constants.py
```diff
@@ -856,6 +862,11 @@ def lang_enum(name: str, values: Sequence[str]) -> Dict[str, str]:
     'tfc.enum.nutrient.vegetables': 'Vegetables',
     'tfc.enum.nutrient.protein': 'Protein',
     'tfc.enum.nutrient.dairy': 'Dairy',
+    'tfc.enum.forgingbonus.none': 'No Forging Bonus',
+    'tfc.enum.forgingbonus.poorly_forged': 'Poorly Forged',
+    'tfc.enum.forgingbonus.well_forged': 'Well Forged',
+    'tfc.enum.forgingbonus.expertly_forged': 'Expertly Forged',
+    'tfc.enum.forgingbonus.perfectly_forged': 'Perfectly Forged!',
```
eerussianguy, line 870: my idea:
- Show nothing if no bonus
- some adverbs you may prefer: flawlessly, exquisitely, impeccably, superbly

resources/recipes.py
```diff
@@ -1,11 +1,47 @@
...
+from enum import Enum
+
 from mcresources import ResourceManager, RecipeContext, utils
-from mcresources.type_definitions import Json
+from mcresources.type_definitions import ResourceIdentifier, Json
+
 from constants import *
 
 
+class Rules(Enum):
```
eerussianguy, line 12: not in constants?

src/main/java/net/dries007/tfc/common/blockentities/AnvilBlockEntity.java
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    @Override
+    public boolean isItemValid(int slot, ItemStack stack)
+    {
+        return switch (slot)
+            {
+                case SLOT_INPUT_MAIN, SLOT_INPUT_SECOND -> true;
+                case SLOT_HAMMER -> Helpers.isItem(stack, TFCTags.Items.HAMMERS);
+                case SLOT_CATALYST -> Helpers.isItem(stack, TFCTags.Items.FLUX);
```
eerussianguy, line 110: so I assume we're not gonna have a special catalyst relation in the recipe?

src/main/java/net/dries007/tfc/common/blocks/rock/Rock.java
```diff
@@ -80,8 +80,8 @@ public String getSerializedName()
 
     public enum BlockType implements StringRepresentable
     {
-        RAW((rock, self) -> new Block(Block.Properties.of(Material.STONE).sound(SoundType.STONE).strength(2 + rock.getCategory().getHardness(), 10).requiresCorrectToolForDrops()), true),
-        HARDENED((rock, self) -> new Block(Block.Properties.of(Material.STONE).sound(SoundType.STONE).strength(2.25f + rock.getCategory().getHardness(), 10).requiresCorrectToolForDrops()), false),
+        RAW((rock, self) -> RockConvertableToAnvilBlock.createForIgneousOnly(Block.Properties.of(Material.STONE).sound(SoundType.STONE).strength(2 + rock.getCategory().getHardness(), 10).requiresCorrectToolForDrops(), rock), true),
```
eerussianguy, line 83: extremely descriptive class name

src/main/java/net/dries007/tfc/common/blockentities/AnvilBlockEntity.java
alcatrazEscapee: I wasn't going to, no, as it overcomplicates the slot eligibility, which affects stuff like shift clicking inputs. If it becomes a requested feature we can look into it, but there are notable downsides.

resources/recipes.py
alcatrazEscapee: Like the enums in `data` that are used only here, I think it makes sense here as it's only used in the anvil recipe function.

resources/constants.py
alcatrazEscapee: > Show nothing if no bonus

It does, the enum translation key is included for completeness, since `Helpers.translateEnum(ForgingBonus)` could technically still be called elsewhere.

src/main/java/net/dries007/tfc/common/blocks/rock/Rock.java
alcatrazEscapee: :D

PR #1958, opened 2022-05-19
Predator Tweaks

Therighthon:
- Added Sounds for Predators
- Removed persistence required
- Walk animations faster/slower based on walking speed


src/main/java/net/dries007/tfc/client/model/entity/BearModel.java
```diff
@@ -162,7 +169,9 @@ else if (predator.isInWater() || (blockStateBelow.getFriction(predator.level, bl
                 body.y = 22f;
                 SWIM.tick(parts, predator.walkProgress);
             }
-            else if (predator.walkProgress > 0)
+
+            else if (predator.walkProgress > 0 || predator.isMoving())
+
```
eerussianguy: This isn't really our code style, was this accidentally done by intelliJ?

eerussianguy (changes requested):

eerussianguy (approved):

PR #1960, opened 2022-05-21
upgrade and enable fishing

eerussianguy:


src/main/java/net/dries007/tfc/client/ClientForgeEventHandler.java
```diff
@@ -189,6 +190,12 @@ public static void onItemTooltip(ItemTooltipEvent event)
             stack.getCapability(HeatCapability.CAPABILITY).ifPresent(cap -> cap.addTooltipInfo(stack, text));
             stack.getCapability(EggCapability.CAPABILITY).ifPresent(cap -> cap.addTooltipInfo(text));
 
+            ItemStack bait = TFCFishingRodItem.getBaitItem(stack);
+            if (!bait.isEmpty())
+            {
+                text.add(new TranslatableComponent("tfc.tooltip.fishing.bait").append(bait.getHoverName()));
+            }
```
alcatrazEscapee: This should be in `addInformation()` since it's specific to `TFCFishingRodItem`s

src/main/java/net/dries007/tfc/common/recipes/AddBaitRecipe.java
```diff
@@ -0,0 +1,92 @@
+/*
```
alcatrazEscapee: This feels very awful, yet I see why it's nessecary. What I would *like* to do here is simply add an item stack modifier for "add_bait_to_rod", throw it in a advanced shapeless recipe, and call it a day.

but there would be no way to access the bait.

so here's my very hacky yet genius idea, based on the idea of `ForgeHooks.getCraftingPlayer()`

- In `AdvancedShapedRecipe` and `AdvancedShapelessRecipe`, just before the call to `assemble`, call `RecipeHelpers.setCraftingContainer()`:
- This sets a `ThreadLocal<CraftingContainer>`
- Then in a custom `AttachBaitToRodModifier` implementation, call `RecipeHelpers.getCraftingContainer()`
- Use this to obtain the "bait" item, just by searching for items that match a tag. (the rod item will be given as the input)
- Profit.

```java
private static final ThreadLocal<CraftingContainer> CRAFTING_CONTAINER = ThreadLocal.withInitial(() -> null);

/**
 * @return The crafting container of the player currently crafting an recipe which might need it. Provided by {@link AdvancedShapedRecipe} and {@link AdvancedShapelessRecipe}. Used by {@link AddBaitToRodModifier}.
 */
@Nullable
public static CraftingContainer getCraftingContainer()
{
	return CRAFTING_CONTAINER.get();
}

public static void setCraftingContainer(CraftingContainer container)
{
	CRAFTING_CONTAINER.set(container);
}
```

src/main/java/net/dries007/tfc/common/recipes/AdvancedShapelessRecipe.java
```diff
@@ -18,6 +18,7 @@
 import net.minecraft.world.item.crafting.ShapelessRecipe;
```
alcatrazEscapee, line 18: These modifications should all be unnecessary, see above comment.

alcatrazEscapee (changes requested):

alcatrazEscapee (approved):
Those item stack modifiers are *so clean* :D

PR #1961, opened 2022-05-25
Blast Furnace, Metal Sheets, Ingot Piles

alcatrazEscapee:
Still todo:
- [x] Fix bugs and inconsistencies with sheet pile placement in `InteractionManager`
- [x] Fix bellows interaction result weirdness
- [x] Implement blast furnace screen and UI
- [x] Implement calendar fast forwarding for blast furnace
- [x] Test blast furnace operation


src/main/java/net/dries007/tfc/client/render/blockentity/SheetPileBlockEntityRenderer.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class SheetPileBlockEntityRenderer implements BlockEntityRenderer<SheetPileBlockEntity>
+{
+    @Override
+    public void render(SheetPileBlockEntity pile, float partialTick, PoseStack poseStack, MultiBufferSource buffer, int packedLight, int packedOverlay)
+    {
+        final BlockState state = pile.getBlockState();
+        if (state.getBlock() instanceof DirectionPropertyBlock)
+        {
+            @SuppressWarnings("deprecation") final Function<ResourceLocation, TextureAtlasSprite> textureAtlas = Minecraft.getInstance().getTextureAtlas(TextureAtlas.LOCATION_BLOCKS);
```
eerussianguy: maybe it's time we copy this RL to a place in RenderHelpers where its not deprecated?

src/main/java/net/dries007/tfc/common/blockentities/BlastFurnaceBlockEntity.java
```diff
@@ -0,0 +1,579 @@
...
+            Helpers.consumeItemsFromEntitiesIndividually(foundFuels, totalFuelCapacity, fuelStacks::add);
+
+            markForSync();
+        }
+    }
+
+    /**
+     * Sets a molten block inside the bloomery structure. If there is nothing in the bloomery, attempts to delete any molten blocks left over.
+     */
+    private void updateMoltenBlock(boolean cooking)
```
eerussianguy: can this be made common to the bloomery? The bloomery slag layers have had reports of not working quite correctly.

src/main/java/net/dries007/tfc/common/blocks/devices/IngotPileBlock.java
```diff
@@ -0,0 +1,152 @@
...
+        registerDefaultState(getStateDefinition().any().setValue(COUNT, 1));
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public BlockState updateShape(BlockState state, Direction direction, BlockState neighborState, LevelAccessor level, BlockPos currentPos, BlockPos neighborPos)
+    {
+        if (direction == Direction.DOWN && !neighborState.isFaceSturdy(level, neighborPos, direction.getOpposite()) && !Helpers.isBlock(neighborState, this))
+        {
+            level.scheduleTick(currentPos, this, 0);
```
eerussianguy: does 0 tick delay work? not aware of its use in vanilla

src/main/java/net/dries007/tfc/common/blocks/devices/SheetPileBlock.java
```diff
@@ -0,0 +1,215 @@
...
+     */
+    @Nullable
+    public static Direction getTargetedFace(Level level, BlockState state, Player player)
+    {
+        final BlockHitResult result = Helpers.rayTracePlayer(level, player, ClipContext.Fluid.NONE);
+        if (result.getType() == HitResult.Type.BLOCK)
+        {
+            final Vec3 hit = result.getLocation();
+            @Nullable Direction firstDirection = null;
+            for (Map.Entry<BooleanProperty, VoxelShape> entry : SHAPES.entrySet())
```
eerussianguy, line 105: more of a comment than a review but I wonder if this is possible to do with comparing the difference between the hit vector and the block position to Direction's coordinate step field (ie, North and 0, 0, 1)

src/main/java/net/dries007/tfc/util/InteractionManager.java
```diff
@@ -305,6 +309,144 @@ else if (!level.getBlockState(relativePos.below()).isAir())
...
+                if (blockContext.replacingClickedOnBlock())
+                {
+                    // todo: this causes weird behavior, how to handle properly?
+                    return InteractionResult.FAIL;
+                }
+
+                // Sheets behave differently than ingots, because we need to check the targeted face if it's empty or not
+                // We assume immediately that we want to target the relative pos and state
+                if (Helpers.isBlock(relativeState, TFCBlocks.SHEET_PILE.get()))
+                {
```
eerussianguy, line 424: IHighlightHandler could help with players targeting

src/main/java/net/dries007/tfc/util/InteractionManager.java
```diff
@@ -305,6 +309,144 @@ else if (!level.getBlockState(relativePos.below()).isAir())
...
+                final Direction sheetFace = clickedFace.getOpposite(); // i.e. place on DOWN
+                final BlockPos relativePos = context.getClickedPos().relative(clickedFace);
+                final BlockState relativeState = level.getBlockState(relativePos);
+
+                final BlockPlaceContext blockContext = new BlockPlaceContext(context);
+                final BooleanProperty property = DirectionPropertyBlock.getProperty(sheetFace);
+
+                if (blockContext.replacingClickedOnBlock())
+                {
+                    // todo: this causes weird behavior, how to handle properly?
```
eerussianguy: vanilla solution -- override getStateForPlacement to return null if you're holding a sheet. This allows the logic to pass

src/main/java/net/dries007/tfc/common/container/BlastFurnaceContainer.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private BlastFurnaceContainer(int windowId, BlastFurnaceBlockEntity blockEntity)
+    {
+        super(TFCContainerTypes.BLAST_FURNACE.get(), windowId, blockEntity);
+    }
+
+    @Override
+    protected boolean moveStack(ItemStack stack, int slotIndex)
+    {
+        // todo: impl
```
eerussianguy: people will complain immediately if this doesnt get done

src/main/java/net/dries007/tfc/common/blockentities/BlastFurnaceBlockEntity.java
eerussianguy: This could also be a gametest

src/main/java/net/dries007/tfc/common/blocks/devices/IngotPileBlock.java
alcatrazEscapee: This is interesting, because
- iirc it did work (I'll test again to make sure)
- I can't find anywhere in vanilla that does 0 ticks
- I can find other places in TFC that do 0 ticks.

src/main/java/net/dries007/tfc/util/InteractionManager.java
alcatrazEscapee: That's not possible as we're not controlling the item. It's not about the logic passing, it's about figuring out exactly how to place correctly (there's still some kinks I am trying to work out there)

src/main/java/net/dries007/tfc/common/blocks/devices/SheetPileBlock.java
alcatrazEscapee: possibly? but that feels like it would be more unreliable, the voxel shape method seems to work pretty good so far.

src/main/java/net/dries007/tfc/util/InteractionManager.java
alcatrazEscapee: While true, this isn't the case where they're needed (since this is just placing a block targeting an unrelated one), it's already pretty visible when breaking, and I'm lazy so I think I'll throw a todo and leave it at that

PR #1962, opened 2022-05-27
Grass Colormaps and Aesthetic Tweaks

Therighthon:
Made tall grass and grass blocks use separate colormaps
Updated Grass colormaps to have more yellow
Moved water plants/hanging vines to the foliage color map so they don't turn yellow
Added an actual model for water canna
Made clay dirt blocks not have randomly oriented top textures
Banished reindeer lichen to the arctic


src/main/java/net/dries007/tfc/client/TFCColors.java
```diff
@@ -7,6 +7,9 @@
 package net.dries007.tfc.client;
 
 import java.util.function.ToIntFunction;
+
+import org.apache.commons.lang3.RandomUtils;
+
```
alcatrazEscapee, line 12: This looks unused.

src/main/java/net/dries007/tfc/common/blocks/plant/Plant.java
```diff
@@ -16,6 +16,7 @@
 import net.dries007.tfc.common.blocks.ExtendedProperties;
 import org.jetbrains.annotations.Nullable;
 
+import net.minecraft.client.color.block.BlockColor;
```
alcatrazEscapee, line 19: This also looks unused.

alcatrazEscapee (commented):
I think you did a pretty great job here at balancing colors, just a few cleanups otherwise good.

PR #1963, opened 2022-05-28
javelins

eerussianguy:


src/main/java/net/dries007/tfc/common/items/JavelinItem.java
```diff
@@ -6,15 +6,106 @@
...
+ */
 public class JavelinItem extends SwordItem
 {
-    public JavelinItem(Tier tier, float attackDamage, float attackSpeed, Properties properties)
+    private final ResourceLocation textureLocation;
+
+    public JavelinItem(Tier tier, float attackDamage, float attackSpeed, Properties properties, String name)
     {
         super(tier, (int) ToolItem.calculateVanillaAttackDamage(attackDamage, tier), attackSpeed, properties);
+        this.textureLocation = Helpers.identifier("textures/entity/projectiles/" + name + "_javelin.png");
```
alcatrazEscapee: Should have two constructors: one which takes a string and passes `Helpers.identifier("textures/entity/projectile/" + name + "_javelin.png")` to the second constructor which takes a `ResourceLocation`

alcatrazEscapee (approved):

PR #1979, opened 2022-06-01
bugfixes

### LINKED ISSUE #1967: [1.18.2] Daub Mud Recipe Overridden
In 1.18.2-2.0.25, the recipe for daub that uses straw and mud doesn't work; any positioning of straw and mud in the 2x2 player crafting grid produces mud bricks instead.

### LINKED ISSUE #1968: [1.18.2] Fruit tree saplings not dropping from branch elbows when broken with axe
In 1.18.2-2.0.25, I haven't been able to get fruit tree saplings when breaking the "elbow" branches with an axe. I've used both stone and metal axes, and tested both natural and grown fruit trees (though haven't been able to test a metal axe on a natural fruit tree), all in survival mode.

### LINKED ISSUE #1970: [1.18.2] Stone supported by non-stone blocks
in 1.18.2-2.0.25, blocks of raw stone consider non-solid blocks (leaves, tall grass, etc.) as supports when checking to see if they should drop themselves as an item.

### LINKED ISSUE #1975: [1.18.2] Wild 2 tall plants do not get harvested if only the top part is broken (in survival only)
Here is an example of a partially broken Wild Sugarcane plant. This is also the case for all 2 block tall food plants.

### LINKED ISSUE #1976: [1.18.2] Trees overwritten by bushes
in 1.18.2-2.0.25, it's possible for the leaf blocks of the small bushes to overwrite the trunks of trees.

.

eerussianguy:
fix fruit trees not dropping saplings, closes #1968, fix mud bricks not craftable, closes #1967, fix wild and dead tall crops not breaking both blocks, closes #1975, fix bushes destroying things, closes #1976, fix stone not popping off when touching a transparent block, closes #1970, make composters reset counter


alcatrazEscapee (approved):
LGTM

PR #1986, opened 2022-06-14
1.18.x subtitles

Charnuz:
Mainly standardizes subtitles & sound events


src/main/java/net/dries007/tfc/config/animals/MammalConfig.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package net.dries007.tfc.config.animals;
+
+import java.util.function.Function;
+
+import net.minecraftforge.common.ForgeConfigSpec;
+
+public record MammalConfig(AnimalConfig inner, ForgeConfigSpec.IntValue gestationDays, ForgeConfigSpec.IntValue childCount)
+{
```
alcatrazEscapee, line 14: You could add one constructor to this,
```java
public MamalConfig(AnimalConfig animal, ...) {
    this(animal.familiarityCap, animal.adulthoodDays, ... );
}
```

which then removes the need for having these nested `.inner` accesses, but still only keeps the config declarations in one place.

alcatrazEscapee (changes requested):

alcatrazEscapee (approved):

PR #1997, opened 2022-06-19
Horses

eerussianguy:
Not tested or complete but here take a look


src/main/java/net/dries007/tfc/common/entities/livestock/horse/TFCChestedHorse.java
```diff
@@ -79,29 +91,191 @@ public TFCChestedHorse(EntityType<? extends TFCChestedHorse> type, Level level,
...
+                    return false;
+                }
+            }
+        } : super.getSlot(slot);
     }
 
     @Override
     protected SoundEvent getEatingSound()
     {
+        super.getEatingSound();
```
alcatrazEscapee, line 241: Are these `super` calls meant to be here?

PR #2001, opened 2022-06-23
Tag-ify anvil for mod compat

Nyonyix:
Replaced the item requirements for the anvil recipes with tags for future mod compatibility. I plan to do the rest of the recipes that need it, This is just a sample to see if I am doing this right.


resources/recipes.py
```diff
@@ -664,45 +664,48 @@ def chisel_stair_slab(name: str, ingredient: str):
...
+            anvil_recipe(rm, '%s_unfinished_greaves' % metal, item_tag('forge', 'double_sheet'), item('unfinished_greaves'), metal_data.tier, Rules.bend_any, Rules.draw_any, Rules.hit_any)
+            anvil_recipe(rm, '%s_unfinished_boots' % metal, item_tag('forge', 'sheet'), item('unfinished_boots'), metal_data.tier, Rules.bend_last, Rules.bend_second_last, Rules.shrink_third_last)
 
         if 'utility' in metal_data.types:
-            anvil_recipe(rm, '%s_trapdoor' % metal, item('sheet'), item('trapdoor'), metal_data.tier, Rules.bend_last, Rules.draw_second_last, Rules.draw_third_last)
-            anvil_recipe(rm, '%s_lamp' % metal, item('ingot'), item('lamp'), metal_data.tier, Rules.bend_last, Rules.bend_second_last, Rules.draw_third_last)
-            anvil_recipe(rm, '%s_chain' % metal, item('ingot'), '16 tfc:metal/chain/%s' % metal, metal_data.tier, Rules.hit_any, Rules.hit_any, Rules.draw_last)
+            anvil_recipe(rm, '%s_trapdoor' % metal, item_tag("forge", "sheets"), item('trapdoor'), metal_data.tier, Rules.bend_last, Rules.draw_second_last, Rules.draw_third_last)
```
Claycorp: `item_tag` already appends the `s` to make these plural. Looking at the changed files causes all these to have the tag of `sheetss` instead of `sheets` as intended.

src/main/resources/data/tfc/recipes/anvil/bismuth_bronze_trapdoor.json
```diff
@@ -2,7 +2,7 @@
   "__comment__": "This file was automatically created by mcresources",
   "type": "tfc:anvil",
   "input": {
-    "item": "tfc:metal/sheet/bismuth_bronze"
+    "tag": "forge:sheetss/bismuth_bronze"
```
Claycorp: This is an example of the result of the `sheets` issue.

resources/recipes.py
```diff
@@ -674,9 +674,9 @@ def item_tag(namespace: str, _variant: str) -> str:
             anvil_recipe(rm, '%s_sheet' % metal, item_tag('forge', 'double_ingot'), item('sheet'), metal_data.tier, Rules.hit_last, Rules.hit_second_last, Rules.hit_third_last)
             anvil_recipe(rm, '%s_rod' % metal, item_tag('forge', 'ingot'), '2 tfc:metal/rod/%s' % metal, metal_data.tier, Rules.bend_last, Rules.draw_second_last, Rules.draw_third_last)
 
-        # Toolss
+        # Tools
         if 'tool' in metal_data.types:
-            anvil_recipe(rm, '%s_tuyere' % metal, '#forge:double_sheets/%s' % metal, 'tfc:metal/tuyere/%s' % metal, metal_data.tier, Rules.bend_last, Rules.bend_second_last)
+            anvil_recipe(rm, '%s_tuyere' % metal, '#forge:double_sheet/%s' % metal, 'tfc:metal/tuyere/%s' % metal, metal_data.tier, Rules.bend_last, Rules.bend_second_last)
```
Claycorp: This was correct then changed to be incorrect as it is a direct reference to the tag instead of using `item_tag`, Thus the `s` is required and isn't concatenated at generation time like with `item_tag`.

As per the forge docs and vanilla convention all tags should be plural. It also separates the tag used here compared to all other sheet tags as they are `sheets`.

>  In particular, item and block groupings are plural instead of singular (e.g. minecraft:logs, minecraft:saplings).


That brings the question of why this uses a direct reference instead of `item-tag` like everything else.

src/main/resources/data/minecraft/tags/items/brass.json
```diff
@@ -0,0 +1,5 @@
+ {
+    "values": [
+        "tfc:metal/sheet/brass"
```
Claycorp: `brass` isn't a vanilla tag, Why is this in the `minecraft` resource space?

This tag also doesn't make any sense as `items/brass` would imply it contains all brass items when this only has brass sheets. I assume this shouldn't exist.

Claycorp (changes requested):
[Hey, that's pretty good! ](https://www.youtube.com/watch?v=JeimE8Wz6e4)

resources/recipes.py
Nyonyix: Yea, Got confused a bit, Had an issue where it would gen tags with 2 's'. Got it sorted now.

As to why it's not a function, I was just following the pre-existing code. Looking at the unmodified file, It's the same.

src/main/resources/data/minecraft/tags/items/brass.json
eerussianguy: the proper namespace should be forge, yeah

resources/recipes.py
```diff
@@ -676,7 +676,7 @@ def item_tag(namespace: str, _variant: str) -> str:
 
         # Tools
         if 'tool' in metal_data.types:
-            anvil_recipe(rm, '%s_tuyere' % metal, '#forge:double_sheet/%s' % metal, 'tfc:metal/tuyere/%s' % metal, metal_data.tier, Rules.bend_last, Rules.bend_second_last)
+            anvil_recipe(rm, '%s_tuyere' % metal, item_tag('forge', 'double_sheet'), 'tfc:metal/tuyere/%s' % metal, metal_data.tier, Rules.bend_last, Rules.bend_second_last)
```
Claycorp, line 679: Could have fixed the item result to match everything else using `item('tuyere)` but not a huge deal.

Just something to keep in mind when working in the future that it's best to update stuff in one shot while your already there vs just leaving it, Keeps things cleaner for the future. There's also likely other places this could apply too.

Claycorp (approved):
Looks good over all minus the nitpick but looks fine. 👍

PR #2005, opened 2022-06-27
Prospecting book entries

Gaelmare:


resources/generate_book.py
```diff
@@ -515,7 +515,13 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
             # todo: supports, more on collapses?
         )),
         entry('prospecting', 'Prospecting', 'tfc:metal/propick/wrought_iron', pages=(
-            image('tfc:textures/gui/book/wip.png'),
+            text('You remembered where you picked up those $(l:getting_started/finding_ores)small metal nuggets$(), right? Finding additional ores may require extensive exploration and mining. You should become very familiar with the $(l:the_world/ores_and_minerals)Ores and Minerals$(). If you need a specific resource, you must find the rock type it spawns in, either under your feet or across the world.'),
```
eerussianguy: Might capitalize small metal nuggets

resources/generate_book.py
```diff
@@ -515,7 +515,13 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            # todo: reference configuration values rather than defaults?
```
eerussianguy: I think for now we are not referenciing configs.

resources/generate_book.py
```diff
@@ -515,7 +515,13 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            text('When picking up small nuggets or mining exposed ore becomes unsatisfying, it is time to consider prospecting to find concealed ore veins.$(br2)The small nuggets occur when ore is nearby, within 15 blocks horizontally and 35 vertically.$(br2)If you find the center of a group of nuggets, it\'s likely that the vein is beneath you.'),
+            item_spotlight('tfc:metal/propick/copper', 'Prospector\'s Pick', text_contents='If you\'re looking for minerals, which have no metal nuggets, or you can\'t find the vein by guessing, it\'s time to pull out the $(thing)Prospector\'s Pick$(). It searches the 25x25x25 area centered on the block clicked, and reports to the action bar the amount and type of ore located.'),
+            text('The Prospector\'s Pick can give false negatives, but will never give a false positive. Higher tier tools can reduce or eliminate false negatives.$(br2)A propick of the same metal tier will always have the same result when used on the same block unless ores were removed.$(br2)If the Prospector\'s Pick finds multiple ores nearby, it will select only one to report.'),
```
eerussianguy: I don't know if its obvious what false negative means in this case to someone who hasn't played. I would say more like, "there is a chance it will not find anything, even if there is something there" or something like that

resources/generate_book.py
```diff
@@ -515,8 +515,12 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            text('When picking up small nuggets becomes unsatisfying, it is time to start prospecting to find ore veins:$(br)$(li)Small nuggets occur when ore is nearby, within 15 blocks horizontally and 35 vertically.If you find the center of a group of nuggets, it\'s likely that the vein is beneath you.$(li)Exposed ore can occur in cliffs and water bodies, which may be seen from farther away.'),
+            item_spotlight('tfc:metal/propick/copper', 'Prospector\'s Pick', text_contents='If you\'re looking for metal ores or mineral veins (which have no nuggets), and you can\'t find the vein by guessing, it\'s time to pull out the $(thing)Prospector\'s Pick$(). It searches the 25x25x25 area centered on the block clicked, and reports to the action bar the amount and type of ore located.').link(*['tfc:metal/propick/%s' % m for m in TOOL_METALS]),
```
alcatrazEscapee: Insert three pages here - knapping recipe for propick mold, heating recipe for propick, and anvil recipe for propick head. Talk a little about how to obtain both versions.

resources/generate_book.py
```diff
@@ -515,8 +515,12 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            text('The Prospector\'s Pick may incorrectly say nothing is there. It will never report finding something when nothing is actually there. Higher tier tools will reduce or eliminate false reports.$(br2)A propick of the same metal tier will always have the same result when used on the same block unless ores were removed.$(br2)If the Prospector\'s Pick finds multiple ores nearby, it will select only one to report.'),
+            text('Right-clicking a propick on a block will report finding one of these possible results:$(br)$(li)nothing (may be false)$(li)traces$(li)small sample$(li)medium sample$(li)large sample$(li)very large sample$(br2)Very large samples indicate at least eighty and potentially many more blocks.'),
+            text('Remember that the ores can be in any direction from the block clicked, including up and down.$(br2)You can place markers like torches or straw where propick results change. If you do so in multiple directions, the vein will be between and near the middle of your markers.'),
```
alcatrazEscapee: Remove this page, it's essentially saying "reminder you can place blocks" which is dumb, and makes an unsubstantiated claim.

resources/generate_book.py
```diff
@@ -515,8 +515,12 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            item_spotlight('tfc:metal/propick/copper', 'Prospector\'s Pick', text_contents='If you\'re looking for metal ores or mineral veins (which have no nuggets), and you can\'t find the vein by guessing, it\'s time to pull out the $(thing)Prospector\'s Pick$(). It searches the 25x25x25 area centered on the block clicked, and reports to the action bar the amount and type of ore located.').link(*['tfc:metal/propick/%s' % m for m in TOOL_METALS]),
+            text('The Prospector\'s Pick may incorrectly say nothing is there. It will never report finding something when nothing is actually there. Higher tier tools will reduce or eliminate false reports.$(br2)A propick of the same metal tier will always have the same result when used on the same block unless ores were removed.$(br2)If the Prospector\'s Pick finds multiple ores nearby, it will select only one to report.'),
+            text('Right-clicking a propick on a block will report finding one of these possible results:$(br)$(li)nothing (may be false)$(li)traces$(li)small sample$(li)medium sample$(li)large sample$(li)very large sample$(br2)Very large samples indicate at least eighty and potentially many more blocks.'),
```
alcatrazEscapee:
```suggestion
            text('Right-clicking a Prospector\'s pick on a block will report finding one of these possible results:$(br)$(li)Nothing (may be false)$(li)Traces$(li)A Small Sample$(li)A Medium Sample$(li)A Large Sample$(li)A Very Large Sample$(br2)Very large samples indicate finding at least eighty and potentially many more blocks.'),
```

resources/generate_book.py
```diff
@@ -515,8 +515,12 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            item_spotlight('tfc:metal/propick/copper', 'Prospector\'s Pick', text_contents='If you\'re looking for metal ores or mineral veins (which have no nuggets), and you can\'t find the vein by guessing, it\'s time to pull out the $(thing)Prospector\'s Pick$(). It searches the 25x25x25 area centered on the block clicked, and reports to the action bar the amount and type of ore located.').link(*['tfc:metal/propick/%s' % m for m in TOOL_METALS]),
+            text('The Prospector\'s Pick may incorrectly say nothing is there. It will never report finding something when nothing is actually there. Higher tier tools will reduce or eliminate false reports.$(br2)A propick of the same metal tier will always have the same result when used on the same block unless ores were removed.$(br2)If the Prospector\'s Pick finds multiple ores nearby, it will select only one to report.'),
```
alcatrazEscapee:
```suggestion
            text('The Prospector\'s Pick may incorrectly say nothing is there. It will never report finding something when nothing is actually there. Higher tier tools will reduce or eliminate these false negatives.$(br2)A Prospector\'s Pick of the same metal tier will always have the same result when used on the same block unless ores were removed.$(br2)If the Prospector\'s Pick finds multiple ores nearby, it will select only one to report.'),
```

alcatrazEscapee (changes requested):

resources/generate_book.py
Gaelmare: I described how I usually prospect for veins if I get a prospecting pick return, either at the surface or while mining. With metaballs, I know the ore vein shapes will be much more varied, but enough to invalidate this technique? Should I attempt to describe a prospecting technique at all?

resources/generate_book.py
Gaelmare: I feel this leaves the list looking a little strange: 
image


If that's what you want, I'm good with it.

Gaelmare: I was matching the case of the in-game text from propick results.
image

resources/generate_book.py
Gaelmare: Knapping recipe for mold is already present in the pottery section, should I just link there instead?
When you say heating recipe, do you mean casting metal into a mold or firing the mold itself?

Gaelmare: I described casting, and put the firing recipe in.

PR #2017, opened 2022-07-17
start of spreading crops. Also require light to grow crops.

### LINKED ISSUE #2018: Mammals have difficulty giving birth if their birth time passes while unloaded
Still needs to be confirmed and tested properly. Part of #2012

eerussianguy:


src/main/java/net/dries007/tfc/common/blockentities/RottingBlockEntity.java
```diff
@@ -0,0 +1,65 @@
...
+package net.dries007.tfc.common.blockentities;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.world.level.block.entity.BlockEntityType;
+import net.minecraft.world.level.block.state.BlockState;
+
+import net.dries007.tfc.util.calendar.Calendars;
+
+public class RottingBlockEntity extends TFCBlockEntity
```
alcatrazEscapee: As this is just `TickCounterBlockEntity` but with a different name / helper methods, I think it might just be cleaner to have all the rotting specific logic in the block, and then use the tick counter block entity.

src/main/java/net/dries007/tfc/common/blocks/crop/RottingBlock.java
```diff
@@ -0,0 +1,70 @@
+/*
```
alcatrazEscapee: Suggest `DecayingBlock` instead.

src/main/java/net/dries007/tfc/common/blocks/crop/TFCPumpkinBlock.java
```diff
@@ -0,0 +1,65 @@
...
+    {
+        super(properties, rotted);
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public InteractionResult use(BlockState state, Level level, BlockPos pos, Player player, InteractionHand hand, BlockHitResult hit)
+    {
+        ItemStack held = player.getItemInHand(hand);
+        if (Helpers.isItem(held, TFCTags.Items.KNIVES) && TFCConfig.SERVER.enablePumpkinCarving.get())
```
alcatrazEscapee: I'd make this knives || shears, since vanilla does shears it's reasonable to allow that as well.

src/main/java/net/dries007/tfc/util/loot/TFCLoot.java
```diff
@@ -37,12 +37,14 @@
...
     public static final LootContextParam<Boolean> PANNED = new LootContextParam<>(Helpers.identifier("panned"));
+    public static final LootContextParam<Boolean> DECAY_HANDLED = new LootContextParam<>(Helpers.identifier("decay_handled"));
 
     public static final RegistryObject<LootItemConditionType> IS_PANNED = lootCondition("is_panned", new PannedCondition.Serializer());
     public static final RegistryObject<LootItemConditionType> IS_ISOLATED = lootCondition("is_isolated", new IsIsolatedCondition.Serializer());
     public static final RegistryObject<LootItemConditionType> ALWAYS_TRUE = lootCondition("always_true", new AlwaysTrueCondition.Serializer());
     public static final RegistryObject<LootNumberProviderType> CROP_YIELD = numberProvider("crop_yield_uniform", new CropYieldProvider.Serializer());
     public static final RegistryObject<LootItemFunctionType> COPY_FLUID = lootFunction("copy_fluid", new CopyFluidFunction.Serializer());
+    public static final RegistryObject<LootItemFunctionType> ROTTING_BLOCK = lootFunction("rotting_block", new RottingBlockLootFunction.Serializer());
```
alcatrazEscapee: This name doesn't really specify what it does, I'd suggest `copy_decay` and `CopyDecayLootFunction`

src/main/java/net/dries007/tfc/common/blocks/crop/RottingBlock.java
```diff
@@ -0,0 +1,70 @@
...
+                level.setBlockAndUpdate(pos, rotted.get().defaultBlockState());
+            }
+        }
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public List<ItemStack> getDrops(BlockState state, LootContext.Builder builder)
+    {
+        builder = builder.withParameter(TFCLoot.DECAY_HANDLED, true);
```
alcatrazEscapee: As I understand, this is to prevent the GLM from running in this block specifically. I don't *like* this solution, but I can't think of a better way to avoid it at the moment. I'd like that just a custom function would be enough? but apparently not.

Also curious if there is a need to keep the GLM for entities specifically. Looking at `LootItem`, it does not appear to construct an `ItemStack` (thus forcing decay), and constructs them on demand instead?

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/common/items/Food.java
```diff
@@ -89,6 +89,8 @@
     CATTAIL_ROOT,
     TARO_ROOT,
     DRIED_KELP,
+    PUMPKIN_PIE,
+    MELON_SLICE,
```
alcatrazEscapee: Is there a reason we can't use vanilla items here, since these are supposedly 1-1 drop ins?

eerussianguy:
Also fixed: pregnancy serialization, food decay

resources/generate_book.py
```diff
@@ -869,6 +869,10 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
             multimultiblock('', *[two_tall_block_spotlight('', '', 'tfc:farmland/loam', 'tfc:crop/potato[age=%d]' % i) for i in range(7)]),
+            text(f'{detail_crop("pumpkin")}Pumpkins are a spreading crop. Pumpkin seeds can be planted on farmland to be grown, and will place up to two $(thing)Pumpkin Blocks$() on the ground next to it while it is mature. If the pumpkin blocks are harvested, if the pumpkin plant matures again, it can grow more pumpkins.').link('tfc:seeds/pumpkin').anchor('pumpkin'),
+            multimultiblock('', *[two_tall_block_spotlight('', '', 'tfc:farmland/loam', 'tfc:crop/pumpkin[age=%d]' % i) for i in range(8)]),
+            text(f'{detail_crop("melon")}Melons are a spreading crop. Melon seeds can be planted on farmland to be grown, and will place up to two $(thing)Melon Blocks$() on the ground next to it while it is mature. If the melon blocks are harvested, if the pumpkin plant matures again, it can grow more melon.').link('tfc:seeds/melon').anchor('melon'),
+            multimultiblock('', *[two_tall_block_spotlight('', '', 'tfc:farmland/loam', 'tfc:crop/pumpkin[age=%d]' % i) for i in range(8)]),
```
alcatrazEscapee: Maybe make these actually show the pumpkin and melon blocks getting placed on adjacent blocks? Have a 3x3 wide footprint?

src/main/java/net/dries007/tfc/common/blockentities/DecayingBlockEntity.java
```diff
@@ -0,0 +1,66 @@
...
+    public ItemStack getStack()
+    {
+        return stack;
+    }
+
+    public void setStack(ItemStack stack)
+    {
+        stack = stack.copy();
+        stack = stack.split(1);
+        this.stack = stack;
```
alcatrazEscapee: This is redundant - could just be `this.stack = Helpers.copyWithSize(stack, 1)`

src/main/java/net/dries007/tfc/common/blockentities/TickCounterBlockEntity.java
```diff
@@ -38,6 +38,23 @@ public long getTicksSinceUpdate()
...
+     */
+    public void setLastUpdateTick(long tick)
+    {
+        lastUpdateTick = tick;
+        setChanged();
+    }
+
+    /**
+     * Use getTicksSinceUpdate over this. Only use if you know what you're doing.
+     */
```
alcatrazEscapee: These are bad comments. If these functions have strange or unexpected behavior, *document that behavior*. This is like code gatekeeping lol.

src/main/java/net/dries007/tfc/common/blocks/crop/ClimbingCropBlock.java
```diff
@@ -85,6 +85,10 @@ protected void createBlockStateDefinition(StateDefinition.Builder<Block, BlockSt
     @Override
     public float getGrowthLimit(Level level, BlockPos pos, BlockState state)
     {
+        if (!level.canSeeSky(pos))
+        {
+            return 0f;
+        }
```
alcatrazEscapee: Why not allow block light?

src/main/java/net/dries007/tfc/common/blocks/crop/ICropBlock.java
```diff
@@ -26,7 +26,7 @@
      */
     default float getGrowthLimit(Level level, BlockPos pos, BlockState state)
     {
-        return CropHelpers.GROWTH_LIMIT;
+        return level.canSeeSky(pos) ? CropHelpers.GROWTH_LIMIT : 0f;
```
alcatrazEscapee: Also with the block light.

PR #2031, opened 2022-08-12
wip book work

### LINKED ISSUE #2019: Hostile Spawn Eggs are affected by the `enableVanillaMonstersOnSurface` config option.
Hostile mobs cannot be spawned in light areas from spawn eggs when `enableVanillaMonstersOnSurface = false`. Affects creative mode only.

eerussianguy:


resources/generate_book.py
```diff
@@ -560,6 +560,14 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
             text('The $(thing)Thatch Bed$() is a primitive bed which can be used to set your spawn, although not to sleep through the night. To make a thatch bed, place two $(thing)Thatch$() blocks adjacent to each other. Then, right click with a $(thing)Large Raw Hide$(). Large hides are dropped by larger animals, like $(thing)bears$() and $(thing)cows$().'),
             multiblock('Thatch Bed', 'A constructed thatch bed.', False, mapping={'0': 'tfc:thatch_bed[part=head,facing=west]', 'D': 'tfc:thatch_bed[part=foot,facing=east]'}, pattern=((' D ', ' 0 '),)),
         )),
+        entry('size_and_weight', 'Size and Weight', 'tfc:wood/chest/kapok', pages=(
+            text('Every item has a $(thing)Size ⇲$() and $(thing)Weight \u2696$(). An item\'s size and wieght is shown on the item\'s $(thing)tooltip$(), which appears when you hover over it with your mouse.'),
```
alcatrazEscapee, line 583: Why one unicode char (⇲) and one codepoint (`\u2696`)?

resources/generate_book.py
```diff
@@ -560,6 +560,14 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            text('Size determines what storage blocks an item can fit inside of.$(br)$(li)$(thing)Tiny$() items fit in anything.$(li)$(thing)Very Small$() items fit in anything.$(li)$(thing)Small$() items are the largest item that will fit in $(l:mechanics/decay#small_vessels)Small Vessels$()$().$(li)Normal$() items are the largest that will fit in $(l:mechanics/decay#small_vessels)Large Vessels$().', title='Size ⇲'),
+            text('$(li)$(thing)Large$() items are the largest that will fit in Chests. $(l:getting_started/pit_kiln)Pit Kilns$() or Placed Items can hold four.$(li)$(thing)Very Large$() items are stored alone in Pit Kilns and Placed Items.$(li)$(thing)Huge$() items do not fit in any normal storage device. They count towards overburdening.'),
```
alcatrazEscapee: Just say pit kilns - we don't call anything explicitly "placed items", and the distinction doesn't exist in 1.18 from the player PoV. Also worth mentioning pit kilns can hold either four < very large items or one >= very large item.

resources/generate_book.py
```diff
@@ -560,6 +560,14 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            text('$(thing)Overburdening happens when you carry $(thing)Huge$() items. Carrying just one Huge item causes you to become exhausted, making your food burn quicker. Carrying two or more gives you the $(thing)Overburdened$() status effect, which makes movement very slow.$(br)Devices that hold their contents when dropped, such as $(l:mechanics/barrels)Barrels$() and $(l:mechanics/crucible)Crucibles$(), as well as $(l:mechanics/anvils)Anvils$() are Huge. $(l:mechanics/animal_husbandry#horses)Horses$() can be overburdened.', title='Overburdening'),
```
alcatrazEscapee: It shouldn't just be huge - it should be huge + very heavy. Barrels are huge, but their weight changes when sealed vs. not sealed, which triggers the overburdening effect. (And if that isn't how it's implemented, it should be made that way).

alcatrazEscapee (changes requested):

resources/generate_book.py
```diff
@@ -329,25 +329,56 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+        entry('wild_fruits', 'Wild Fruits', 'tfc:food/elderberry', pages=(
             # Wild fruits
             text('Many different varieties of wild fruits can be found growing in the world. These can be collected to be eaten, or farmed, with the right equipment. These can be found on different varieties of bushes or trees. In general, fruits can be found in three types of plants: $(l:the_world/wild_fruits#fruit_trees)Fruit Trees$(), $(l:the_world/wild_fruits#tall_bushes)Tall Bushes$()Tall Bushes$(), and $(l:the_world/wild_fruits#small_bushes)Small Bushes$().$(br2)All fruiting plants have a common lifecycle. They will grow, form flowers, sprout fruit, and then lay dormant in a yearly cycle.'),
-            text('Different fruit plants will grow, flower, and fruit at different times of year.'),  # todo: finish
+            text('Fruit plants are seasonal. During their cold season, these plants will appear brown and lifeless. In the spring, they become green and healthy, getting ready to produce fruit and grow larger. The exact times this happen varies by the fruit, with no two being the same. Fruit plants can die, as well: of old age, and of improper climate conditions.'),
```
alcatrazEscapee, line 335: Old age?

resources/generate_book.py
```diff
@@ -329,25 +329,56 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
+            text('Saplings can also be placed on the first \'elbow\' sections a fruit tree produces, places where the branch is attached to one side and upwards. This allows one fruit tree to grow multiple fruits. Breaking these elbows with a $(thing)Axe$() also yields saplings. Harvesting a fruit tree leaf is done with $(item)$(k:key.use)$() when the fruit tree is at its fruiting stage. This will give one fruit, and revert the plant back to its growing stage, until it is time for it to become dormant.'),
             page_break(),
-            text('', title='Tall Bushes').anchor('tall_bushes'),
-            text(''),
-            # todo: blurb on how to harvest and farm tall bushes in general
-            # todo: blurb on all tall bush species
+            detail_fruit_tree_left('cherry'),  # todo: fancy tree vis?
+            detail_fruit_tree_right('cherry'),
```
alcatrazEscapee: Rather than having these `left` and `right` methods, have one that returns a tuple. Then you can call it like

```python
    *detail_fruit_tree('cherry'),
    *detail_fruit_tree('green_apple')
```

And, you can also include explicit `page_break()`s:

```python
def detail_fruit_tree():
    return (
        page_break(),
        # the detail pages
```

alcatrazEscapee (changes requested):

resources/generate_book.py
eerussianguy: I guess this isnt a thing anymore

alcatrazEscapee (approved):

PR #2047, opened 2022-08-21
Grammar and a few other edits

Gaelmare:


resources/generate_book.py
```diff
@@ -472,7 +472,7 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
             ),
         )),
         entry('pottery', 'Pottery', 'tfc:ceramic/vessel', pages=(
-            text('$(thing)Clay$() is an incredibly useful and balanced material, which can be used for pottery. However first, it needs to be located. Clay is usually hidden by grass, but it is found often in two locations. In areas with of at least 175mm $(l:the_world/climate#rainfall)Annual Rainfall$(), clay can be found in patches all over the place, however these patches are usually marked the by presence of certain $(thing)Plants$().'),
+            text('$(thing)Clay$() is an incredibly useful material which can be used for pottery. It can prove challenging to locate at first. Clay is usually hidden by grass, but it is often found in two locations. In areas with at least 175mm $(l:the_world/climate#rainfall)Annual Rainfall$(), clay can be found in patches all over the place, usually marked by the presence of certain $(thing)Plants$().'),
```
alcatrazEscapee: Hey I wrote that joke and I like it. The other edits in this page are fine.

resources/generate_book.py
```diff
@@ -515,33 +515,33 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
-            text('First, open the $(thing)Small Vessel$() and put the ores inside. Count up the total amount of metal in the ores carefully! Then, you need to build a $(l:getting_started/pit_kiln)Pit Kiln$(), and this time, put the entire small vessel inside it. As the vessel heats, the ores inside it will melt, and you\'ll be left with a vessel of molten metal.$(br2)Take the vessel out and $(item)$(k:key.use)$() it, to open the $(thing)Casting$() interface.'),
+            text('You will need:$(br)$(li)A $(l:getting_started/pottery#vessel)Small Vessel$()$(li)Enough materials for a $(l:getting_started/pit_kiln)Pit Kiln$().$(li)One or more $(l:getting_started/pottery#mold)Mold(s)$() to cast the molten metal.$(li)And finally, at least 100 mB total of a metal which is suitable for casting: $(thing)Copper$() including its three ores.$(br2)$(br)$(italic)Note: Casting can also be done with some $(l:getting_started/primitive_alloys)Alloys$()'),
```
alcatrazEscapee: "Copper, including its three ores" doesn't make sense. You need an ore of copper, namely, one of the three ores of copper. There is no "including", i.e. you need "copper, *and* one of it's three ores".

resources/generate_book.py
```diff
@@ -515,33 +515,33 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
-            crafting('tfc:crafting/metal/pickaxe/copper', text_contents='With a tool head in hand, you are now able to craft your first pickaxe! Find enough copper to make a single pickaxe head, cast it using a mold and a few pit kilns, and then slap it on a stick, and voila!'),
+            text('With the casting interface open, place your empty fired mold in the center slot. It will fill up as long as the vessel remains liquid. (If the vessel solidifies, it can be reheated in another pit kiln.) Once the mold is full, it can be removed and left to cool. Once cool, the mold and its contents can be extracted by using the mold, or putting it in the crafting grid.'),
+            crafting('tfc:crafting/metal/pickaxe/copper', text_contents='With a tool head in hand, you are now able to craft your first pickaxe! Find enough copper to make a pickaxe head, fire a pickaxe mold and melt the ore using pit kilns, then cast a head. Slap it on a stick, and voila!'),
```
alcatrazEscapee:
```suggestion
            crafting('tfc:crafting/metal/pickaxe/copper', text_contents='With a tool head in hand, you are now able to craft your first pickaxe! Find enough copper to make a pickaxe head, fire a pickaxe mold and melt the ore using a pit kiln, then cast a head. Slap it on a stick, and voila!'),
```

resources/generate_book.py
```diff
@@ -576,22 +576,26 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
             image('tfc:textures/gui/book/tutorial/wattle_add_stick.png', text_contents='Adding sticks to wattle.'),
             image('tfc:textures/gui/book/tutorial/wattle_add_daub.png', text_contents='Using $(thing)Daub on $(thing)Woven Wattle$() creates a solid block.'),
             image('tfc:textures/gui/book/tutorial/wattle_stained.png', text_contents='It can then be $(thing)stained$() by using dye on it.'),
+            # Clay Blocks and Peat
+            page_break(),
+            # todo: Clay block recipe and peat spawning details/screenshot
+            text('Clay obtained from the ground can be crafted into clay blocks. While unattractive, they are easy to get. When placed and dug again, they\'ll turn back into clay. Peat spawns in the world in patches along water bodies, and is also easy to obtain with stone tools. Peat is flammable, however.', title='Clay Blocks and Peat').anchor('clay_and_peat'),
```
alcatrazEscapee: This will error when generating, because you've broken the page boundaries. `page_break()` does not *insert* a page break, it requires that the page boundaries that naturally occur, match up with where you expect them to be. Either write this page that you've `//todo`'d, or remove this entry, because it's broken as is.

resources/generate_book.py
```diff
@@ -607,57 +611,57 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False):
...
-            leather_knapping('tfc:leather_knapping/saddle', '$(thing)Horses$() are $(l:mechanics/animal_husbandry#mammals)Mammals$() that can be ridden when tamed. Horses become rideable after reaching 15% familiarity.').anchor('horses'),
-            text(' They need a $(thing)Saddle$() to ride, which can be $(thing)Knapped$(). This includes $(l:mechanics/animal_husbandry#mule)Mules$(), $(l:mechanics/animal_husbandry#donkey)Donkeys$(), and $(l:mechanics/animal_husbandry#Horses)Horses$().$(br2)The next few pages will go over the different types of domesticatable animals.'),
+            heat_recipe('tfc:heating/cooked_egg', '$(thing)Eggs$() can be cooked or boiled for food. Male oviparous animals can fertilize females, which causes the next egg laid in the nest box to be fertilized. Fertilized eggs will have a tooltip with how long until they are ready to hatch.'),
+            leather_knapping('tfc:leather_knapping/saddle', '$(thing)Equines() are $(l:mechanics/animal_husbandry#mammals)Mammals$() that can be ridden when tamed. They become rideable after reaching 15% familiarity.').anchor('horses'),
```
alcatrazEscapee:
```suggestion
            leather_knapping('tfc:leather_knapping/saddle', '$(thing)Equines$() are $(l:mechanics/animal_husbandry#mammals)Mammals$() that can be ridden when tamed. They become rideable after reaching 15% familiarity.').anchor('horses'),
```

alcatrazEscapee (changes requested):
Please run `python resources book` and fix the validation errors that occur.

resources/generate_book.py
Gaelmare: intention was that you could melt down other copper items, you're not limited to just ores. Make sense?

resources/generate_book.py
Gaelmare: Thanks for catching this. This was a preexisting issue with the Mule and Donkey entries which I fixed, and then introduced this one!

resources/generate_book.py
Gaelmare: it also doesn't have to be only one of the ores

resources/generate_book.py
Gaelmare: I don't know which joke/edit you're referring to here, sorry!

alcatrazEscapee: > Clay is an incredibly useful and balanced material,

resources/generate_book.py
alcatrazEscapee: While yes, this is a getting started guide to obtaining your first metal - it is written as such. If the guide to obtaining your first metal said you could obtain it by using some copper ingots, it would seem counter intuitive. Hence referring to ores.

And despite that intention I still think your wording is off. "such as Copper, which can be obtained from one if its three ore forms" is not exclusive, but it gets the desired point across.

resources/generate_book.py
Gaelmare: Oh, good, my first guess was correct!

Gaelmare:
Alc, did you mean `python resources book` or `generate_book.py`?

PR #2049, opened 2022-08-21
Minor cleanup for JEI slot backgrounds, mark optional dependency on JEI 9.7+

mezz:
Just wanted to check out the JEI integration since it's such a big mod. Looks pretty nice, here's some minor improvements to the slot drawing since I was already poking at it. Cheers.


alcatrazEscapee:
Much appreciated!

PR #2050, opened 2022-08-22
Advancements

eerussianguy:
…to accomodate advancements


resources/advancements.py
```diff
@@ -0,0 +1,104 @@
+from mcresources import ResourceManager
+from mcresources import advancements as adv
```
alcatrazEscapee: This is really bad and makes refactoring almost impossible. Prefer the following:

```python
from mcresources import ResourceManager, advancements
from mcresources.advancements import AdvancementCategory

File: resources/assets.py [hunk 2]
```diff
@@ -110,9 +110,10 @@ def generate(rm: ResourceManager):
                     block.with_lang(lang('%s %s', rock, block_type))
 
         if rock_data.category == 'igneous_extrusive' or rock_data.category == 'igneous_intrusive':
-            rm.blockstate('tfc:rock/anvil/%s' % rock, model='tfc:block/rock/anvil/%s' % rock).with_lang(lang('%s Anvil', rock)).with_block_loot('1-4 tfc:rock/loose/%s' % rock).with_item_model()
+            rm.blockstate('tfc:rock/anvil/%s' % rock, model='tfc:block/rock/anvil/%s' % rock).with_lang(lang('%s Anvil', rock)).with_block_loot('1-4 tfc:rock/loose/%s' % rock).with_item_model().with_tag('tfc:rock_anvils')
             rm.block_model('tfc:rock/anvil/%s' % rock, parent='tfc:block/rock/anvil', textures={'texture': 'tfc:block/rock/raw/%s' % rock})
             rm.blockstate('tfc:rock/magma/%s' % rock, model='tfc:block/rock/magma/%s' % rock).with_block_model(parent='minecraft:block/cube_all', textures={'all': 'tfc:block/rock/magma/%s' % rock}).with_lang(lang('%s magma block', rock)).with_item_model().with_block_loot('tfc:rock/magma/%s' % rock)
+            rm.item_tag('rock_anvils', 'tfc:rock/anvil/%s' % rock)
```
alcatrazEscapee: If possible, can we move tags to `data` where all the other tags are? Rather than introducing new ones here as it makes it easier to track down them.

resources/advancements.py
```diff
@@ -0,0 +1,177 @@
+from mcresources import ResourceManager, utils
+from mcresources import advancements as adv
+from mcresources.type_definitions import Json
+
+from constants import *
+
+
+def main(rm: ResourceManager):
```
alcatrazEscapee: Not `main`, it's not a main method (invoked from `__main__`) `generate` instead.

resources/data.py
```diff
@@ -915,6 +916,9 @@ def needs_tool(_tool: str) -> str:
     mob_loot(rm, 'horse', 'tfc:food/horse_meat', 4, 10, 'medium', bones=6)
     mob_loot(rm, 'minecraft:zombie', 'minecraft:rotten_flesh', 0, 2)  # it drops vanilla stuff we do not want
 
+    import advancements
+    advancements.main(rm)
```
alcatrazEscapee: Don't add this here - add it as a separate target in `__main__.py`

src/main/java/net/dries007/tfc/ForgeEventHandler.java
```diff
@@ -603,6 +604,10 @@ else if (block == TFCBlocks.LOG_PILE.get())
         else if (block == TFCBlocks.PIT_KILN.get() && state.getValue(PitKilnBlock.STAGE) == 15)
         {
             level.getBlockEntity(pos, TFCBlockEntities.PIT_KILN.get()).ifPresent(PitKilnBlockEntity::tryLight);
+            if (event.getPlayer() instanceof ServerPlayer serverPlayer)
+            {
+                SpecialBlockTrigger.LIT.trigger(serverPlayer, state);
+            }
```
alcatrazEscapee: This feels a bit weird, that it only triggers on pit kilns despite being a `LIT` trigger. What about literally everything else in this event handler? What if `tryLight` fails?

src/main/java/net/dries007/tfc/util/events/SpecialBlockTrigger.java
```diff
@@ -0,0 +1,73 @@
...
+    public static void registerSpecialBlockTriggers()
+    {
+        CriteriaTriggers.register(CHISELED);
+        CriteriaTriggers.register(LIT);
+        CriteriaTriggers.register(ROCK_ANVIL);
+        CriteriaTriggers.register(FIREPIT_CREATED);
+    }
+
+    public static final SpecialBlockTrigger CHISELED = new SpecialBlockTrigger(Helpers.identifier("chiseled"));
+    public static final SpecialBlockTrigger LIT = new SpecialBlockTrigger(Helpers.identifier("lit"));
```
alcatrazEscapee: As mentioned above, this is not `LIT`, this is specifically a pit kiln **attempted** to be lit.

src/main/java/net/dries007/tfc/util/events/SpecialBlockTrigger.java
```diff
@@ -0,0 +1,73 @@
...
+import net.minecraft.advancements.critereon.SimpleCriterionTrigger;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.level.block.state.BlockState;
+
+import net.dries007.tfc.common.recipes.ingredients.BlockIngredient;
+import net.dries007.tfc.common.recipes.ingredients.BlockIngredients;
+import net.dries007.tfc.util.Helpers;
+
+public class SpecialBlockTrigger extends SimpleCriterionTrigger<SpecialBlockTrigger.TriggerInstance>
```
alcatrazEscapee: I don't like the name `SpecialBlockTrigger` but the best I have is `BlockActionTrigger`

src/main/java/net/dries007/tfc/util/events/SpecialBlockTrigger.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Licensed under the EUPL, Version 1.2.
+ * You may obtain a copy of the Licence at:
+ * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ */
+
+package net.dries007.tfc.util.events;
```
alcatrazEscapee: This is the wrong package, it's not an event. None of this is. If anything it should be in an `util.advancements` like `util.loot`

src/main/java/net/dries007/tfc/util/events/SpecialBlockTrigger.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class SpecialBlockTrigger extends SimpleCriterionTrigger<SpecialBlockTrigger.TriggerInstance>
+{
+    public static void registerSpecialBlockTriggers()
+    {
+        CriteriaTriggers.register(CHISELED);
+        CriteriaTriggers.register(LIT);
+        CriteriaTriggers.register(ROCK_ANVIL);
+        CriteriaTriggers.register(FIREPIT_CREATED);
+    }
```
alcatrazEscapee: I also really don't like having all these instances collected in `SpecialBlockTrigger` and `SpecialEntityTrigger`. Can we have a `TFCAdvancements` which contains all the registered instances (and also a `register()` method for convenience rather than re-calling `CriteriaTrigers.register()`, `Helpers.identifier`, etc.)

src/main/java/net/dries007/tfc/util/events/SpecialEntityTrigger.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    protected SpecialEntityTrigger.TriggerInstance createInstance(JsonObject json, EntityPredicate.Composite predicate, DeserializationContext context)
+    {
+        EntityPredicate ingredient = EntityPredicate.fromJson(json.get("entity"));
+        if (ingredient == EntityPredicate.ANY)
+        {
+            TerraFirmaCraft.LOGGER.error("Entity predicate '{}' failed to load: ", id);
+        }
```
alcatrazEscapee: 1. Don't use the `TerraFirmaCraft` logger.
2. Can you not throw a `JsonSyntaxException` here? I don't believe this is in codec land of unable-to-throw-exception hell.

src/main/java/net/dries007/tfc/util/events/SpecialEventTrigger.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.google.gson.JsonObject;
+import net.minecraft.advancements.CriteriaTriggers;
+import net.minecraft.advancements.critereon.*;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.level.ServerPlayer;
+
+import net.dries007.tfc.util.Helpers;
+
+public class SpecialEventTrigger extends SimpleCriterionTrigger<SpecialEventTrigger.TriggerInstance>
```
alcatrazEscapee: `GenericTrigger` maybe?

src/main/java/net/dries007/tfc/util/events/SpecialEventTrigger.java
```diff
@@ -0,0 +1,65 @@
...
+{
+    public static void registerSpecialEventTriggers()
```
alcatrazEscapee: Again the note about collecting and centralizing all the registrations,

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
```diff
@@ -63,9 +68,10 @@ public static boolean startFire(Level world, BlockPos pos, BlockState state, Dir
     private final BlockState state;
     private final Direction direction;
     private final Player player;
+    @Nullable
     private final ItemStack stack;
```
alcatrazEscapee: Nullable item stacks are really ew, is this actually necessary? Surely an `ItemStack.EMPTY` would make more sense here, unless there's something actually to be gained from knowing the stack is null?

src/main/java/net/dries007/tfc/util/loot/CustomItemPredicate.java
```diff
@@ -0,0 +1,109 @@
...
+    {
+        if (json.has("metal"))
+        {
+            ResourceLocation metalLocation = new ResourceLocation(JsonHelpers.getAsString(json, "metal"));
+            Metal metal = Metal.MANAGER.get(metalLocation);
+            if (metal != null)
+            {
+                return metal;
+            }
+            TerraFirmaCraft.LOGGER.error("Not a metal: " + metalLocation);
```
alcatrazEscapee: Don't use the `TerraFirmaCraft` logger, and throw `JsonSynaxException`, not log errors.

resources/advancements.py
```diff
@@ -0,0 +1,177 @@
...
+    world.advancement('metallurgist', icon('tfc:metal/ingot/gold'), 'Metallurgist', 'Obtain a metal-bearing specimen for every metal in TFC.', 'nugget', metal_criteria(), requirements=[[m] for m in METALS], frame='goal')
+
+def metal_criteria() -> Json:
+    criteria = dict()
+    for metal in METALS.keys():
+        if metal in ('high_carbon_red_steel', 'high_carbon_steel', 'high_carbon_black_steel', 'high_carbon_blue_steel', 'wrought_iron'):
+            criteria.update(inventory_changed('tfc:metal/ingot/%s' % metal, name=metal))
+        else:
+            criteria.update(inventory_changed({'type': 'tfc:metal_item', 'metal': 'tfc:%s' % metal}, name=metal))
+    return criteria
```
alcatrazEscapee: Why does any of this exist - this seems like it would be easily solved by tags rather than this weird custom predicate type.

src/main/java/net/dries007/tfc/util/loot/CustomItemPredicate.java
```diff
@@ -0,0 +1,109 @@
...
+
+import net.dries007.tfc.TerraFirmaCraft;
+import net.dries007.tfc.common.recipes.HeatingRecipe;
+import net.dries007.tfc.common.recipes.inventory.ItemStackInventory;
+import net.dries007.tfc.util.Helpers;
+import net.dries007.tfc.util.JsonHelpers;
+import net.dries007.tfc.util.Metal;
+import org.jetbrains.annotations.Nullable;
+
+public class CustomItemPredicate extends ItemPredicate
```
alcatrazEscapee: I want to say this entire class is unnecessary. It is hardcoding what should really just be a tag, and it doesn't even look like it's an efficient or better version of just specifying a tag here.

alcatrazEscapee (changes requested):

alcatrazEscapee (approved):

PR #2078, opened 2022-09-03
Update ru_ru translation

Tak3aBreak:


eerussianguy:
Please update this, we have changed a lot of our translation keys to be more friendly for translators

alcatrazEscapee:
Merged manually in b29b6356d86665bc40566b631d3a5e560f0185ef

In lieu of an update here, I've culled any keys that were changed in recent updates introducing `%s` formatting parameters. If so so desire, please re-add these with their proper translation. Thanks!

PR #2097, opened 2022-09-09
Korean translation update

papertoy1127:


eerussianguy:
Please update this, we have changed a lot of our translation keys to be more friendly for translators

papertoy1127:
> Please update this, we have changed a lot of our translation keys to be more friendly for translators

Updated

PR #2102, opened 2022-09-10
loot table support

### LINKED ISSUE #2080: Using loot tables with generated TFC containers
Most TFC inventory containers could conceivably be populated by loot tables (large vessels, chests, etc.) Some requirements:

- need to be unpacked if we think the container may be accessed at all by a non-spectator (even an `isEmpty` check)
- need to be loaded/saved in the nbt of the BE
- after this, can only be used if either you use a custom structure pieces class that loads a builtin loot table, or you save a structure nbt file with a chest that you /setblock to have the loot table ID encoded

See `RandomizableContainerBlockEntity`

eerussianguy:
see InventoryBlockEntity for impl

closes #2080


src/main/java/net/dries007/tfc/client/ClientEventHandler.java
```diff
@@ -452,7 +453,7 @@ public static void registerColorHandlerBlocks(ColorHandlerEvent.Block event)
         TFCBlocks.WILD_CROPS.forEach((crop, reg) -> registry.register(grassColor, reg.get()));
 
         registry.register((state, level, pos, tintIndex) -> TFCColors.getWaterColor(pos), TFCBlocks.SALT_WATER.get(), TFCBlocks.SEA_ICE.get(), TFCBlocks.RIVER_WATER.get(), TFCBlocks.CAULDRONS.get(FluidType.SALT_WATER).get());
-        registry.register(blockColor(0x5FB5B8), TFCBlocks.SPRING_WATER.get(), TFCBlocks.CAULDRONS.get(FluidType.SPRING_WATER).get());
+        registry.register(springWater(new int[] {0xcc6600, 0xffc266, 0xffffb3, 0x5FB5B8}), TFCBlocks.SPRING_WATER.get(), TFCBlocks.CAULDRONS.get(FluidType.SPRING_WATER).get());
```
alcatrazEscapee: wat is this?

alcatrazEscapee (commented):
I have one major concern, which is how does this interact with `IItemHandler` ? i.e. if *anything* calls `getCapability()` and asks for an item handler, does that not need to call `unpackLootTable`?

For example, in vanilla, putting a hopper underneath a vessel with a loot table.
1. What does a chest do? I'm assuming it drains the contents?
2. What does a vessel do?

What happens if you pipe items into a loot table'd vessel with a hopper? I would think it just deletes the items? Or deletes the loot table?

PR #2125, opened 2022-09-14
IFarmland

eerussianguy:


src/main/java/net/dries007/tfc/common/blockentities/IFarmland.java
```diff
@@ -0,0 +1,121 @@
...
+    default void addNutrient(NutrientType type, float value)
+    {
+        setNutrient(type, getNutrient(type) + value);
+    }
+
+    default void addNutrients(Fertilizer fertilizer)
+    {
+        setNutrient(NITROGEN, Math.min(1, getNutrient(NITROGEN) + fertilizer.getNitrogen()));
+        setNutrient(PHOSPHOROUS, Math.min(1, getNutrient(PHOSPHOROUS) + fertilizer.getNitrogen()));
+        setNutrient(POTASSIUM, Math.min(1, getNutrient(POTASSIUM) + fertilizer.getNitrogen()));
```
alcatrazEscapee: `min` call is not necessary here as it's calling `setNutrient` which should be clamping the value anyway.

src/main/java/net/dries007/tfc/common/blockentities/IFarmland.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+        for (int i = 0; i < (int) (k > 0 ? Mth.clamp(k * 10, 1, 5) : 0); i++)
+        {
+            level.sendParticles(TFCParticles.POTASSIUM.get(), pos.getX() + level.random.nextFloat(), pos.getY() + level.random.nextFloat() / 5D, pos.getZ() + level.random.nextFloat(), 0, 0D, 0D, 0D, 1D);
+        }
+    }
+
+    float getNutrient(NutrientType type);
+
+    void setNutrient(NutrientType type, float value);
```
alcatrazEscapee: Document these (as they're now public API), and mention `getNutrient` returns a value in [0, 1]

alcatrazEscapee (changes requested):

alcatrazEscapee (approved):

PR #2129, opened 2022-09-16
Minor Jade fixes

### LINKED ISSUE #2114: Hydration and Temperature Info Displayed twice on Jade specifically with all Tall Bushes.

### LINKED ISSUE #2117: Incorrect information shown on Jade display
The gesteration period for (at least) goats and donkeys is a negative percentage rather than a positive percentage.

The "Becomes an adult in" display on baby animals doesn't update as time passes, it stays on the total grow time for that animal.

Ironcanon:
Fixes loom desync where client would be one 'progress' ahead of server (caused the progress counter on Jade to jump around). Haven't tested on server, only singleplayer, so might cause other issues that I'm unaware of.

Fixes #2117, I added one day to the display as otherwise it would show 0 days left on the day before the animal would grow up, if this is intended I will revert it.

Fixes #2114, As spreading bush extends staitionary bush it was getting the tooltips from both, if it needs a different tooltip from stationary bushes in the future a different fix will be needed.

Fixes pit klins tooltip displaying that all kilns always have 8 logs and straw.

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/LoomBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/LoomBlockEntity.java
@@ -57,7 +57,6 @@ public class LoomBlockEntity extends InventoryBlockEntity<ItemStackHandler>
                         loom.inventory.setStackInSlot(SLOT_RECIPE, ItemStack.EMPTY);
                         loom.inventory.setStackInSlot(SLOT_OUTPUT, recipe.assemble(new ItemStackInventory(loom.inventory.getStackInSlot(SLOT_RECIPE))));
                     }
-                    loom.markForSync();
                 }
             }
         }
--- a/src/main/java/net/dries007/tfc/common/blockentities/PitKilnBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/PitKilnBlockEntity.java
@@ -31,6 +31,8 @@ import net.dries007.tfc.config.TFCConfig;
 import net.dries007.tfc.util.Helpers;
 import net.dries007.tfc.util.calendar.Calendars;
 
+import java.util.function.Predicate;
+
 public class PitKilnBlockEntity extends PlacedItemBlockEntity
 {
     public static final Vec3i[] DIAGONALS = new Vec3i[] {new Vec3i(1, 0, 1), new Vec3i(-1, 0, 1), new Vec3i(1, 0, -1), new Vec3i(-1, 0, -1)};
@@ -272,6 +274,14 @@ public class PitKilnBlockEntity extends PlacedItemBlockEntity
     public NonNullList<ItemStack> getStraws()
     {
         return strawItems;
+    }
+
+    public Long getLogsCount() {
+        return logItems.stream().filter(Predicate.not(ItemStack::isEmpty)).count();
+    }
+
+    public Long getStrawsCount() {
+        return strawItems.stream().filter(Predicate.not(ItemStack::isEmpty)).count();
     }
 
     private void cookContents()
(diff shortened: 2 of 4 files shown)


src/main/java/net/dries007/tfc/common/blockentities/LoomBlockEntity.java
```diff
@@ -57,7 +57,6 @@ public static void tick(Level level, BlockPos pos, BlockState state, LoomBlockEn
                         loom.inventory.setStackInSlot(SLOT_RECIPE, ItemStack.EMPTY);
                         loom.inventory.setStackInSlot(SLOT_OUTPUT, recipe.assemble(new ItemStackInventory(loom.inventory.getStackInSlot(SLOT_RECIPE))));
                     }
-                    loom.markForSync();
```
alcatrazEscapee, line 60: How does *removing* a call to sync fix a sync issue? I'm also very suspicious that this is the right fix given you explicitly haven't tested it on dedicated server.

src/main/java/net/dries007/tfc/compat/jade/common/BlockEntityTooltips.java
```diff
@@ -79,7 +79,7 @@ public static void register(BiConsumer<BlockEntityTooltip, Class<? extends Block
         registerBlock.accept(HOE_OVERLAY, FruitTreeBranchBlock.class);
         registerBlock.accept(HOE_OVERLAY, FruitTreeLeavesBlock.class);
         registerBlock.accept(HOE_OVERLAY, StationaryBerryBushBlock.class);
-        registerBlock.accept(HOE_OVERLAY, SpreadingBushBlock.class);
+//        registerBlock.accept(HOE_OVERLAY, SpreadingBushBlock.class);
```
alcatrazEscapee, line 82: Don't comment out, if it's not meant to be here, remove.

src/main/java/net/dries007/tfc/compat/jade/common/BlockEntityTooltips.java
```diff
@@ -378,7 +378,7 @@ else if (pot.getOutput() != null && !pot.getOutput().isEmpty())
             final LoomRecipe recipe = loom.getRecipe();
             if (recipe != null)
             {
-                tooltip.accept(Helpers.translatable("tfc.jade.loom_progress", loom.getProgress(), recipe.getStepCount(), recipe.getResultItem()));
+                tooltip.accept(Helpers.translatable("tfc.jade.loom_progress", loom.getProgress(), recipe.getStepCount(), recipe.getResultItem().getDisplayName().getString()));
```
alcatrazEscapee, line 381:
```suggestion
                tooltip.accept(Helpers.translatable("tfc.jade.loom_progress", loom.getProgress(), recipe.getStepCount(), recipe.getResultItem().getDisplayName()));
```
`getString()` shouldn't be required, nor should it exist, as translation components get replaced inline.

src/main/java/net/dries007/tfc/common/blockentities/PitKilnBlockEntity.java
```diff
@@ -274,6 +276,14 @@ public NonNullList<ItemStack> getStraws()
         return strawItems;
     }
 
+    public Long getLogsCount() {
+        return logItems.stream().filter(Predicate.not(ItemStack::isEmpty)).count();
+    }
+
+    public Long getStrawsCount() {
+        return strawItems.stream().filter(Predicate.not(ItemStack::isEmpty)).count();
+    }
```
alcatrazEscapee, line 285:
```suggestion
    public int getLogCount()
    {
        return (int) logItems.stream().filter(Predicate.not(ItemStack::isEmpty)).count().orElse(0);
    }

    public int getStrawCount()
    {
        return (int) strawItems.stream().filter(Predicate.not(ItemStack::isEmpty)).count().orElse(0);
    }
```

- Proper English (`getLogsCount` -> `getLogCount`)
- Use our code style (braces on newlines)
- Don't make this return a nullable `Long` which is then silently unboxed later - it's never intended to be zero, return an `int`, and perform the safe unboxing here.

alcatrazEscapee (changes requested):

PR #2148, opened 2022-09-24
update and fix ru_ru translation

Tak3aBreak:


src/main/resources/assets/tfc/lang/ru_ru.json
```diff
@@ -1,6 +1,6 @@
 {
   "__comment1__": "This file was created by Taper4ik",
-  "__comment2__": "Russian Translation for version 1.18.2 TFC",
+  "__comment2__": "Russian Translation for version 1.18.2 TFC - v2.1.8-beta",
```
alcatrazEscapee, line 3: Please don't include the explicit version, as there's no process to keep this up-to-date.

PR #2164, opened 2022-09-29
Rebalanced and fixed tool damage and speed

lumagatto:
-More believable ("realistic-feeling") handling (just better attack speed values reflecting the tool) -Axes are no longer trash: Axes are weapons, not just tools, and since most players will use them as their main slashing weapon, it should not be annoying to use. -Scythes are no longer OP broken
-Javelins are not as OP as before, although they need a rework to stop being overpowered, which I, with my basic understanding of code, cannot do: 0.5 extra reach (and 1.4 attack speed to finally balance it)


alcatrazEscapee:
I don't understand some of your changes.

- Scythes have already been fixed by #2159 
- You say axes are no longer trash, but you're reducing their attack damage? What's the basis for these changes?
- Moreover, what's the basis for other not-intended-to-be-weapon damage changes? i.e. why does a chisel do more damage than a saw?

Moreover, you've changed the base values for metal tools but not changed the values for stone tools, which seems odd.

lumagatto:
>     * Scythes have already been fixed by [Dev bugfixin + Heat capacity rework #2159](https://github.com/TerraFirmaCraft/TerraFirmaCraft/pull/2159)
I only noticed afterwards, but i kept it because i honestly think that scythes shouldn't get such a massive nerf, they are still meant to be usable as weapons, if not for believability reasons, just for fun reasons (a lot of talk about death/grim reapers in the server)

>     * You say axes are no longer trash, but you're reducing their attack damage? What's the basis for these changes?
Attack speed was the problem. It made them annoying to use, while people are often used to just use axes unless they have a reason to use a slot of their inventory for a sword. While a sword is clearly better, axes needed not to be annoying to use, thus have 1 attack speed: this meant nerfing their attack damage, otherwise they would best maces and probably other weapons in most situations (that is, where no high tier high quality armor is involved).

>     * Moreover, what's the basis for other not-intended-to-be-weapon damage changes? i.e. why does a chisel do more damage than a saw?
Have you held either of those tools, or any of all these tools? I've went with my personal experience from using such tools and my personal HEMA experience, in rebalancing weapons and tools (This last being the reason of why im concerned about how weapons work) tldr: chisels are more dangerous than saws

>     *Moreover, you've changed the base values for metal tools but not changed the values for stone tools, which seems odd.
That is my mistake, from what i understood of the code that would have affected stone tools too, im gonna fix it now.

PR #2176, opened 2022-10-06
Fix being unable to pick up pot & grill

vasilevp:
#### The bug:
Sometimes the player is unable to crouch+right click on the firepit to pick up the attachment or receive damage.
#### Fix description:
The check `held == ItemStack.EMPTY` fails after installing a pot/grill onto firepit because in that case the player will hold a stack of `0 x AIR`, while `ItemStack.EMPTY` is `1 x AIR`. Other interactions that use `stack.split()` could also cause this to happen because they might also create a stack of 0 air in the player's hand.

I didn't see other uses of this comparison after a quick search, but it makes sense to double-check.


alcatrazEscapee:
Our current development branch is 1.18.x, and we are no longer supporting or maintaining 1.12.x, sorry.

PR #2193, opened 2022-10-19
Buildscript update

alcatrazEscapee:
- Update for new plugin syntax in FG5
- Fix JUnit not loading in gameTestServer, removes a bunch of redundant crap code.
- Tweak code style - just change the import rules for simplicity, and with the goal of allowing `import static junit.*` imports without these getting replaced with individual imports by auto reformat all the time.

A couple user facing bug fixes:

- Fixes vessel infinite expiry prolonging bug
- Add more recipes for boiling eggs
- Add missing aqueduct item tag.


build.gradle
```diff
@@ -84,29 +62,32 @@ repositories {
...
-            property 'forge.logging.console.level', 'debug'
-            property 'mixin.env.remapRefMap', 'true'
-            property 'mixin.env.refMapRemappingFile', "${projectDir}/build/createSrgToMcp/output.srg"
-            property 'forge.enabledGameTestNamespaces', mod_id
-            jvmArgs('-ea', '-Xmx4G', '-Xms4G')
+        all {
+            args("-mixin.config=${mod_id}.mixins.json")
+
+            property("forge.logging.console.level", "debug")
+            property("forge.enabledGameTestNamespaces", mod_id)
```
eerussianguy, line 77: should this be in all?

build.gradle
```diff
@@ -119,83 +100,57 @@ minecraft {
             }
         }
 
+        client {
+            workingDirectory project.file("run/client")
```
eerussianguy: why not just run?

alcatrazEscapee: Thought was so it's clear we have run/client, run/server, and run/gametest. i.e. any individual files belong to either client, server, or gametest (no files are just in run, not a priori knowing what process they came from)

build.gradle
alcatrazEscapee: Yes, to enable /test in run client / server. The difference between test server and gametest server is just that the latter is for CI.

PR #2195, opened 2022-10-21
Mini collapse rework.

alcatrazEscapee:
- Fix an outstanding visual issue with rock spike falling, where the whole spike didn't fall, only one part.
- New sounds for rock spike smoosh
- New sounds for collapses.
- Buff support beam vertical range to +2/-2
- "Fake Collapses" - when a normal collapse would've occurred, just throws some particles and a scary-but-not-too-scary sound. The particles come from blocks which could've started the collapse, so it's an early warning that those blocks are unsafe. ~30% of the time, a fake collapse will occur, and 10% of the time, a regular collapse will (unchanged).


src/main/java/net/dries007/tfc/common/recipes/CollapseRecipe.java
```diff
@@ -71,24 +70,50 @@ public static boolean tryTriggerCollapse(Level level, BlockPos pos)
...
                             level.playSound(null, pos, TFCSounds.ROCK_SLIDE_LONG.get(), SoundSource.BLOCKS, 1.0f, 1.0f);
                         }
                         return true; // Don't need to check other blocks, regardless of if we managed to collapse any blocks.
                     }
                 }
+
+                if (!fakeCollapseStarts.isEmpty())
+                {
+                    // Play sound
+                    level.playSound(null, pos, TFCSounds.ROCK_SLIDE_LONG_FAKE.get(), SoundSource.BLOCKS, 1.0f, 1.0f);
```
eerussianguy, line 104: Helpers#playSound

src/main/java/net/dries007/tfc/config/ServerConfig.java
```diff
@@ -16,7 +15,7 @@
 import net.dries007.tfc.config.animals.ProducingMammalConfig;
 import net.dries007.tfc.util.Alloy;
 
-import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
+import static net.dries007.tfc.TerraFirmaCraft.*;
```
eerussianguy, line 18: IDE is goofing around here?

src/main/java/net/dries007/tfc/common/blocks/rock/RockSpikeBlock.java
```diff
@@ -103,52 +130,40 @@ else if (belowState.isFaceSturdy(level, belowPos, Direction.UP))
...
-            return;
-        }
-        else if (aboveState.isFaceSturdy(level, abovePos, Direction.DOWN))
-        {
-            // Full block above, this is supported
-            return;
+            // No support below, try above
+            final BlockPos abovePos = pos.above();
+            final BlockState aboveState = level.getBlockState(abovePos);
+            if (aboveState.getBlock() == this && aboveState.getValue(PART).isLargerThan(state.getValue(PART)))
```
eerussianguy: shouldn't we make this work for spikes of mixed stone types?

src/main/java/net/dries007/tfc/common/blocks/rock/RockSpikeBlock.java
```diff
@@ -103,52 +130,40 @@ else if (belowState.isFaceSturdy(level, belowPos, Direction.UP))
...
+            {
+                // Larger spike above. Tick to ensure that it is supported
+                level.scheduleTick(abovePos, this, 1);
+                return;
+            }
+            else if (aboveState.isFaceSturdy(level, abovePos, Direction.DOWN))
```
eerussianguy, line 143: this seems weird because of how baby spikes don't need a full face.

src/main/java/net/dries007/tfc/common/entities/TFCFallingBlockEntity.java
```diff
@@ -153,8 +153,9 @@ public void tick()
                 else if (!level.isClientSide)
                 {
                     remove(RemovalReason.DISCARDED);
-                    return;
                 }
+                // If we spawn two falling block entities on the same tick, in adjacent positions, and the one above ticks first, it can cause a situation where the block below gets deleted by the falling block destruction code. This causes the next block entity to disappear. So, we don't do anything on first tick except capture and replace the block.
```
eerussianguy: \n

src/main/java/net/dries007/tfc/config/ServerConfig.java
alcatrazEscapee: This is a result of the import rule changes I made in 7df3754be0860c4760a57ddd0cc506a024be1256. I didn't apply it to the whole project (because ew + lazy) so this will just slowly standardize as files are edited. Essentially:
```java
import anything; // no wildcards

import tfc;

import static .*; 
```

I explicitly set wildcards to always/allowed because of using them for assertions i.e. `import static org.junit.Assertions.*`

src/main/java/net/dries007/tfc/common/blocks/rock/RockSpikeBlock.java
alcatrazEscapee: They do though? I'm not sure what you're talking about (this logic is also unchanged - just moved). There's not any functionality bug that I'm aware of, nor that the tests have caught.

src/main/java/net/dries007/tfc/common/recipes/CollapseRecipe.java
alcatrazEscapee: Switched the other instance but this is using specific volume + pitch, so no.

PR #2197, opened 2022-10-23
Shelves and data

### LINKED ISSUE #906: Off-hand reliant rock knapping.
**Describe the request**
Make sure to include:
Ability to knapp with two different kinds of rock, one in your main hand and one in your off hand, the one in your main hand would be the rock that gets consumed in the process.
A config to toggle whether two handed knapping is required would also be great. (similarly to off-handed hammers + chisels.)

**Meta Info**
 - TFC Version: 29.5.120 (at time of request)
TFC.
Nah.

### LINKED ISSUE #2190: All Predators aren't using their sleeping animation
They just use their standing animation to sleep, which has their legs clip through the floor while their eyes are wide open.

### LINKED ISSUE #2191: [1.18.2] Bow's recipe has its ingredients swapped
As per title
image

eerussianguy:
- Add chiseled bookshelves
- Update colorization of images to use HSV
- Add python asset validation for blockstates and models
- Fix extra seeds being dropped from some crops
- Fix dead stick crops not dropping sticks when needed
- Remove most unused textures and item models
- Fix double crops sometimes placing a fully mature dead crop when they should place an immature one
- Add mirrored bow recipe (Closes #2191)
- Fix predators not using sleeping anim (also fixed in entity branch) (Closes #2190)
- Firepits can take logs in from the environment
- Enchantment power now can use a ToFloatFunction<BlockState>
- Sapling growth config options, incl. individual tree configs and global modifiers
- Config option to require using both hands to knap rocks, default disabled (Closes #906)
- Fix up / standardize some things related to item entity handling in TFC devices
- Add missing heat capacity to clay block
- Tagify milk vinegar recipes
- Sound asset validation, add missing translations
- Fix incompatibility with create alloyed (fishing hook mixin)
- Bottom supported device block abstraction. Only actually changes behavior for looms, everything else was implemented that way already
- Fix nest boxes not always being placeable
- Global crop growth modifier
- Abstract Vector3f mulPose calls


resources/__main__.py
```diff
@@ -35,6 +36,7 @@ def main():
     parser.add_argument('actions', nargs='+', choices=(
         'clean',  # clean all resources (assets / data), including book
         'validate',  # validate no resources are changed when re-running
+        'validate_assets',  # manual validation for certain important resources
```
alcatrazEscapee, line 39: Add this as a validation step in CI [here](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/a8ef59ef7b89328607047c12132e6bac9afc7e6e/.github/workflows/test.yml#L46)?

resources/recipes.py
```diff
@@ -229,6 +229,7 @@ def plank(thing: str):
     rm.crafting_shaped('crafting/vanilla/bowl', ['XYX', ' X '], {'X': '#tfc:lumber', 'Y': 'tfc:glue'}, (3, 'minecraft:bowl')).with_advancement('#tfc:lumber')
     rm.crafting_shaped('crafting/vanilla/scaffolding', ['XYX', 'X X', 'X X'], {'X': 'minecraft:bamboo', 'Y': '#forge:string'}, (6, 'minecraft:scaffolding')).with_advancement('minecraft:bamboo')
     rm.crafting_shaped('crafting/vanilla/bow', [' YX', 'Y X', ' YX'], {'X': '#forge:rods/wooden', 'Y': '#forge:string'}, 'minecraft:bow').with_advancement('#forge:string')
+    rm.crafting_shaped('crafting/vanilla/bow2', ['XY ', 'X Y', 'XY '], {'X': '#forge:rods/wooden', 'Y': '#forge:string'}, 'minecraft:bow').with_advancement('#forge:string')
```
alcatrazEscapee: This is misunderstanding the issue. It is not one of mirroring (recipes will mirror themselves automatically.) The ingredients are swapped.

Vanilla
minecraft

TFC
tfc

resources/validate_assets.py
```diff
@@ -0,0 +1,102 @@
+from glob import glob
+import json
+from mcresources import utils
+
+assets_path = './src/main/resources/assets/'
+texture_forgiveness_paths = ('_fluff', 'block/burlap', 'block/molten_flow', 'yellow_bell', 'red_bell', 'green_bell', 'metal/full', 'plant/', 'sandstone/side', 'quiver', 'placed_item')
```
alcatrazEscapee: constants should be `UPPER_CAMEL_CASE`

resources/validate_assets.py
```diff
@@ -0,0 +1,102 @@
...
+
+def main():
+    model_locations = glob(assets_path + 'tfc/models/**/*.json', recursive=True)
+    state_locations = glob(assets_path + 'tfc/blockstates/**/*.json', recursive=True)
+    validate_model_parents(model_locations)
+    validate_textures(model_locations)
+    validate_blockstate_models(state_locations)
```
alcatrazEscapee: In order to use this in validation, all that should be necessary is for these to accumulate errors, and throw a single `AssertionError` if any errors were printed.

src/main/java/net/dries007/tfc/common/blocks/ExtendedProperties.java
```diff
@@ -141,9 +142,9 @@ public ExtendedProperties pathType(BlockPathTypes type)
         return this;
     }
 
-    public ExtendedProperties enchantmentPower(int power)
+    public ExtendedProperties enchantmentPower(ToFloatFunction<BlockState> powerGetter)
```
alcatrazEscapee: Can you add an overload for this this that takes a single `int` (for backwards binary compat)

src/main/java/net/dries007/tfc/common/blocks/crop/ClimbingCropBlock.java
```diff
@@ -99,8 +99,9 @@ public void die(Level level, BlockPos pos, BlockState state, boolean fullyGrown)
         final BlockPos posAbove = pos.above();
         final BlockState stateAbove = level.getBlockState(posAbove);
         final boolean hasTop = stateAbove.getBlock() == this;
-        final BlockState deadState = dead.get().defaultBlockState().setValue(DeadCropBlock.MATURE, fullyGrown || hasTop).setValue(STICK, state.getValue(STICK));
-        if (hasTop || (stateAbove.isAir() && state.getValue(STICK)))
+        final BlockState deadState = dead.get().defaultBlockState().setValue(DeadCropBlock.MATURE, fullyGrown).setValue(STICK, state.getValue(STICK));
+        // the dead immature version of the crop is NEVER two tall, so we cannot allow it to place the top block
```
alcatrazEscapee: Are you sure? Dead crop blocks with a stick placed will still have a stick on the top block, does this account for that?

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -505,9 +508,38 @@ public static void consumeInStackSizeIncrements(ItemStack stack, int totalCount,
...
+        final List<ItemEntity> availableItemEntities = new ArrayList<>();
+        int availableItems = 0;
+        int availableSlots = 0;
+        for (int slot = minSlotInclusive; slot <= maxSlotInclusive; slot++)
+        {
+            if (inventory.getStackInSlot(slot).isEmpty())
+            {
+                availableSlots++;
+            }
+        }
```
alcatrazEscapee: Since we do `insertSlots().isEmpty()`, we shouldn't have to do this "available slots" thing. Just pass in the available items, and `safelyConsumeItemsFromEntitiesIndividually` will stop as soon as it can.

src/main/java/net/dries007/tfc/util/InteractionManager.java
```diff
@@ -367,10 +367,11 @@ else if (!level.getBlockState(relativePos.below()).isAir())
         }
 
         // Knapping
+        final BiPredicate<ItemStack, Player> rockPredicate = (stack, player) -> TFCConfig.SERVER.requireOffhandForRockKnapping.get() ? Helpers.isItem(player.getMainHandItem(), TFCTags.Items.ROCK_KNAPPING) && Helpers.isItem(player.getOffhandItem(), TFCTags.Items.ROCK_KNAPPING) : stack.getCount() >= 2;
```
alcatrazEscapee: Honestly, I think having a config option to disable this is kind of dumb, but whatever. I *do* however, think this is executed incorrectly. Default behavior should be to allow off hand *or* two in one stack. As all it cares about is that you have two rocks. This config option (which was requested, so fine, even though I think it's pointless), should just disable one of those.

```java
(offhand && mainhand) || (!require && mainhand >= 2)
```

alcatrazEscapee (changes requested):

build.gradle
```diff
@@ -213,6 +213,8 @@ jar {
 mixin {
     add(sourceSets.main, "${mod_id}.refmap.json")
     add(sourceSets.test, "${mod_id}.refmap.json")
+
+    showMessageTypes=true
```
alcatrazEscapee: This is presumably just for debugging, and doesn't need to be here anymore.

resources/constants.py
```diff
@@ -933,8 +940,14 @@ def lang_enum(name: str, values: Sequence[str]) -> Dict[str, str]:
     'subtitles.entity.tfc.sabertooth.death': 'Sabertooth dies',
     'subtitles.entity.tfc.sabertooth.attack': 'Sabertooth roars',
     'subtitles.entity.tfc.sabertooth.ambient': 'Sabertooth calls',
-    'subtitles.entity.tfc.sabertooth.hurt': "Sabertooth yowls",
-    'subtitles.entity.tfc.sabertooth.sleep': "Sabertooth snores",
+    'subtitles.entity.tfc.sabertooth.hurt': 'Sabertooth yowls',
+    'subtitles.entity.tfc.sabertooth.sleep': 'Sabertooth snores',
+    'subtitles.entity.tfc.rooster.cry': 'Rooster calls',
+    'subtitles.generic.tfc.dirt_slide': 'Dirt landslides',
```
alcatrazEscapee:
```suggestion
    'subtitles.generic.tfc.dirt_slide': 'Soil landslides',
```
Since it's used for sand/gravel as well.

resources/validate_assets.py
```diff
@@ -0,0 +1,138 @@
...
+            if isinstance(textures, dict):
+                files_tested += 1
+                for texture in textures.values():
+                    if '#' not in texture:
+                        res = utils.resource_location(texture)
+                        if res.domain == 'tfc':
+                            tested += 1
+                            path = ASSETS_PATH + 'tfc/textures/%s.png' % res.path
+                            if len(glob(path)) == 0:
+                                print('Using missing texture, unable to load %s : java.io.FileNotFoundException: %s' % (f, path))
```
alcatrazEscapee: Why `java.io.FileNotFoundException` ?

src/main/java/net/dries007/tfc/common/blockentities/IFarmland.java
```diff
@@ -68,19 +68,35 @@ default void addNutrients(Fertilizer fertilizer)
...
         return 1 / 6f;
     }
 
+    /**
+     * @return A positive number to multiply the resupply amount of a crop block by.
+     */
+    default float resupplyModifier()
+    {
+        return 1f;
+    }
```
alcatrazEscapee: These are, as the comment indicates, way too precise a method to be worth having as they allow you to do all of exactly one thing, which you *could've already done* by overriding the target method. This kind of "one line helper method override for addons" is silly when we actually have to change and maintain our actual implementation.

Deprecate `resupplyFraction`, remove `resupplyModifier`. If an addon wants to override or change the behavior when a farmland tries to resupply nutrients, they can do so... by overriding the method that does that.

src/main/java/net/dries007/tfc/common/blockentities/IFarmland.java
```diff
@@ -68,19 +68,35 @@ default void addNutrients(Fertilizer fertilizer)
...
+    @Deprecated
+    default float consumeNutrientAndResupplyOthers(NutrientType type, float amount)
+    {
+        return consumeNutrientAndResupplyOthers(type, amount, 1 / 6f);
+    }
+
     /**
      * Consume up to {@code amount} of nutrient {@code type}.
-     * Resupplies other nutrient by default 1/6 of the amount consumed.
+     * Resupplies other nutrient by default what.
```
alcatrazEscapee: Revert

src/main/java/net/dries007/tfc/common/blockentities/IFarmland.java
```diff
@@ -68,19 +68,35 @@ default void addNutrients(Fertilizer fertilizer)
...
      * @return The amount of nutrient {@code type} that was actually consumed.
      */
-    default float consumeNutrientAndResupplyOthers(NutrientType type, float amount)
+    default float consumeNutrientAndResupplyOthers(NutrientType type, float amount, float resupplyFraction)
```
alcatrazEscapee: And this is a breaking change, despite the deprecation. No. Leave the method as is, inline `1/6`. This method can be overriden, or the callers overriden.

src/main/java/net/dries007/tfc/common/blocks/ExtendedProperties.java
```diff
@@ -141,9 +142,16 @@ public ExtendedProperties pathType(BlockPathTypes type)
         return this;
     }
 
+    @Deprecated
```
alcatrazEscapee: There's no reason for this to be deprecated. It's fine as a bouncer.

src/main/java/net/dries007/tfc/common/blocks/ExtendedProperties.java
```diff
@@ -141,9 +142,16 @@ public ExtendedProperties pathType(BlockPathTypes type)
...
     public ExtendedProperties enchantmentPower(int power)
     {
-        enchantmentPower = power;
+        enchantmentPower = s -> power;
+        return this;
```
alcatrazEscapee:
```suggestion
        return enchantmentPower(() -> power);
```

src/main/java/net/dries007/tfc/common/blocks/crop/ClimbingCropBlock.java
```diff
@@ -52,9 +53,9 @@ public IntegerProperty getAgeProperty()
         };
     }
 
-    protected ClimbingCropBlock(ExtendedProperties properties, int maxSingleAge, int maxAge, Supplier<? extends Block> dead, Supplier<? extends Item> seeds, FarmlandBlockEntity.NutrientType primaryNutrient, Supplier<ClimateRange> climateRange)
+    protected ClimbingCropBlock(ExtendedProperties properties, int maxSingleAge, int maxAge, RegistryCrop crop)
```
alcatrazEscapee: This is a breaking change again; why?

src/main/java/net/dries007/tfc/common/blocks/crop/Crop.java
```diff
@@ -69,38 +76,45 @@ private static ExtendedProperties dead()
     private final Supplier<Block> factory;
     private final Supplier<Block> deadFactory;
     private final Supplier<Block> wildFactory;
+    private final int growthTime;
+    private final int nutrientTime;
```
alcatrazEscapee: Why are we adding these parameters and not using them? What's the point if they're all constants?

src/main/java/net/dries007/tfc/common/blocks/crop/Crop.java
```diff
@@ -124,8 +138,65 @@ public Block createWild()
...
+    public FarmlandBlockEntity.NutrientType getDefaultPrimaryNutrient()
     {
         return primaryNutrient;
     }
+
+    @Override
+    public Supplier<FarmlandBlockEntity.NutrientType> getPrimaryNutrient()
+    {
+        return TFCConfig.SERVER.cropPrimaryNutrients.get(this);
+    }
```
alcatrazEscapee: Another pointless breaking change.

src/main/java/net/dries007/tfc/common/blocks/crop/Crop.java
```diff
@@ -11,6 +11,7 @@
 import java.util.function.Supplier;
```
alcatrazEscapee, line 11: This crop configurability just seems frankly a bit moronic.

- Per-crop nutrient type. Has this ever been requested? Why would someone need to do this when they can change fertilizers and other nutrients?
- Crop growth - A global modifier. Makes sense, given we have other calendar time related changes. Per-crop modifiers make *no* sense unless we're utilizing them ourselves, which we aren't.
- Nutrient time and resupply modifier are silly. The chance anyone editing the config file understands the mechanics of how nutrient consumption works and it's affect on crop growth is low (No, saying "this config option doesn't work like it sounds like it should" is not sufficient. And if they don't understand it beyond "nutrient makes crops faster" then why is it there. Or the resupply fraction, on a per-crop basis, just no.

src/main/java/net/dries007/tfc/common/blocks/crop/CropBlock.java
```diff
@@ -41,32 +40,39 @@
...
     protected final Supplier<ClimateRange> climateRange;
+    protected final Supplier<Integer> baseGrowthTime;
+    protected final Supplier<Integer> nutrientConsumptionTime;
+    protected final Supplier<Double> nutrientResupplyFactor;
 
     private final ExtendedProperties extendedProperties;
     private final int maxAge;
 
-    protected CropBlock(ExtendedProperties properties, int maxAge, Supplier<? extends Block> dead, Supplier<? extends Item> seeds, FarmlandBlockEntity.NutrientType primaryNutrient, Supplier<ClimateRange> climateRange)
+    protected CropBlock(ExtendedProperties properties, int maxAge, RegistryCrop crop)
```
alcatrazEscapee: More unnecessary breaking changes. Again, what is the motivation for this? Is there a reason? Because I don't see one.

src/main/java/net/dries007/tfc/common/blocks/crop/CropBlock.java
```diff
@@ -223,7 +232,25 @@ public ClimateRange getClimateRange()
...
+    }
+
+    @Override
+    public float getNutrientConsumptionFactor(Level level, BlockPos pos, BlockState state, CropBlockEntity crop)
+    {
+        return 1f / nutrientConsumptionTime.get();
+    }
+
+    @Override
+    public float getNutrientResupplyFactor(Level level, BlockPos pos, BlockState state, CropBlockEntity crop)
```
alcatrazEscapee: why in the everloving fuck would you need to know the level, position, block state, and entire block entity to know how much the nutrient could resupply by in one single method call when like I said previously... you could just override that one method.

src/main/java/net/dries007/tfc/common/blocks/crop/DeadClimbingCropBlock.java
```diff
@@ -6,21 +6,24 @@
...
 import net.dries007.tfc.common.blocks.ExtendedProperties;
 import net.dries007.tfc.common.blocks.TFCBlockStateProperties;
+import net.dries007.tfc.util.climate.ClimateRange;
 
 public class DeadClimbingCropBlock extends DeadDoubleCropBlock
 {
     public static final BooleanProperty STICK = TFCBlockStateProperties.STICK;
 
-    public DeadClimbingCropBlock(ExtendedProperties properties, Crop crop)
+    public DeadClimbingCropBlock(ExtendedProperties properties, Supplier<ClimateRange> range)
```
alcatrazEscapee: More breaking changes.

src/main/java/net/dries007/tfc/common/blocks/crop/DeadCropBlock.java
```diff
@@ -41,10 +41,10 @@ public class DeadCropBlock extends TFCBushBlock implements HoeOverlayBlock
 
     private final Supplier<ClimateRange> climateRange;
 
-    public DeadCropBlock(ExtendedProperties properties, Crop crop)
+    public DeadCropBlock(ExtendedProperties properties, Supplier<ClimateRange> range)
```
alcatrazEscapee: More breaking changes.

src/main/java/net/dries007/tfc/common/blocks/crop/DeadDoubleCropBlock.java
```diff
@@ -32,15 +33,16 @@
...
 import net.dries007.tfc.util.Helpers;
+import net.dries007.tfc.util.climate.ClimateRange;
 import org.jetbrains.annotations.Nullable;
 
 public class DeadDoubleCropBlock extends DeadCropBlock
 {
     public static final EnumProperty<DoubleCropBlock.Part> PART = TFCBlockStateProperties.DOUBLE_CROP_PART;
 
-    public DeadDoubleCropBlock(ExtendedProperties properties, Crop crop)
+    public DeadDoubleCropBlock(ExtendedProperties properties, Supplier<ClimateRange> range)
```
alcatrazEscapee: More breaking changes.

src/main/java/net/dries007/tfc/common/blocks/crop/DefaultCropBlock.java
```diff
@@ -39,9 +32,9 @@ public IntegerProperty getAgeProperty()
         };
     }
 
-    protected DefaultCropBlock(ExtendedProperties properties, int maxAge, Supplier<? extends Block> dead, Supplier<? extends Item> seeds, FarmlandBlockEntity.NutrientType primaryNutrient, Supplier<ClimateRange> climateRange)
+    protected DefaultCropBlock(ExtendedProperties properties, int maxAge, RegistryCrop crop)
```
alcatrazEscapee: More breaking changes.

src/main/java/net/dries007/tfc/common/blocks/crop/DoubleCropBlock.java
```diff
@@ -57,9 +51,9 @@ public IntegerProperty getAgeProperty()
     protected final int maxSingleAge;
     protected final float maxSingleGrowth;
 
-    protected DoubleCropBlock(ExtendedProperties properties, int maxSingleAge, int maxAge, Supplier<? extends Block> dead, Supplier<? extends Item> seeds, FarmlandBlockEntity.NutrientType primaryNutrient, Supplier<ClimateRange> climateRange)
+    protected DoubleCropBlock(ExtendedProperties properties, int maxSingleAge, int maxAge, RegistryCrop crop)
```
alcatrazEscapee: More breaking changes.

src/main/java/net/dries007/tfc/common/blocks/crop/FloodedCropBlock.java
```diff
@@ -49,9 +43,9 @@ public IntegerProperty getAgeProperty()
         };
     }
 
-    protected FloodedCropBlock(ExtendedProperties properties, int maxAge, Supplier<? extends Block> dead, Supplier<? extends Item> seeds, FarmlandBlockEntity.NutrientType primaryNutrient, Supplier<ClimateRange> climateRange)
+    protected FloodedCropBlock(ExtendedProperties properties, int maxAge, RegistryCrop crop)
```
alcatrazEscapee: More breaking changes.

src/main/java/net/dries007/tfc/common/blocks/crop/FloodedDeadCropBlock.java
```diff
@@ -23,14 +25,15 @@
...
 import net.dries007.tfc.common.fluids.IFluidLoggable;
 import net.dries007.tfc.util.Helpers;
+import net.dries007.tfc.util.climate.ClimateRange;
 
 public class FloodedDeadCropBlock extends DeadCropBlock implements IFluidLoggable
 {
     public static final FluidProperty FLUID = TFCBlockStateProperties.FRESH_WATER;
 
-    public FloodedDeadCropBlock(ExtendedProperties properties, Crop crop)
+    public FloodedDeadCropBlock(ExtendedProperties properties, Supplier<ClimateRange> range)
```
alcatrazEscapee: etc.

src/main/java/net/dries007/tfc/common/blocks/crop/ICropBlock.java
```diff
@@ -34,4 +37,10 @@ default float getGrowthLimit(Level level, BlockPos pos, BlockState state)
     ClimateRange getClimateRange();
 
     FarmlandBlockEntity.NutrientType getPrimaryNutrient();
+
+    float getGrowthFactor(Level level, BlockPos pos, BlockState state, CropBlockEntity crop);
+
+    float getNutrientConsumptionFactor(Level level, BlockPos pos, BlockState state, CropBlockEntity crop);
+
+    float getNutrientResupplyFactor(Level level, BlockPos pos, BlockState state, CropBlockEntity crop);
```
alcatrazEscapee: Even more breaking changes. These methods do not need to exist, they do not need the level of context given to them, it's ridiculous that they would be an API. They're completely undocumented and just... why.

src/main/java/net/dries007/tfc/common/items/RottenCompostItem.java
```diff
@@ -34,9 +34,9 @@ public InteractionResult useOn(UseOnContext context)
 
         if (state.getBlock() instanceof ICropBlock crop)
         {
-            if (!level.isClientSide)
+            if (!level.isClientSide && level.getBlockEntity(pos) instanceof CropBlockEntity be)
```
alcatrazEscapee: `entity`, `blockEntity`, `cropEntity`, just not `be` because that's a verb, not a `CropBlockEntity`.

src/main/java/net/dries007/tfc/config/ServerConfig.java
```diff
@@ -216,6 +234,7 @@
         ).define("enableForcedTFCGameRules", true);
         enableFireArrowSpreading = builder.apply("enableFireArrowSpreading").comment("Enable fire arrows and fireballs to spread fire and light blocks.").define("enableFireArrowSpreading", true);
         fireStarterChance = builder.apply("fireStarterChance").comment("Base probability for a firestarter to start a fire. May change based on circumstances").defineInRange("fireStarterChance", 0.5, 0, 1);
+        requireOffhandForRockKnapping = builder.apply("requireOffhandForRockKnapping").comment("If a rock is needed in the offhand in order to knap.").define("requireOffhandForRockKnapping", false);
```
alcatrazEscapee: This is badly worded.
```suggestion
        requireOffhandForRockKnapping = builder.apply("requireOffhandForRockKnapping").comment(
            "If true, knapping with rocks will only work when one rock is held in each hand (main hand and off hand)",
            "If false, knapping with rocks will work either with main and off hand, or by holding at least two rocks in the main hand"
        ).define("requireOffhandForRockKnapping", false);
```

src/main/java/net/dries007/tfc/config/ServerConfig.java
```diff
@@ -359,6 +378,43 @@
 
         hotWaterHealAmount = builder.apply("hotWaterHealAmount").comment("An amount that sitting in hot water will restore health, approximately twice per second.").defineInRange("hotWaterHealAmount", 0.08, 0.0, 20.0);
 
+        innerBuilder.pop().push("saplings");
+
+        globalSaplingGrowthModifier = builder.apply("globalSaplingGrowthModifier").comment("Modifier applied to the growth time of every (non-fruit) sapling").defineInRange("globalSaplingGrowthModifier", 1d, 0d, Double.MAX_VALUE);
+        globalFruitSaplingGrowthModifier = builder.apply("globalFruitSaplingGrowthModifier").comment("Modifier applied to the growth time of every fruit tree sapling").defineInRange("globalFruitSaplingGrowthModifier", 1d, 0d, Double.MAX_VALUE);
```
alcatrazEscapee: Neither of these specify if it gets larger or shorter based on modifier changes.

src/main/java/net/dries007/tfc/config/ServerConfig.java
```diff
@@ -359,6 +378,43 @@
...
+            final String valueName = String.format("%sSaplingGrowthDays", wood.getSerializedName());
+            saplingGrowthDays.put(wood, builder.apply(valueName).comment(String.format("Days for a %s tree sapling to be eligible to grow", wood.getSerializedName())).defineInRange(valueName, wood.defaultDaysToGrow(), 0, Integer.MAX_VALUE));
+        }
+        fruitSaplingGrowthDays = new EnumMap<>(FruitBlocks.Tree.class);
+        for (FruitBlocks.Tree tree : FruitBlocks.Tree.values())
+        {
+            final String valueName = String.format("%sSaplingGrowthDays", tree.getSerializedName());
+            fruitSaplingGrowthDays.put(tree, builder.apply(valueName).comment(String.format("Days for a %s tree sapling to be eligible to grow", tree.getSerializedName())).defineInRange(valueName, tree.defaultDaysToGrow(), 0, Integer.MAX_VALUE));
+        }
+        bananaSaplingGrowthDays = builder.apply("bananaSaplingGrowthDays").comment("Days for a banana tree sapling to be eligible to grow").defineInRange("bananaSaplingGrowthDays", 6, 0, Integer.MAX_VALUE);
```
alcatrazEscapee: "Eligible to grow" is a strange turn of phrase and does not explain what it means. Is it growth time? Average growth time? Does it need to be eligible multiple times before a tree grows?

Also why are these all measured in days (`growthDays`); where below the crop times (`growthTime`) are measured in ticks and all other devices are measured in ticks (`ticks`). Use the existing standard (`fooTicks`)

src/main/java/net/dries007/tfc/mixin/client/FishingHookRendererMixin.java
```diff
@@ -21,7 +21,7 @@
 @Mixin(FishingHookRenderer.class)
 public abstract class FishingHookRendererMixin
 {
-    @Redirect(method = "render(Lnet/minecraft/world/entity/projectile/FishingHook;FFLcom/mojang/blaze3d/vertex/PoseStack;Lnet/minecraft/client/renderer/MultiBufferSource;I)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/item/ItemStack;is(Lnet/minecraft/world/item/Item;)Z"))
+    @Redirect(method = "render(Lnet/minecraft/world/entity/projectile/FishingHook;FFLcom/mojang/blaze3d/vertex/PoseStack;Lnet/minecraft/client/renderer/MultiBufferSource;I)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/item/ItemStack;is(Lnet/minecraft/world/item/Item;)Z"), require=0)
```
alcatrazEscapee: Formatting, and please add a note explaining why require=0.

Also, why require=0 ?

alcatrazEscapee (changes requested):
Please re-import `CodeStyle.xml` as it has changed w.r.t import rules and this PR is still using old rules.

build.gradle
eerussianguy: Taking it out, but i have not fixed anything

resources/validate_assets.py
eerussianguy: joke. i remove

src/main/java/net/dries007/tfc/common/blocks/ExtendedProperties.java
eerussianguy: i guess, but it doesn't really demonstrate the type properly

src/main/java/net/dries007/tfc/common/blocks/crop/Crop.java
eerussianguy: these feed the config values

src/main/java/net/dries007/tfc/common/blocks/crop/CropBlock.java
eerussianguy: Ex if you wanted nutrients to not resupply as much in the dark

src/main/java/net/dries007/tfc/common/blocks/crop/Crop.java
eerussianguy: Gaelmare was asking for some of these, but I agree it's not all needed

src/main/java/net/dries007/tfc/config/ServerConfig.java
eerussianguy: eligible to grow = it is done counting ticks and needs a random tick to grow

eerussianguy: Not possible to change it from days now without ugly breaking changes to RegistryWood that would leave the constructor ambiguous as the type would still be int and supplier<int>

src/main/java/net/dries007/tfc/common/blocks/crop/CropBlock.java
alcatrazEscapee: so... override that one method. What if you want only phosphorus to only resupply as much in the dark? Then you couldn't do that with this, you'd need to override that one method.

src/main/java/net/dries007/tfc/common/blocks/crop/CropBlock.java
```diff
@@ -158,7 +160,10 @@ public void tick(BlockState state, ServerLevel level, BlockPos pos, Random rand)
         {
             if (canSurvive(state, level, pos))
             {
-                level.getBlockEntity(pos, TFCBlockEntities.CROP.get()).ifPresent(crop -> growthTick(level, pos, state, crop));
+                if (level.getBlockEntity(pos) instanceof CropBlockEntity crop)
+                {
+                    growthTick(level, pos, state, crop);
+                }
```
alcatrazEscapee, line 166: As a thought - instead of switching all these optional usages to `instanceof` - to allow addons with their own block entity types, could we prefer instead just replacing the explicit block entity type with `getExtendedProperties().blockEntity()` ? It's a bit neater, and it preserves the intended use (restricting based on the block entity type)

I think that still solves the issue at hand and doesn't require refactoring other usages of `getBlockEntity` as much.

src/main/java/net/dries007/tfc/common/blocks/crop/DeadClimbingCropBlock.java
```diff
@@ -6,21 +6,28 @@
...
 
     public DeadClimbingCropBlock(ExtendedProperties properties, Crop crop)
     {
-        super(properties, crop);
+        this(properties, crop.getClimateRange());
+    }
+
+    public DeadClimbingCropBlock(ExtendedProperties properties, Supplier<ClimateRange> range)
+    {
+        super(properties, range);
```
alcatrazEscapee, line 30: Do dead climbing crops actually require a climate range? It makes more sense to me intuitively to pass in `() -> ClimateRange.NOOP` to the super constructor here, as it's explicitly intended to be unused. (See `DeadBerryBushBlock`)

src/main/java/net/dries007/tfc/common/blocks/devices/BottomSupportedDeviceBlock.java
```diff
@@ -0,0 +1,72 @@
...
+    public BottomSupportedDeviceBlock(ExtendedProperties properties, InventoryRemoveBehavior removeBehavior)
+    {
+        this(properties, removeBehavior, Shapes.block());
+    }
+
+    public BottomSupportedDeviceBlock(ExtendedProperties properties, InventoryRemoveBehavior removeBehavior, VoxelShape shape)
+    {
+        super(properties, removeBehavior);
+        this.shape = shape;
+    }
```
alcatrazEscapee:
```suggestion
    protected BottomSupportedDeviceBlock(ExtendedProperties properties, InventoryRemoveBehavior removeBehavior)
    {
        this(properties, removeBehavior, Shapes.block());
    }

    protected BottomSupportedDeviceBlock(ExtendedProperties properties, InventoryRemoveBehavior removeBehavior, VoxelShape shape)
    {
        super(properties, removeBehavior);
        this.shape = shape;
    }
```

src/main/java/net/dries007/tfc/common/blocks/wood/TFCSaplingBlock.java
```diff
@@ -31,9 +32,14 @@
 public class TFCSaplingBlock extends SaplingBlock implements IForgeBlockExtension, EntityBlockExtension
 {
     private final ExtendedProperties properties;
-    private final int daysToGrow;
+    private final Supplier<Integer> daysToGrow;
```
alcatrazEscapee, line 35:
```suggestion
    private final IntSupplier daysToGrow;
```

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/common/blocks/wood/TFCSaplingBlock.java
eerussianguy: this is incompatible with config values provided by forge

src/main/java/net/dries007/tfc/common/blocks/crop/DeadClimbingCropBlock.java
eerussianguy: you need it for the hoe overlay so you can see why it didn't grow

src/main/java/net/dries007/tfc/common/blocks/crop/CropBlock.java
eerussianguy: Ship may have sailed on this for some cases, but sure

alcatrazEscapee: Not saying we need to refactor every usage we currently have, for sure. But for cases like these, and for new code - we can still use the block entity type that already exists on the block.

src/main/java/net/dries007/tfc/common/blocks/crop/DeadClimbingCropBlock.java
alcatrazEscapee: Makes sense.

src/main/java/net/dries007/tfc/common/blocks/wood/TFCSaplingBlock.java
alcatrazEscapee: Yeah, but you already coerce it by passing in `this::daysToGrow`. You should be able to change the type of this without issue.

In either case, it's not a huge issue and is fine to leave as-is. Also lets addons pass in `configValue` instead of `configValue::get`. So sure, why not.

alcatrazEscapee (approved):

PR #2205, opened 2022-10-29
More translation: zh_cn 100% finished, also add zh_tw

recolic:


recolic:
basically I did 3 things in this PR: 

1. modify `zh_cn.json` and finish all translation
2. convert `zh_cn.json` to `zh_tw.json` with OpenCC chinese conversion. I have checked the result text, and the quality is acceptable. 
3. run `python resources book --translate zh_cn` and `python resources book --translate zh_tw` to generate books

PR #2215, opened 2022-11-02
Fix problem 'patchouli.entity is not translated', add chinese tranaslation for all new entries

recolic:
This PR solves two problems: 

1. In `generate_books.py`, all `entity`s are not translated to local language. This causes many untranslated texts in our field guide.
2. After fixing problem 1, there are more entries to translate. I also finished the translation for `zh_cn` and `zh_tw`, and re-generated all books.


alcatrazEscapee (approved):
Well spotted.

alcatrazEscapee:
@recolic Looking at the web edition of the book, there's [another localization file](https://github.com/TerraFirmaCraft/Field-Guide/blob/main/assets/lang/en_us.json) for specific web translation keys, if you want to do a pass over that as well. Otherwise I'll try and get to updating the book website along with the next release.

recolic:
@alcatrazEscapee wait for a minute. I will update that translation key, so the translation will be included in the web guide.

recolic:
I have raised a PR <https://github.com/TerraFirmaCraft/Field-Guide/pull/7> to update the web edition of the book. 
Thanks for your review! 

> @recolic Looking at the web edition of the book, there's [another localization file](https://github.com/TerraFirmaCraft/Field-Guide/blob/main/assets/lang/en_us.json) for specific web translation keys, if you want to do a pass over that as well. Otherwise I'll try and get to updating the book website along with the next release.

PR #2226, opened 2022-11-13
Adjust rarity_filter value basing on my survival world test

recolic:
The new rarity value gives a similar survival experience with traditional TFC 1.7.10 & TFC-TNG 1.12.2. The default rarity value is generating too much.

[Please use `squash and merge` because there are many meaningless commits for testing]


recolic:
I will solve the build issue and reopen the PR after that.

recolic:
Build issue fixed. Thanks

alcatrazEscapee:
No. Being strictly a balance change; and one that can be accomplished entirely by datapacks, this needs a lot more justification than both the claim that it's more similar, and/or the fact it is more similar. Neither 1.12 or 1.7.10 can have apples-to-apples comparisons made with 1.18 due to the environment change. We tend to solicit feedback about balance and rarity in our discord, as a lot of it can be quite subjective and based on anecdotes.

Issue #905, opened 2020-03-22
Config to disable the automate-ability of TNG containers.


UnlimatedStone9:
**Describe the request**
Make sure to include:
1. What problem or issue would this enhancement solve?
Modpack configurability.
2. Describe the feature itself - what would change, what would happen. In general more detail is good!
It would be a config to disable the ability to filter items into containers like crucibles/barrels/vessels etc with "machines" like pipes, hoppers, filters, etc.
I've seen other mods allow this function, like PyroTech's rock bags for example.
This would allow restricting automating some TFC mechanics to incentivize making more modern systems (via other mods), or encourage using improved storage systems for automatic storage, rather than basic wooden chests or pottery.

**Meta Info**
 - TFC Version: 1.12.2-29.5.120 (at time of request)
 - Is this a compatibility feature for other mods? Or a enhancement for TFC alone?
Mainly for modpack integration with other mods. Low priority but very useful.
 - If necessary, what other mods (including versions) would this require?
Nah.

PR #2262, opened 2022-12-18
Comparator and Redstone interactivity

### LINKED ISSUE #2079: Comparator and Redstone interactivity.
As title, I don't believe pretty much any of our devices implement comparator outputs, at all. Where they probably all could.

- Basic inventories should show total filled / total capacity
- Fluid inventories, crucible, barrel, etc. should show fluid inventory.

In addition, we could consider adding some *limited* redstone control to devices, where it makes sense that automation exists (it's a stretch but hey):

- Dispenser + Flint and Steel interact with TFC devices, i.e. lighting them.

Another thing to consider: it might be worth while, developing some sort of in-dev test against a typical pipe based automation mod, to verify our containers all interact in expected ways (we can do item based tests with hoppers, but that also doesn't fully compare since hoppers are vanilla and mods will interact in different ways). Stuff like verifying the sidedness of our interactions. (Also, sync issues, because I uncovered one just by implementing barrel + bucket + dispenser interactions)

Charnuz:
This covers most of #2079, with the goal of comparators/dispensers giving players QoL improvements. This would allow players to automate most barrel recipes as well.

Fixes #2079


src/main/java/net/dries007/tfc/common/blockentities/ToolRackBlockEntity.java
```diff
@@ -90,4 +89,19 @@ private void insertItem(int slot, ItemStack stack)
         level.playSound(null, worldPosition, TFCSounds.TOOL_RACK_PLACE.get(), SoundSource.BLOCKS, 1, 1 + ((level.random.nextFloat() - level.random.nextFloat()) / 16));
     }
 
+    // Sorry for this.
+    // This returns a unique analog value for every permutation of items within slots
+    @Override
+    public int getAnalogValue()
```
alcatrazEscapee, line 95: This just seems unintuitive to me, imo. Yes, in theory, people could use this to detect which element in a tool rack is present, but it makes much more sense to just output a count of the number of items in the tool rack.

src/main/java/net/dries007/tfc/common/blocks/devices/BarrelBlock.java
```diff
@@ -90,4 +92,25 @@ protected void addExtraInfo(List<Component> tooltip, CompoundTag inventoryTag)
...
+    }
+
+    @Override
+    public void neighborChanged(BlockState state, Level level, BlockPos pos, Block block, BlockPos neighborPos, boolean s)
+    {
+        if (!level.isClientSide && level.hasNeighborSignal(pos))
+        {
+            toggleSeal(level, pos, state);
+        }
+    }
```
alcatrazEscapee, line 115: Should barrels (which are pretty much definitionally a big pile o wood with no moving parts), actually react themselves to redstone signals? I'm on the fence for this, because I think comparator outputs are good and can be justified by "magic comparator block doing stuff" but idk, can "magic dust block doing stuff" also justify this?

@eerussianguy thoughts?

If it's contentious we could also add a config option for this doing anything (see #905)

src/main/java/net/dries007/tfc/common/blocks/devices/LogPileBlock.java
```diff
@@ -51,7 +51,7 @@ public BlockState getStateForPlacement(BlockPlaceContext context)
     {
         return defaultBlockState().setValue(AXIS, context.getHorizontalDirection().getAxis());
     }
-
+
```
alcatrazEscapee, line 54: Revert pls.

src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
```diff
@@ -121,4 +122,25 @@ public boolean canDropFromExplosion(BlockState state, BlockGetter level, BlockPo
...
+    {
+        return true;
+    }
+
+    @Override
+    public void neighborChanged(BlockState state, Level level, BlockPos pos, Block block, BlockPos neighborPos, boolean s)
+    {
+        if (!level.isClientSide && level.hasNeighborSignal(pos))
+        {
+            toggleSeal(level, pos, state);
```
alcatrazEscapee, line 143: Moreover, if we do add config options, one per device ("barrel reacts to redstone", "powderkeg reacts to redstone") I think would be good. For some reason I think the comparator output is a fairly no-brainer but the redstone input I could see being controversial.

src/main/java/net/dries007/tfc/common/blocks/devices/TFCComposterBlock.java
```diff
@@ -131,6 +131,18 @@ public boolean isRandomlyTicking(BlockState state)
         return state.getValue(TYPE) != CompostType.ROTTEN;
     }
 
+    @Override
+    public int getAnalogOutputSignal(BlockState state, Level level, BlockPos pos)
+    {
+        return state.getValue(TYPE).ordinal();
```
alcatrazEscapee, line 137: imo this should output `STAGE`, not `TYPE` ? I understand that makes it harder to automate, but it *is* more inline with what comparators do - measure the full-ness of things.

src/main/java/net/dries007/tfc/config/ServerConfig.java
```diff
@@ -365,6 +367,10 @@
         enablePlacingItems = builder.apply("enablePlacingItems").comment("If true, players can place items on the ground with V.").define("enablePlacingItems", true);
         usePlacedItemWhitelist = builder.apply("usePlacedItemWhitelist").comment("If true, the tag 'tfc:placed_item_whitelist' will be checked to allow items to be in placed items and will exclude everything else.").define("usePlacedItemWhitelist", false);
 
+        innerBuilder.pop().push("dispenser");
+
+        dispenserEnableLighting = builder.apply("dispenserEnableAutomation").comment("If true, dispensers will be able to light blocks.").define("dispenserEnableAutomation", true);
```
alcatrazEscapee, line 372: config value `dispenserEnableLighting` should match the name `dispenserEnableAutomation`

src/main/java/net/dries007/tfc/mixin/accessor/DispenserBlockAccessor.java
```diff
@@ -0,0 +1,21 @@
...
+import net.minecraft.world.item.Item;
+import net.minecraft.world.level.block.DispenserBlock;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(DispenserBlock.class)
+public interface DispenserBlockAccessor
+{
+    @Accessor("DISPENSER_REGISTRY")
+    static Map<Item, DispenseItemBehavior> accessor$getDispenserRegistry() {throw new AssertionError();}
```
alcatrazEscapee:
```suggestion
    static Map<Item, DispenseItemBehavior> accessor$getDispenserRegistry() { throw new AssertionError(); }
```

src/main/java/net/dries007/tfc/util/DispenserBehaviors.java
```diff
@@ -152,6 +182,50 @@ protected void playSound(BlockSource source)
...
+            final BlockPos pos = source.getPos().relative(source.getBlockState().getValue(DispenserBlock.FACING));
+            if (Helpers.isItem(stack, TFCItems.HANDSTONE.get()) && level.getBlockState(pos).getBlock() instanceof QuernBlock)
+            {
+                return level.getBlockEntity(pos, TFCBlockEntities.QUERN.get()).filter(quern -> !quern.hasHandstone()).flatMap(quern ->
+                    quern.getCapability(Capabilities.ITEM).map(item -> {
+                            item.insertItem(QuernBlockEntity.SLOT_HANDSTONE, stack, false);
+                            return ItemStack.EMPTY;
+                        }
+
+                    )
```
alcatrazEscapee:
```suggestion
                    quern.getCapability(Capabilities.ITEM).map(item -> {
                        item.insertItem(QuernBlockEntity.SLOT_HANDSTONE, stack, false);
                        return ItemStack.EMPTY;
                    })
```
formatting is a bit off

PR #2271, opened 2022-12-27
Add .rotate() and .mirror() methods to directional blocks

natrow:
Implements missing rotation and mirror methods on several blocks including:

 - Sheet piles
 - Anvils
 - Bloomery
 - Sluices
 - Creeping plants
 - Epiphyte plants
 - Spreading bushes
 - Giant kelp trees & flowers
 - Aqueducts
 - Bookshelves
 - Bellows
 - Fruit tree branches
 - Looms
 - Tool Racks

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/SheetPileBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/SheetPileBlockEntity.java
@@ -7,7 +7,6 @@
 package net.dries007.tfc.common.blockentities;
 
 import java.util.Arrays;
-
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
 import net.minecraft.nbt.CompoundTag;
@@ -32,7 +31,11 @@ public class SheetPileBlockEntity extends TFCBlockEntity
     private static final BooleanProperty MIRROR = SheetPileBlock.MIRROR;
 
     /**
-     * Apply internal rotation & mirroring to map requested face to an inventory slot.
+     * Sheet piles use a seperate rotation + mirror block state property to map their states (up, down, left, etc.) to an index.
+     * We do this as we want to be able to mirror and rotate the block (which only provide a state, not a block entity).
+     * So, this method takes the state and a requested direction, and converts it to a index into the (un-rotated) {@code stacks} array.
+     * <p>
+     * A mirror of {@code Mirror.NONE} and a rotation of {@code Direction.NORTH} is considered the reference frame of no-rotation, i.e. where the {@code stacks} is indexed by direction ordinal.
      */
     private int faceToIndex(Direction face)
     {
--- a/src/main/java/net/dries007/tfc/common/blockentities/SheetPileBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/SheetPileBlockEntity.java
@@ -9,7 +9,6 @@ package net.dries007.tfc.common.blockentities;
 import java.util.Arrays;
 
 import net.minecraft.core.BlockPos;
-import net.minecraft.core.Direction;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.nbt.Tag;
 import net.minecraft.world.item.ItemStack;
@@ -33,17 +32,15 @@ public class SheetPileBlockEntity extends TFCBlockEntity
         Arrays.fill(stacks, ItemStack.EMPTY);
     }
 
-    public void addSheet(Direction direction, ItemStack stack)
+    public void addSheet(int index, ItemStack stack)
     {
-        final int index = direction.ordinal();
         stacks[index] = stack;
         cachedMetals[index] = null;
         markForSync();
     }
 
-    public ItemStack removeSheet(Direction direction)
+    public ItemStack removeSheet(int index)
     {
-        final int index = direction.ordinal();
         final ItemStack stack = stacks[index];
         stacks[index] = ItemStack.EMPTY;
         cachedMetals[index] = null;
@@ -51,9 +48,9 @@ public class SheetPileBlockEntity extends TFCBlockEntity
         return stack;
     }
 
-    public ItemStack getSheet(Direction direction)
+    public ItemStack getSheet(int index)
     {
-        return stacks[direction.ordinal()].copy();
+        return stacks[index].copy();
     }
 
     /**
@@ -61,9 +58,8 @@ public class SheetPileBlockEntity extends TFCBlockEntity
      * The metal is defined by checking what metal the stack would melt into if heated.
      * Any other items turn into {@link Metal#unknown()}.
      */
-    public Metal getOrCacheMetal(Direction direction)
+    public Metal getOrCacheMetal(int index)
     {
-        final int index = direction.ordinal();
         final ItemStack stack = stacks[index];
 
         Metal metal = cachedMetals[index];
--- a/src/main/java/net/dries007/tfc/common/blockentities/SheetPileBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/SheetPileBlockEntity.java
@@ -9,11 +9,17 @@ package net.dries007.tfc.common.blockentities;
 import java.util.Arrays;
 
 import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.nbt.Tag;
 import net.minecraft.world.item.ItemStack;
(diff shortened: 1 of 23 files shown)


src/main/java/net/dries007/tfc/common/blockentities/SheetPileBlockEntity.java
```diff
@@ -33,37 +32,34 @@ public SheetPileBlockEntity(BlockPos pos, BlockState state)
         Arrays.fill(stacks, ItemStack.EMPTY);
     }
 
-    public void addSheet(Direction direction, ItemStack stack)
+    public void addSheet(int index, ItemStack stack)
```
alcatrazEscapee: I don't like exposing an `int index` here without any context to what it means. The sheet pile should still have the same interface as before: take a direction, add the sheet for that direction. It should be able to infer the index based on the rotation.

Also, this would mean this doesn't have to be a binary incompatible change, which is always nice.

src/main/java/net/dries007/tfc/common/blocks/devices/SheetPileBlock.java
```diff
@@ -60,15 +61,31 @@ public class SheetPileBlock extends ExtendedBlock implements EntityBlockExtensio
         .put(DOWN, box(0, 0, 0, 16, 1, 16))
         .build();
 
-    public static void removeSheet(Level level, BlockPos pos, BlockState state, Direction face, @Nullable Player player, boolean doDrops)
-    {
+
+    public static int faceToIndex(BlockState state, Direction face) {
```
alcatrazEscapee: Please add a javadoc explaining the purpose of this method and why it's necessary, as it's not intuitive why we need a rotation + mirror saved here at all (except once you consider the effects of `rotate()` and `mirror()`.

src/main/java/net/dries007/tfc/common/blocks/devices/SheetPileBlock.java
```diff
@@ -257,15 +276,52 @@ public boolean onDestroyedByPlayer(BlockState state, Level level, BlockPos pos,
...
+                .setValue(SOUTH, state.getValue(NORTH))
+                .setValue(FACING, mirror.mirror(state.getValue(FACING)))
+                .cycle(MIRROR);
+            case FRONT_BACK -> state.setValue(EAST, state.getValue(WEST))
+                .setValue(WEST, state.getValue(EAST))
+                .setValue(FACING, mirror.mirror(state.getValue(FACING)))
+                .cycle(MIRROR);
+            default -> state;
+        };
+    }
```
alcatrazEscapee: For these methods (ones which rotate `NORTH`, `SOUTH`, `EAST`, and `WEST`, and on blocks that inherit `DirectionPropertyBlock`, (`SheetPileBlock`, `CreepingPlantBlock`, also for `KelpTreeBlock`), could you delegate these to a static method in `DirectionPropertyBlock` ? As it's a fair amount of duplicated code otherwise.

src/main/java/net/dries007/tfc/common/blocks/plant/CreepingPlantBlock.java
```diff
@@ -148,4 +146,28 @@ private boolean isEmpty(BlockState state)
...
+        };
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public BlockState mirror(BlockState state, Mirror mirror) {
+        return switch (mirror) {
+            case LEFT_RIGHT -> state.setValue(NORTH, state.getValue(SOUTH)).setValue(SOUTH, state.getValue(NORTH));
+            case FRONT_BACK -> state.setValue(EAST, state.getValue(WEST)).setValue(WEST, state.getValue(EAST));
+            default -> super.mirror(state, mirror);
```
alcatrazEscapee: Why `super` here but not above? Consistency would assume using `state` as it's the most clear about behavior.

src/main/java/net/dries007/tfc/common/blocks/rock/AqueductBlock.java
```diff
@@ -39,12 +38,13 @@
 import org.jetbrains.annotations.Nullable;
 
 
-public class AqueductBlock extends HorizontalDirectionalBlock implements IFluidLoggable
+public class AqueductBlock extends Block implements IFluidLoggable
```
alcatrazEscapee: Please revert, this serves no purpose and is binary compat breaking.

src/main/java/net/dries007/tfc/common/blocks/rock/AqueductBlock.java
```diff
@@ -323,4 +323,41 @@ else if (isValidSource(adjacentState) && getFluidProperty().canContain(adjacentF
...
+        return switch (mirror) {
+            case LEFT_RIGHT -> state.setValue(NORTH, state.getValue(SOUTH))
+                .setValue(SOUTH, state.getValue(NORTH))
+                .setValue(FACING, mirror.mirror(state.getValue(FACING)));
+            case FRONT_BACK -> state.setValue(EAST, state.getValue(WEST))
+                .setValue(WEST, state.getValue(EAST))
+                .setValue(FACING, mirror.mirror(state.getValue(FACING)));
+            default -> super.mirror(state, mirror);
+        };
+    }
```
alcatrazEscapee: This could be done with two calls - one to a static method in `DirectionPropertyBlock` and one to `super`.

alcatrazEscapee (changes requested):
Please reformat to use our code style (mostly braces on newlines, import order, no wildcard imports).

src/main/java/net/dries007/tfc/common/blocks/rock/AqueductBlock.java
natrow: This was changed because the only 2 methods in HorizontalDirectionalBlock are being overwritten. And it was trying to override the methods from the HorizontalDirectionalBlock class rather than Block which was doing weird stuff with the deprecation warnings.

src/main/java/net/dries007/tfc/common/blockentities/SheetPileBlockEntity.java
natrow: In order to apply the rotation and mirroring on the block to convert the Direction to the index (as I'm doing in `faceToIndex`), the method would need access to the BlockState.

One option would be to change `int faceToIndex(BlockState, Direction)` to something like `Direction applyRotationAndMirror(BlockState, Direction)`, then input the result to `addSheet`. Personally, I think that's even more confusing and doesn't signify to the caller that this functionality has been changed.

Another option is to change the parameter to `int stackIndex` and rename `faceToIndex()` to `mapFaceToStackIndex()` to make this relationship more obvious.

alcatrazEscapee: >  the method would need access to the BlockState.

It has access, it can call `getBlockState()` in the method.

src/main/java/net/dries007/tfc/common/blocks/devices/SheetPileBlock.java
```diff
@@ -268,4 +300,22 @@ public VoxelShape getShape(BlockState state, BlockGetter level, BlockPos pos, Co
...
+    {
+        return DirectionPropertyBlock.rotate(state, rot).setValue(FACING, rot.rotate(state.getValue(FACING)));
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public BlockState mirror(BlockState state, Mirror mirror)
+    {
+        if (mirror == Mirror.NONE) {
+            return state; // don't flip MIRROR bit
```
alcatrazEscapee:
```suggestion
        if (mirror == Mirror.NONE)
        {
            return state; // don't flip MIRROR bit
```

alcatrazEscapee (changes requested):
One missed formatting, and please address the remaining outstanding comments in re: `SheetPileBlockEntity`

natrow:
I had no idea the getBlockState() method existed - that makes things so much simpler! Everything should be ready to go now.

By the way, is there an automated code reformatter set up in the CI/CD pipeline? For some reason, the default settings in my Intellij project were all messed up.

natrow:
> I had no idea the getBlockState() method existed - that makes things so much simpler! Everything should be ready to go now.
> 
> By the way, is there an automated code reformatter set up in the CI/CD pipeline? For some reason, the default settings in my Intellij project were all messed up.

Oh I'm a dummy, I didn't even realize you have to import the CodeStyle.xml first. But even so, it seems like a lot of classes haven't been formatted with that...

alcatrazEscapee (approved):

PR #2286, opened 2023-01-17
 Lang Updates

neighborsbear:
The translation of high carbon blue steel was incorrectly translated as black steel. The translation of black bronze is translated as the name of another alloy.


alcatrazEscapee:
cc @papertoy1127

Issue #2270, opened 2022-12-24
If the XP bar is disabled, move health, hunger, etc. down.


MagmaSlime123:
Title. Not sure why there's a big blank space.

Bars are from a resourcepack being used from the Content channel.
![2022-12-23_20 18 56](https://user-images.githubusercontent.com/51277766/209419460-327a16b7-1c6c-4c5a-8c10-9ade077f5414.png)

eerussianguy:
#2316

PR #2316, opened 2023-02-20
An Additional Config for Disabled XP Bar

Notenoughmail:
This implements a new config option, `disabledExperienceBarStyle`, which changes how various HUD elements are displayed when `enableExperienceBar` is set to `false`.

It has three options:
- `HOVER`: The default option, identical to current behavior.
- `BUMP`: Moves the health, food, and water bars closer to the hot bar unless the fishing or jumping meter are present, in which case they will be moved to their default positions.
- `LEFT_HOTBAR`: Moves the health, food, and water bars closer to the hot bar. The fishing and jumping meters become smaller, vertical bars placed between the offhand slot and the rest of the hot bar.

This seeks to address #2270.


src/main/java/net/dries007/tfc/client/IngameOverlays.java
```diff
@@ -312,4 +375,31 @@ public static boolean setup(ForgeIngameGui gui, Minecraft minecraft)
...
+                        heightOut = heightIn;
+                    }
+                    else
+                    {
+                        heightOut = heightIn + 6;
+                    }
+                }
+                default -> heightOut = heightIn;
+            }
+            return heightOut;
```
alcatrazEscapee: This can be quite simplified using a switch expression:
```suggestion
            return switch (TFCConfig.CLIENT.disabledExperienceBarStyle.get())
            {
                case LEFT_HOTBAR -> heightIn + 6;
                case BUMP ->player != null && (player.fishing instanceof TFCFishingHook || player.isRidingJumpable()) ? heightIn : heightIn + 6;
                default -> heightIn;
            }
```

eerussianguy (changes requested):
I would much rather see this incorporate the gui.right_height and gui.left_height and other constants better. As it stands we hardcode positions of all that stuff in a lot of places when in reality the health/hunger bars should automatically drop, because when the exp bar is rendered, we don't add to right/left height.

So basically rather than relying on our own system to always work, it's better for mod compat that we use the forge API heights, for ex. if another mod adds an exp bar. This solves how you basically added two options, the health bar jumping (the correct behavior) and it staying still (which is how it is now, which is wrong).

If you don't mind investigating if this is possible, that would be great. If you determine it's not possible then we can just look at merging this and absorbing the maintenance cost, but as it stands I feel like it's not the best solution.

I'm also not sure what the change is to the icons png, if you could elaborate on that that would be good.

Notenoughmail:
I can definitely look into adding to the left/right heights, thank you for pointing that out.

Insofar as the changes to overlay.png, it adds the bars which are used when the style is set to `LEFT_HOTBAR`. They let the fishing and jump bars be vertical and the fishing bar to be a different height depending on if the player is riding a horse or not.
apologies for the low quality
This is roughly what it looks like in-game, except the bars now charge upwards.

Notenoughmail:
So, I've been toying with ways of getting betting better compat with mods which have their own XP bar-like element. I've put up a [rough idea](https://github.com/Notenoughmail/TerraFirmaCraft/commit/0d6833947271732a02660c87eeed4beda17fae1f) of how it could possibly look, but it would require other mods to use a new TFC method or TFC to mixin into their render methods.

I looked at what some mods (Create and Botania) do for their own bars and they don't add to the left and right heights, so ensuring the left/right height is proper with them would have to be somewhat cludged either way.

I'm not entirely sure if something like that would be worth while to you as this would be a minor issue only visible after changing two config values, but if you think its worth pursuing or have alternative suggestions I can look into it further.

Edit: I've also come up with another idea for compat through a config option [here](https://github.com/Notenoughmail/TerraFirmaCraft/commit/b56afe6e9ae584306cb071ecaad27ba6a4a81b06), which would be a better option.

eerussianguy (approved):

PR #2317, opened 2023-02-20
Fix dogs and cats don't move

### LINKED ISSUE #2300: Dogs and cats don't move
Even when fully familiarized, dogs and cats don't move when given any of the commands. The sit command works as intended and they stand when given other commands, but they don't follow or walk around.

alesiong:
Fix #2300

Seems that we have to set `checkWhileRunningAlso` of `RunIf` to true, so that it will check the predicate (sleeping state) every tick, otherwise `MoveToTargetSink` will stop when checking `RunIf#canStillUse`


eerussianguy (approved):

eerussianguy:
Thanks, we will have to apply this to some other mobs as well.

PR #2324, opened 2023-03-05
Editing a translation file UA

Grigoriewich:
Editing a translation file (partial)


alcatrazEscapee:
Please resolve the merge conflicts in this PR. In particular, this seems to be based on an older version of both translation files and hence hasn't merged in newer changes.

PR #2325, opened 2023-03-06
[Lang] Update of the pt_br translation

FITFC:
* patchouli_book added pt_br translation (Mostly)
* New translations added in game file

The files have been tested and are loading normally in the game

Note: This version may contain some translation errors as it has not yet been fully proofread.


alcatrazEscapee:
For future reference, you have not committed the one file we actually need for the book translation - `resources/lang/pt_br.json`. Fortunately, as this has happened before, I do happen to have a tool to reverse engineer this, although it may have botched some of your work.

I'd also ask, if you are up to it, there is one more [`en_us.json`](https://github.com/TerraFirmaCraft/Field-Guide/blob/main/assets/lang/en_us.json) which has online field guide only elements, that you could translate if desired. (This one is much smaller though - only 60 entries, not ~900)

PR #2331, opened 2023-03-08
Reverse Intrusive/Extrusive in primitive anvils article

Gaelmare:
Actually caught this while watching an ATM Gravitas video.


alcatrazEscapee:
This is missing updates for translated books.

Gaelmare:
Not sure what you're expecting here. Update the English to be translated string of every translation?

Gaelmare:
Updated remaining files via __main__.py all. Not sure what the correct invocation is, but that appeared to work.

PR #2333, opened 2023-03-09
Add support for Antique Atlas

Jummit:
Instead of looking empty, the antique atlas now looks like this:

Screenshot from 2023-03-09 17-54-15

I didn't add any custom terrain textures because I felt the vanilla ones fit fairly well.

I used a Python script to generate this, I think I'll take a look at how the other generations scripts work and push it later.

### DIFF
--- a/resources/world_gen.py
+++ b/resources/world_gen.py
@@ -53,39 +53,39 @@ def generate(rm: ResourceManager):
     placed_feature_tag(rm, 'feature/volcanoes', 'tfc:volcano_rivulet', 'tfc:volcano_caldera', 'tfc:random_volcano_fissure')
 
     # Biomes
-    biome(rm, 'badlands', 'mesa', lake_features=False)
-    biome(rm, 'inverted_badlands', 'mesa', lake_features=False)
-    biome(rm, 'canyons', 'plains', boulders=True, lake_features=False, volcano_features=True, hot_spring_features=True)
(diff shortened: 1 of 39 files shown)


alcatrazEscapee:
Thanks! This should be a fairly straightforward addition to our existing data generation, yes, and I'd be happy to have this compat.

Jummit:
> Thanks! This should be a fairly straightforward addition to our existing data generation, yes, and I'd be happy to have this compat.

I'm working on it right now, what would be a good way to guarantee that new biomes are added to the atlas data files? Should I put it in the `world_gen.py` script where biomes are registered?

alcatrazEscapee:
Add the atlas type as a parameter to `biome()`, and generate the file there:

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/af1e782067e62ce1d9bffa3c2db5a88216551790/resources/world_gen.py#L1247

Then change the callsites:

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/af1e782067e62ce1d9bffa3c2db5a88216551790/resources/world_gen.py#L56-L88

Jummit:
I've implemented the data generation.

eerussianguy (approved):

eerussianguy (changes requested):
you can see resource validation fails because there's a diff between when you run `resources all` and the current state. please regen resources.

Jummit:
> you can see resource validation fails because there's a diff between when you run `resources all` and the current state. please regen resources.

Done.

PR #2345, opened 2023-03-23
Allow redstone activation of powderkegs and sealing of barrels and large vessels

Gaelmare:


src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
```diff
@@ -121,4 +121,12 @@ public boolean canDropFromExplosion(BlockState state, BlockGetter level, BlockPo
     {
         return false;
     }
+
+    @Override
+    public void neighborChanged(BlockState state, Level level, BlockPos pos, Block blockIn, BlockPos fromPos, boolean isMoving) {
```
eerussianguy: bracket is in the wrong spot

src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
```diff
@@ -121,4 +121,12 @@ public boolean canDropFromExplosion(BlockState state, BlockGetter level, BlockPo
...
+        if (level.hasNeighborSignal(pos) && !state.getValue(LIT))
+        {
+            level.getBlockEntity(pos, TFCBlockEntities.POWDERKEG.get()).ifPresent(keg -> keg.setLit(true,null));
```
alcatrazEscapee:
```suggestion
            level.getBlockEntity(pos, TFCBlockEntities.POWDERKEG.get()).ifPresent(keg -> keg.setLit(true, null));
```

src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
agilmore2: Ugh. Thought I had the codestyle loaded. :/

Gaelmare:
Thanks for reviews, I've added barrel/large vessel redstone to this PR, and I think it's ready for merge

eerussianguy:
why not override neighborChanged in the original method and include the static  method's contents there, and then call super()?

You also need to override getStateForPlacement to add POWERED during placement

Gaelmare:
> why not override neighborChanged in the original method and include the static method's contents there, and then call super()?

return type of the static method doesn't match, and needs to make a tri-state decision (null, true, false). I don't think I can make this fit into a super() call.

> 
> You also need to override getStateForPlacement to add POWERED during placement

Thanks! I handled this by adding the POWERED state in the super class method.

Gaelmare:
Needs a config option still, doh! Please don't merge yet.

Gaelmare:
Ready. again?!?

src/main/java/net/dries007/tfc/common/blocks/devices/SealableDeviceBlock.java
```diff
@@ -59,7 +62,8 @@ public VoxelShape getShape(BlockState state, BlockGetter level, BlockPos pos, Co
     @Override
     public BlockState getStateForPlacement(BlockPlaceContext context)
     {
-        return context.getItemInHand().getTag() != null ? defaultBlockState().setValue(SEALED, true) : defaultBlockState();
+        return context.getItemInHand().getTag() != null ?
+            defaultBlockState().setValue(SEALED, true).setValue(POWERED, false) : defaultBlockState().setValue(POWERED, false);
```
alcatrazEscapee: Neither of these is necessary, that's the point of `defaultBlockState()` and `registerDefaultState()`

Gaelmare: I was merely following the existing code. I can clean up.

Gaelmare: More changes incoming. Still not quite sure how the lambda would work, going to try it next.

src/main/java/net/dries007/tfc/common/blockentities/BarrelBlockEntity.java
```diff
@@ -415,7 +414,7 @@ public void tickPouring(Level level, BlockPos pos, boolean sealed, Direction fac
         }
     }
 
-    public void onSeal()
+    public void onSeal(Level level, BlockPos pos)
```
alcatrazEscapee: These methods, being implemented on a `BlockEntity` which already has access to the `level` and `worldPosition`, shouldn't have these parameters. Moreover this signature change is now a breaking change, which if possible I'd rather avoid, as it's unnecessary.

This is partially my mistake - I had seen the usage written as 

```java
 handleNeighborChanged(level, pos, state, () -> barrel.onSeal(level, pos), () -> barrel.onUnseal(level, pos));
```

and suggested using a `BiConsumer<Level, BlockPos>`, as this allows you to reduce the call to `barrel::onSeal`, `barrel::onUnseal`, etc - which I did without realizing there was a signature change here. However since we don't want these parameters on the block entity anyway, this makes more sense to just be a `Runnable` parameter, still passing in `barrel::onSeal` and `barrel::onUnseal`.

- Method ref: good (not capturing redundant args as lambda parameters), means the method is typed correctly
- Not modifying the signature here where we don't have to: also good. (This method doesn't need `level` and `pos`, as fields exist with both of these present.)

eerussianguy (approved):

PR #2365, opened 2023-04-20
Update zh_cn, zh_tw, zh_hk

aquila-zyy:
Clipped some entries from the lang files under tfc namespace to minecraft namespace, because overriding vanilla item names doesn't work in a different namespace.

Ready to merge.


alcatrazEscapee:
> Add vanilla override back to tfc namespace bc the workflow breaks if entries are missing

The issue here is the workflow, along with related data generation, need to be fixed to generate and handle both files. This change also needs to be ported to all other existing translations. Not band-aid fixed so it passes.

aquila-zyy:
I see. Will handle it later today.

aquila-zyy:
Manually ported vanilla overrides of all languages to minecraft namespace. Modified format_lang.py and usages of it so that the workflow now checks every other lang file against en_us in both namespaces. Temporarily commented out vanilla overides defined in constants.py, so they do not appear in data generation for lang files under tfc. I did not set up checks for minecraft namespace because I'm not sure how to do it, but someone might want to do it later.

eerussianguy:
Is this ready to go?

aquila-zyy:
Yes.

PR #2370, opened 2023-04-26
Bloomery recipes now require more catalyst per units of input

DavidTriphon:
WIP: Current issue: does not always return all catalyst items upon breaking.


src/main/java/net/dries007/tfc/common/blockentities/BloomeryBlockEntity.java
```diff
@@ -352,11 +352,29 @@ else if (cachedRecipe.matchesCatalyst(stack))
             }
 
             // Now, based on the number of both inputs and catalysts we have identified, we try and insert as much as we can in pairs, up to the capacity
-            // Note that the number of items and the number of catalyst items will always match, and each stack in the inputStacks + catalystStacks should have stack size = 1
-            final int totalInsertCapacity = Math.min(capacity - inputStacks.size(), Math.min(inputCount, catalystCount));
+            // Note that each stack in the inputStacks + catalystStacks should be one layer in the bloomery, so 1 item per inputStack with according catalyst count
+            for (ItemEntity foundInput : foundInputs)
```
eerussianguy, line 356: tbh i wonder if molten blocks should have more layers? @alcatrazEscapee thoughts?

PR #2408, opened 2023-05-24
Fix for lamps not turning off when the fuel runs out

stuartvc:
I noticed an issue with lamps staying lit when their fuel becomes empty.

I believe this is because the code is checking if the <used> FluidStack is empty instead of the <tank> FluidStack.

This bug is only noticeable when the fuel empties to 0 exactly. If the fuel dips into a negative value, the second check in the if statement is triggered and the lamp is correctly turned off.
Since the fuel level is only checked when the lamp receives a random tick, this bug happens most frequently when the lamp is in a loaded chunk and is receiving random ticks.

I've tested with both Tallow and Olive Oil, and both seem to be working as expected now.

Also, the second check in the if statement doesn't appear to be necessary anymore since tank.isEmpty() should handle all cases, I can remove that if you like.

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/LampBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/LampBlockEntity.java
@@ -62,7 +62,7 @@ public class LampBlockEntity extends TickCounterBlockEntity implements FluidTank
             if (usage >= 1)
             {
                 FluidStack used = tank.drain(usage, IFluidHandler.FluidAction.EXECUTE);
-                if (used.isEmpty() || used.getAmount() < usage)
+                if (tank.isEmpty() || used.getAmount() < usage)
                 {
                     level.setBlockAndUpdate(getBlockPos(), level.getBlockState(getBlockPos()).setValue(LampBlock.LIT, false));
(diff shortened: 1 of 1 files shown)


alcatrazEscapee:
I think I see why this happens - if it empties to zero, `getFuel()` above will return something empty, which won't trigger the fuel depletion. I don't think this is the correct fix here though, because you're effectively making a 1mB lamp burn for zero ticks (it should still burn). But that's a 1mB difference so heck.

stuartvc:
Ya, that makes sense about the `getFuel()` returning null for an empty lamp.
As for the 1mB burning, it seemed to me that the `Mth.floor(...)` would prevent a 1mB lamp from burning instantly. Since it would take at least `fuel.getBurnRate()` number of ticks until any fuel is used. (line 61)

I would add another question: if the lamp tank empties down to 0mB, should the lamp instantly turn off, or does that mean the lamp is currently burning the last remaining bit of fuel? Maybe the fuel has drained from the tank and is being burned in the wick.

So my first attempt to fix this assumed the first scenario, `if (tank.isEmpty()...` then shut the lamp off immediately. But I could see an argument to implement it differently.

PR #2466, opened 2023-06-21
Add jade tooltip for barrel recipe time and charcoal forge remaining fuel time

HermitOwO:


eerussianguy:
We've decided not to include burn ticks for jade tooltips as they are not synced regularly for every device

HermitOwO:
I see, how about barrel recipe time?

eerussianguy:
Seems fine to me.

HermitOwO:
Got it updated

eerussianguy:
BTW i'm changing it to Calendars.get(level) on my own commit as the sidedness of Jade is actually a little up in the air...

PR #2484, opened 2023-07-27
ua_uk localization addup and fixes

zeromero-dev:
More translations have been added but still not all


alcatrazEscapee:
cc @Grigoriewich if you have comments.

Grigoriewich:
I have no comments yet. I am very busy now. I will continue my work after a while. There are a lot of errors in the main file. I am very glad that there are still members engaged in translation.

PR #2487, opened 2023-08-01
Added new advancement for full nutrition

pietro-lopes:


alcatrazEscapee:
Resource validation failed; the new translation keys are missing from all other langs. (This is mostly a note to self, I can fix this when merging).

pietro-lopes:
yeah, I was having problem running python but checked how the validation process works and now managed to generate all languages, it was crashing on es_es before because I didn't know how to run properly

PR #2491, opened 2023-08-05
Update CI

LetterN:
make gradle cache, use local gradlew and verify local gradlew

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,33 +9,32 @@ on:
 
 jobs:
   main:
-
     runs-on: ubuntu-latest
-    timeout-minutes: 20
-
+    concurrency:
+      group: run_ci-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
     steps:
-    - name: Checkout
-      uses: actions/checkout@v3
+    - uses: actions/checkout@v3
+    - uses: gradle/wrapper-validation-action@v1 # this validates gradlew
     - name: Set up JDK 17
       uses: actions/setup-java@v3
       with:
         java-version: '17'
         distribution: 'adopt'
-    - name: Build, Test, Check Licenses
-      uses: gradle/gradle-build-action@v2
-      with:
-        arguments: build
+        cache: 'gradle'
+
+    # use local gadlew
+    - name: Compile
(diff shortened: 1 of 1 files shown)


alcatrazEscapee:
I'm sorry what is the point of these changes?

LetterN:
moreso standarnize tests, use local gradle so if an issue arises it can be tested localy, also verify the gradle used using `   gradle/wrapper-validation-action@v1` 

and also use cache feature so ci is faster on future runs

alcatrazEscapee:
> moreso standarnize [sic] tests,

You've provided absolutely no justification for these changes, and given the rest of the changes are pointless at best, I don't really see a reason to do any of them.

> use local gradle

From the documentation of the gradle build action, this is clearly already the case:

> The action will first look for a Gradle wrapper script in the root directory of your project. If not found, gradle will be executed from the PATH.

> verify the gradle used using   ` gradle/wrapper-validation-action@v1`

This is probably the only thing of substance in this PR. But even the description of that is to avoid PRs which update the gradle wrapper maliciously. We have never gotten nor accepted such a PR, and being a Minecraft mod we don't update gradle at will, we update it when we port across versions (as the toolchain dictates our gradle setup). So the risk of this *right now* is zero and the future risk is extremely low.

> and also use cache feature so ci is faster on future runs

Again, from the documentation, this is already the case and is recommended *over* the cache features you have enabled (anecdotally, which were causing us problems before)

> ### Why use the gradle-build-action?
> It is possible to directly invoke Gradle in your workflow, and the actions/setup-java@v3 action provides a simple way to cache Gradle dependencies.
> 
> However, the gradle-build-action offers a number of advantages over this approach:
> [...]
> 
> - More sophisticated and more efficient caching of Gradle User Home between invocations, compared to setup-java and most custom configurations using actions/cache. [More details below](https://github.com/gradle/gradle-build-action#caching).
> - Detailed reporting of cache usage and cache configuration options allow you to [optimize the use of the GitHub actions cache](https://github.com/gradle/gradle-build-action#optimizing-cache-effectiveness).

Moreover, you've removed the timeout we added (because FG can hang sometimes in mcpconfig setup), added `--no-daemon` which is redundant given we have `org.gradle.daemon=false` already in our `gradle.properties`.

PR #2496, opened 2023-08-13
Add a Collapse Event

Notenoughmail:
Adds an event which provides the `centerPos`, `nextPositions`, `radiusSquared`, and level of a collapse

### DIFF
--- a/src/main/java/net/dries007/tfc/common/recipes/CollapseRecipe.java
+++ b/src/main/java/net/dries007/tfc/common/recipes/CollapseRecipe.java
@@ -9,6 +9,8 @@ package net.dries007.tfc.common.recipes;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
+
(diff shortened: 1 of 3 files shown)


pietro-lopes:
While we are at it, can I ask to add this to FakeCollapses too? maybe using same event and send a boolean if it is a fake or not

Notenoughmail:
That can be done, but the radius of the fake collapses will have to be 0, as that only exists during a real collapse.

Otherwise, I could try to create a radius from the randx/y/z that are used to find unsupported blocks

pietro-lopes:
I'm fine with having no radius, because if the dev needs he can inflate the radius for whatever reason
Does it give at least the center point?

Notenoughmail:
Yes, the centerPos is available

pietro-lopes:
Looking good to me 😀

eerussianguy (approved):

PR #2507, opened 2023-09-06
Backport Ingot Pile shape and Blast Furnace JEI compat to 1.18

Jerome226:


alcatrazEscapee:
Well, since I didn't run the auto tests this is largely my fault but @Jerome226 for future reference, do not edit `en_us` by hand, it is an auto-generated file (and in particular, it generates all the other languages too, not just `en_us`).

PR #2516, opened 2023-10-08
Rock layer rework

alcatrazEscapee:
Overview:

- Rock layers are dependent. "Marble is found under limestone" sort of thing. It's _fairly_ predictable and we'll have to go in later and figure out a good way to explain it in the field guide.
- More rock layers (in general)
- Rock layer height follows the surface rather than inverse (so generally higher terrain = more rock layers)
- Rock layers have skewed boundaries, and general total weirdness that wasn't possible before
- Rock data information is not saved after chunk generation (it's not even saved during chunk generation, just some noise is cached and the rest is computed on demand.


src/main/java/net/dries007/tfc/world/TFCChunkGenerator.java
```diff
@@ -448,6 +446,7 @@ public void applyCarvers(WorldGenRegion level, long seed, RandomState state, Bio
     @Override
     public void applyBiomeDecoration(WorldGenLevel level, ChunkAccess chunk, StructureManager structureFeatureManager)
     {
+        if (true) return;
```
eerussianguy: debug stuff left in

PR #2519, opened 2023-10-13
Hanging sign renderer

Peffern:
- hanging signs for every combo of wood and tool metal (for the chain)
- custom renderer via @Accessor to HangingSignRenderer
- new item chain texture layer
- new hanging sign edit GUI (requires @Injector on HangingSignEditScreen)

the diff is huge because there's MxN hanging sign combinations for texture, model, loot table, crafting, etc.


src/main/java/net/dries007/tfc/client/render/blockentity/TFCHangingSignBlockEntityRenderer.java
```diff
@@ -58,6 +71,42 @@ public void render(SignBlockEntity sign, float partialTick, PoseStack poseStack,
         final WoodType woodtype = SignBlock.getWoodType(signblock);
         final HangingSignRenderer.HangingSignModel model = this.hangingSignModels.get(woodtype);
         model.evaluateVisibleParts(blockstate);
-        ((SignRendererAccessor) (Object) this).invoke$renderSignWithText(sign, poseStack, buffer, light, overlay, blockstate, signblock, woodtype, model);
+        if(signblock instanceof ITFCHangingSignBlock tfcSignBlock) {
```
eerussianguy: Braces go on newlines, just import the code style and let intellij apply it

src/main/java/net/dries007/tfc/client/render/blockentity/TFCHangingSignBlockEntityRenderer.java
```diff
@@ -58,6 +71,42 @@ public void render(SignBlockEntity sign, float partialTick, PoseStack poseStack,
...
+            final Metal metal = Metal.MANAGER.getOrThrow(tfcSignBlock.metal());
+            this.renderSignWithText(sign, poseStack, buffer, light, overlay, blockstate, signblock, woodtype, metal, model);
```
eerussianguy: My intuition is the hanging sign block should not store a _metal_ but rather a texture or model location (ResourceLocation). This way we don't have to do a potentially dangerous lookup like this

src/main/java/net/dries007/tfc/common/blockentities/TFCBlockEntities.java
```diff
@@ -81,7 +84,7 @@ public final class TFCBlockEntities
...
     public static final RegistryObject<BlockEntityType<BookshelfBlockEntity>> BOOKSHELF = register("bookshelf", BookshelfBlockEntity::new, TFCBlocks.WOODS.values().stream().map(map -> map.get(Wood.BlockType.BOOKSHELF)));
     public static final RegistryObject<BlockEntityType<TFCSignBlockEntity>> SIGN = register("sign", TFCSignBlockEntity::new, TFCBlocks.WOODS.values().stream().flatMap(map -> Stream.of(Wood.BlockType.SIGN, Wood.BlockType.WALL_SIGN).map(map::get)));
-    public static final RegistryObject<BlockEntityType<TFCHangingSignBlockEntity>> HANGING_SIGN = register("hanging_sign", TFCHangingSignBlockEntity::new, TFCBlocks.WOODS.values().stream().flatMap(map -> Stream.of(Wood.BlockType.HANGING_SIGN, Wood.BlockType.WALL_HANGING_SIGN).map(map::get)));
+    public static final RegistryObject<BlockEntityType<TFCHangingSignBlockEntity>> HANGING_SIGN = register("hanging_sign", TFCHangingSignBlockEntity::new, TFCBlocks.HANGING_SIGNS.values().stream().flatMap(map -> map.values().stream().flatMap(map2 -> map2.values().stream())));
```
eerussianguy: lets find a better variable name than `map2`

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -6,35 +6,35 @@
...
-import net.minecraft.world.level.block.Blocks;
-import net.minecraft.world.level.block.FlowerPotBlock;
-import net.minecraft.world.level.block.GravelBlock;
-import net.minecraft.world.level.block.LiquidBlock;
-import net.minecraft.world.level.block.SeaPickleBlock;
-import net.minecraft.world.level.block.SlabBlock;
-import net.minecraft.world.level.block.SoundType;
-import net.minecraft.world.level.block.StairBlock;
-import net.minecraft.world.level.block.WallBlock;
+import net.minecraft.world.level.block.*;
```
eerussianguy: code style. we no longer use wildcard imports. This should also get auto fixed if you import the code style

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -241,6 +238,19 @@ public final class TFCBlocks
         )
     );
 
+    public static final Map<Wood, Map<Metal.Default, Map<Class<? extends ITFCHangingSignBlock>, RegistryObject<SignBlock>>>> HANGING_SIGNS = Helpers.mapOfKeys(Wood.class, wood -> {
```
eerussianguy: Is there a reason we are cludging these into a single map? None of our block registrations look like this and it just seems to have a needlessly large code footprint

src/main/java/net/dries007/tfc/common/blocks/TFCBlocks.java
```diff
@@ -241,6 +238,19 @@ public final class TFCBlocks
...
+        Supplier<ExtendedProperties> woodProps = () -> Wood.BlockType.properties(wood).noCollission().strength(1F).flammableLikePlanks().blockEntity(TFCBlockEntities.HANGING_SIGN).ticks(SignBlockEntity::tick);
+        WoodType woodType = wood.getVanillaWoodType();
+        return Helpers.mapOfKeys(Metal.Default.class, metal -> metal.metalTier() != Metal.Tier.TIER_0, metal -> {
```
eerussianguy: there's a helper method called like Metal#has that does this check. Check what 'Part' type chains have (i think its utility?) and check against that, don't use tier which is only for behavior

src/main/java/net/dries007/tfc/common/blocks/wood/ITFCHangingSignBlock.java
```diff
@@ -0,0 +1,7 @@
+package net.dries007.tfc.common.blocks.wood;
+
+import net.minecraft.resources.ResourceLocation;
+
+public interface ITFCHangingSignBlock {
+    ResourceLocation metal();
```
eerussianguy: Not necessary

src/main/java/net/dries007/tfc/common/blocks/wood/TFCCeilingHangingSignBlock.java
```diff
@@ -20,14 +22,20 @@
...
 import net.dries007.tfc.common.blocks.IForgeBlockExtension;
 
-public class TFCCeilingHangingSignBlock extends CeilingHangingSignBlock implements IForgeBlockExtension, EntityBlockExtension
+public class TFCCeilingHangingSignBlock extends CeilingHangingSignBlock implements IForgeBlockExtension, EntityBlockExtension, ITFCHangingSignBlock
 {
     private final ExtendedProperties properties;
+    private final ResourceLocation metal;
 
-    public TFCCeilingHangingSignBlock(ExtendedProperties properties, WoodType type)
+    public TFCCeilingHangingSignBlock(ExtendedProperties properties, WoodType type, ResourceLocation metal)
```
eerussianguy: AFAICT you are mapping a resource location to a lookup of the `Metal` in `Metal.MANAGER` and then just using the metal's path? You should construct the full model and texture location in the ctor of this class and use it directly in the renderer. For ex. in `JavelinItem` that's what we do. We can also use a convenience ctor that takes a `Metal.Default` that directs to  a constructor that takes a `ResourceLocation` for maximum code cleanliness

src/main/java/net/dries007/tfc/common/blocks/wood/Wood.java
```diff
@@ -185,14 +185,12 @@ public enum BlockType
...
-        WALL_HANGING_SIGN(wood -> new TFCWallHangingSignBlock(properties(wood).noCollission().strength(1F).flammableLikePlanks().blockEntity(TFCBlockEntities.HANGING_SIGN).ticks(SignBlockEntity::tick), wood.getVanillaWoodType()), true),
         BARREL((self, wood) -> new BarrelBlock(properties(wood).strength(2.5f).flammableLikePlanks().noOcclusion().blockEntity(TFCBlockEntities.BARREL).serverTicks(BarrelBlockEntity::serverTick)), false, BarrelBlockItem::new),
         LECTERN(wood -> new TFCLecternBlock(properties(wood).noCollission().strength(2.5F).flammableLikePlanks().blockEntity(TFCBlockEntities.LECTERN)), false),
         SCRIBING_TABLE(wood -> new ScribingTableBlock(properties(wood).noOcclusion().strength(2.5F).flammable(20, 30)), false),
         JAR_SHELF(wood -> new JarShelfBlock(properties(wood).noOcclusion().strength(2.5f).flammableLikePlanks().blockEntity(TFCBlockEntities.JARS)), false);
 
-        private static ExtendedProperties properties(RegistryWood wood)
+        public static ExtendedProperties properties(RegistryWood wood)
```
eerussianguy: this is I guess, fine, but given the relatively small amount of properties this convenience static method adds you should probably just copy it over as its just used in `Wood` for readability/consistency. ideally you should see all the properties you're adding without clicking out of the class

src/main/java/net/dries007/tfc/mixin/client/HangingSignEditScreenMixin.java
```diff
@@ -0,0 +1,30 @@
...
+public abstract class HangingSignEditScreenMixin {
+    @Mutable
+    @Final
+    @Shadow
+    private ResourceLocation texture;
+    @Inject(method="<init>(Lnet/minecraft/world/level/block/entity/SignBlockEntity;ZZ)V", at=@At("TAIL"))
+    public void inject$constructor(SignBlockEntity signBlockEntity, boolean isFrontText, boolean filter, CallbackInfo ci) {
+        Block block = signBlockEntity.getBlockState().getBlock();
+        if (block instanceof ITFCHangingSignBlock tfcSignBlock) {
+            this.texture = new ResourceLocation(SignBlock.getWoodType(block).name() + ".png").withPrefix("textures/gui/hanging_signs/" + tfcSignBlock.metal().getPath() + "/");
```
eerussianguy: trailing `/` ?

eerussianguy (changes requested):
Something that we've been exploring lately is moving some information to tooltip flavor text, so it would be like

**Oak Hanging Sign**
*Red Steel Chain*

But we'd need @alcatrazEscapee  to approve that. Also requesting that alcatraz look at some of the design choices here

src/main/java/net/dries007/tfc/common/blocks/wood/ITFCHangingSignBlock.java
Peffern: discussed on discord - using the marker interface is maybe preferable to doing instanceof checks for both TFCCeilingHangingSignBlock and TFCWallHangingSignBlock. but the handling of the resource will be different

src/main/java/net/dries007/tfc/mixin/client/HangingSignEditScreenMixin.java
Peffern: don't think so- compare the one in the base HangingSignEditScreen, it has the trailing slash as well.

src/main/java/net/dries007/tfc/common/blocks/wood/ITFCHangingSignBlock.java
Peffern: split up into gui texture and renderer material

alcatrazEscapee:
I disagree with moving the item name to a tooltip, that's not what we've been doing recently. The case I just used it for was adding "_Igneous Extrusive_" to "Raw Granite". Not splitting up the item name because it was too long.

Utility Metals x Wood Types x Sign Types is the natural conclusion here. Don't introduce extra dynamic behavior to hide what's going on behind the scenes, we don't do that for any other block or item why would we do it here.

If you think it's creating too many items/blockstates, I invite you to first look at wall blocks, then consider "should _chains_ be made out of every metal type or should they be only iron/steel" which will obviously reduce the permutations here.

PR #2523, opened 2023-10-22
Update UA translation

Grigoriewich:
Some corrections and additions to the Ukrainian translation. I hope that there will still be releases of 1.18.x.)


alcatrazEscapee:
The failing checks are from resource gen not having been ran which is fine, it can be fixed on merge (or you can run `python resources book --translate-all` if you desire.

Grigoriewich:
I was unable to use the Python resource book. Probably lack of experience. I will be grateful for your help.

PR #2527, opened 2023-10-24
create explosion proof tag for powderkeg explosion

HermitOwO:


eerussianguy:
Wait -- this makes no sense. We own `PowderkegExplosion`. There is no reason for us to listen to an event for our own explosions. It should be implemented directly in the class.

src/main/java/net/dries007/tfc/util/PowderKegExplosion.java
```diff
@@ -77,7 +80,11 @@ public void finalizeExplosion(boolean spawnParticles)
             this.level.addParticle(ParticleTypes.EXPLOSION_EMITTER, this.x, this.y, this.z, 1.0D, 0.0D, 0.0D);
         }
 
-        final List<BlockPos> affectedBlockPositions = this.getToBlow();
+        List<BlockPos> affectedBlockPositions = this.getToBlow()
```
eerussianguy: Sorry to make you change things again -- should have been clearer.

This is doubling the amount of getBlockState calls per explosion as well as using a stream which has minor performance implications.

It should be like so

```java
for (BlockPos pos : affectedBlockPositions)
  {
    final BlockState state = level.getBlockState(pos);
    
    if (Helpers.isBlock(state, TFCTags.Blocks.EXPLOSION_PROOF))
        continue;
     
  }```

PR #2531, opened 2023-10-27
Ingredient Jsonification

Notenoughmail:
Currently implements `toJson` in item ingredients and fluid (stack) ingredients

Block ingredients and item stack providers are up next

### DIFF
--- a/src/main/java/net/dries007/tfc/common/recipes/ingredients/NotRottenIngredient.java
+++ b/src/main/java/net/dries007/tfc/common/recipes/ingredients/NotRottenIngredient.java
@@ -6,10 +6,12 @@
 
 package net.dries007.tfc.common.recipes.ingredients;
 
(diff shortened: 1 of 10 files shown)


src/main/java/net/dries007/tfc/common/recipes/ingredients/NotRottenIngredient.java
```diff
@@ -54,6 +56,17 @@ protected ItemStack testDefaultItem(ItemStack stack)
         }).orElse(null);
     }
 
+    @Override
+    public JsonElement toJson()
+    {
+        JsonObject json = new JsonObject();
+        json.addProperty("type", CraftingHelper.getID(Serializer.INSTANCE).toString());
+        if (delegate != null) {
```
eerussianguy: braces on newlines please

PR #2532, opened 2023-10-27
New autumn foliage

Therighthon:
- Redid autumn color map. The horizontal axis is progress through autumn, summer on the left, winter on the right, but with a  hash pattern so not all leaves are on the same schedule. The vertical axis is autumn index, which is a parameter assigned per-species of tree. It is constant, so it defines the 
- Redid winter color map. Remains green in moist tropical environments (Above 14C average)
- Conifer colormap now restricted to conifers, kapoks, palms, etc. grown in cold climates will change color in fall, northern seasonal trees grown in the tropics will not change colors
- The timing of autumn color changes is now based on height-adjusted average temperature, so fall colors will come down from the north/mountains, tops of trees will change color first, longer winters farther north.


src/main/java/net/dries007/tfc/util/registry/RegistryWood.java
```diff
@@ -31,6 +31,8 @@ public interface RegistryWood extends StringRepresentable
 
     int daysToGrow();
 
+    int autumnIndex();
```
eerussianguy, line 37: needs a javadoc

src/main/java/net/dries007/tfc/common/blocks/wood/TFCLeavesBlock.java
```diff
@@ -77,18 +77,30 @@ public static void onEntityInside(BlockState state, Level level, BlockPos pos, E
         }
     }
 
+    public static TFCLeavesBlock create(ExtendedProperties properties, int autumnIndex)
```
eerussianguy: I think we might be able to not use these static create methods now that the leaves blocks aren't distance sensitive anymore... try removing and using ctors?

src/main/java/net/dries007/tfc/client/TFCColors.java
```diff
@@ -125,25 +137,61 @@ public static int getWaterFogColor(CommonLevelAccessor level, Biome biome, Block
...
+        final float springStart = 1f - autumnEnd;
+
+        if (timeOfYear > autumnEnd)
+        {
+            return getClimateColor(FOLIAGE_WINTER_COLORS_CACHE, pos);
+        }
+        else if (timeOfYear > autumnStart)
+        {
+            final Block block = state.getBlock();
+            final int autumnIndex = block instanceof TFCLeavesBlock ? ((TFCLeavesBlock) block).getAutumnIndex() : 127;
```
eerussianguy: try `instanceof TFCLeavesBlock block`

I wonder if this should be on the `ILeavesBlock` interface so that for example I can implement the interface on a modded leaf block (for dynamic trees for example)

src/main/java/net/dries007/tfc/client/TFCColors.java
```diff
@@ -125,25 +137,61 @@ public static int getWaterFogColor(CommonLevelAccessor level, Biome biome, Block
...
+        {
+            return getSeasonalFoliageColor(pos, level, state);
         }
         return -1;
     }
 
+    /**
+     * Gets a color based on average temperature and time of year. Autumn occurs at different times of the year at height-adjusted average temperatures from the poles to 14c
+     */
+    private static int getSeasonalFoliageColor(BlockPos pos, LevelAccessor level, BlockState state)
```
eerussianguy: I'm suspicious about this being state sensitive but it seems fine.

alcatrazEscapee: - Remove `BlockState state` parameter and replace with `int autumnIndex`
- Remove the calculation here:

```
final int autumnIndex = block instanceof TFCLeavesBlock ? ((TFCLeavesBlock) block).getAutumnIndex() : 127;
```

src/main/java/net/dries007/tfc/common/blocks/wood/TFCLeavesBlock.java
```diff
@@ -77,18 +77,30 @@ public static void onEntityInside(BlockState state, Level level, BlockPos pos, E
...
+
     /* The maximum value of the decay property. */
     private final int maxDecayDistance;
     private final ExtendedProperties properties;
+    private final int autumnIndex;
     @Nullable private final Supplier<? extends Block> fallenLeaves;
     @Nullable private final Supplier<? extends Block> fallenTwig;
 
-    protected TFCLeavesBlock(ExtendedProperties properties, @Nullable Supplier<? extends Block> fallenLeaves, @Nullable Supplier<? extends Block> fallenTwig)
+    protected TFCLeavesBlock(ExtendedProperties properties, int autumnIndex, @Nullable Supplier<? extends Block> fallenLeaves, @Nullable Supplier<? extends Block> fallenTwig)
```
alcatrazEscapee, line 87: Remove `autumnIndex` as a property on `TFCLeavesBlock` entirely.

src/main/java/net/dries007/tfc/common/blocks/wood/Wood.java
```diff
@@ -159,7 +167,7 @@ public enum BlockType
         STRIPPED_LOG(wood -> new LogBlock(ExtendedProperties.of(state -> state.getValue(RotatedPillarBlock.AXIS) == Direction.Axis.Y ? wood.woodColor() : wood.barkColor()).strength(7.5f).sound(SoundType.WOOD).requiresCorrectToolForDrops().flammableLikeLogs(), null), false),
         WOOD((self, wood) -> new LogBlock(properties(wood).strength(8f).requiresCorrectToolForDrops().flammableLikeLogs(), wood.getBlock(self.stripped())), false),
         STRIPPED_WOOD(wood -> new LogBlock(properties(wood).strength(7.5f).requiresCorrectToolForDrops().flammableLikeLogs(), null), false),
-        LEAVES((self, wood) -> new TFCLeavesBlock(ExtendedProperties.of().mapColor(MapColor.PLANT).strength(0.5F).sound(SoundType.GRASS).randomTicks().noOcclusion().isViewBlocking(TFCBlocks::never).flammableLikeLeaves(), wood.getBlock(self.fallenLeaves()), wood.getBlock(self.twig())), false),
+        LEAVES((self, wood) -> TFCLeavesBlock.create(ExtendedProperties.of().mapColor(MapColor.PLANT).strength(0.5F).sound(SoundType.GRASS).randomTicks().noOcclusion().isViewBlocking(TFCBlocks::never).flammableLikeLeaves(), wood.autumnIndex(), wood.getBlock(self.fallenLeaves()), wood.getBlock(self.twig())), false),
```
alcatrazEscapee: Remove `wood.autumnIndex()` parameter

src/main/java/net/dries007/tfc/client/ClientEventHandler.java
```diff
@@ -673,7 +674,7 @@ public static void registerColorHandlerBlocks(RegisterColorHandlersEvent.Block e
         final BlockColor grassColor = (state, level, pos, tintIndex) -> TFCColors.getGrassColor(pos, tintIndex);
         final BlockColor tallGrassColor = (state, level, pos, tintIndex) -> TFCColors.getTallGrassColor(pos, tintIndex);
         final BlockColor foliageColor = (state, level, pos, tintIndex) -> TFCColors.getFoliageColor(pos, tintIndex);
-        final BlockColor seasonalFoliageColor = (state, level, pos, tintIndex) -> TFCColors.getSeasonalFoliageColor(pos, tintIndex);
+        final BlockColor seasonalFoliageColor = (state, level, pos, tintIndex) -> TFCColors.getSeasonalFoliageColor(pos, tintIndex, state);
```
alcatrazEscapee: Remove this line, and then below, replace

```java
TFCBlocks.WOODS.forEach((wood, reg) -> event.register(wood.isConifer() ? foliageColor : seasonalFoliageColor, reg.get(Wood.BlockType.LEAVES).get(), reg.get(Wood.BlockType.FALLEN_LEAVES).get()));
```

with

```java
TFCBlocks.WOODS.forEach((wood, reg) -> event.register(
    wood.isConifer() ?
        foliageColor :
        (state, level, pos, tintIndex) -> TFCColors.getSeasonalFoliageColor(pos, tintIndex, wood.autumnIndex()),
    reg.get(Wood.BlockType.LEAVES).get(), reg.get(Wood.BlockType.FALLEN_LEAVES).get()));
```

And then fix the plants that use `seasonalFoliageColor` to either pick an autumn index, or not use the seasonal color.

alcatrazEscapee (changes requested):

src/main/java/net/dries007/tfc/client/TFCColors.java
alcatrazEscapee: See my other comments, this can avoid being on the block, the interface, or the state, and just part of registration which is as easy to implement for modded as possible.

src/main/java/net/dries007/tfc/util/registry/RegistryWood.java
alcatrazEscapee: Can actually be removed entirely, because there's nothing about `RegistryWood` that requires it to be exposed. It's only applied for our own blocks so it's fine as a method just on `Wood`.

src/main/java/net/dries007/tfc/common/blocks/wood/Wood.java
```diff
@@ -159,7 +167,7 @@ public enum BlockType
...
         WOOD((self, wood) -> new LogBlock(properties(wood).strength(8f).requiresCorrectToolForDrops().flammableLikeLogs(), wood.getBlock(self.stripped())), false),
         STRIPPED_WOOD(wood -> new LogBlock(properties(wood).strength(7.5f).requiresCorrectToolForDrops().flammableLikeLogs(), null), false),
-        LEAVES((self, wood) -> new TFCLeavesBlock(ExtendedProperties.of().mapColor(MapColor.PLANT).strength(0.5F).sound(SoundType.GRASS).randomTicks().noOcclusion().isViewBlocking(TFCBlocks::never).flammableLikeLeaves(), wood.getBlock(self.fallenLeaves()), wood.getBlock(self.twig())), false),
+        LEAVES((self, wood) -> new TFCLeavesBlock(ExtendedProperties.of().mapColor(MapColor.PLANT).strength(0.5F).sound(SoundType.GRASS).randomTicks().noOcclusion().isViewBlocking(TFCBlocks::never).flammableLikeLeaves(), ((Wood) wood).autumnIndex(), wood.getBlock(self.fallenLeaves()), wood.getBlock(self.twig())), false),
```
alcatrazEscapee:
```suggestion
        LEAVES((self, wood) -> new TFCLeavesBlock(ExtendedProperties.of().mapColor(MapColor.PLANT).strength(0.5F).sound(SoundType.GRASS).randomTicks().noOcclusion().isViewBlocking(TFCBlocks::never).flammableLikeLeaves(), wood.autumnIndex(), wood.getBlock(self.fallenLeaves()), wood.getBlock(self.twig())), false),
```

I don't think this cast is necessary, otherwise this is good work.

Therighthon: Nope, it definitely was not necessary, just a holdover from a much messier implementation

alcatrazEscapee (approved):

PR #2534, opened 2023-11-04
Climate changes again

Therighthon:
- Remap hardiness zones to new temp ranges, end result is plants and trees shift into slightly (~3C) colder climates
- Adjusted the southern extent of fall foliage accordingly
- Tweaked the winter color map to better blend with the autumn colors at their southern extent


alcatrazEscapee:
~~Looks like resource gen needs to be re-ran in order to update for some changes to bamboo placement.~~ On second look this appears to be something that was broken on the main branch, should be fixed by 2c0b293662bb1e3a7ab015671da8249839852c0b

PR #2576, opened 2023-12-09
Localization update

icyselec:
i have updated the localization translation for the Korean language. i have already inspected it and it won't be a problem. right now, only essential things have been translated.


eerussianguy:
Hey, we'd love for you to translate the field guide when you have time!

You can find the file [here](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.20.x/resources/lang/ko_kr.json). Note that it won't populate in game automatically. You can either translate and leave it to us or follow the [guide](https://github.com/TerraFirmaCraft/TerraFirmaCraft/wiki/Info-for-Translators) to translate and test it. You can even translate our website if you have time.

PR #2578, opened 2023-12-10
Rock Layer Size scaling

Gaelmare:
Unclear on right way to expose Customize World Gen available Settings to TFCLayers function call, but this way works.

Needs testing to ensure I got the right default integer.


src/main/java/net/dries007/tfc/client/screen/CreateTFCWorldScreen.java
```diff
@@ -43,15 +42,23 @@ private static OptionInstance<Double> pctOption(String caption, double defaultVa
 
     private static OptionInstance<Integer> kmOption(String caption, int min, int max, int defaultValue)
     {
-        return new OptionInstance<>(caption, OptionInstance.cachedConstantTooltip(Component.translatable(caption + ".tooltip")), (text, value) -> Options.genericValueLabel(text, Component.translatable("tfc.settings.km", String.format("%.1f", value / 1000.0))), new OptionInstance.IntRange(min, max), defaultValue, value -> {});
+        return new OptionInstance<>(caption, OptionInstance.cachedConstantTooltip(Component.translatable(caption + ".tooltip")),
+            (text, value) -> Options.genericValueLabel(text, Component.translatable("tfc.settings.km", String.format("%.1f", value / 1000.0))), new OptionInstance.IntRange(min, max), defaultValue, value -> {});
+    }
```
alcatrazEscapee, line 47: Why this diff?

src/main/java/net/dries007/tfc/client/screen/CreateTFCWorldScreen.java
```diff
@@ -43,15 +42,23 @@ private static OptionInstance<Double> pctOption(String caption, double defaultVa
...
+
+    private static OptionInstance<Integer> scaleOption(String caption, int min, int max, int defaultValue)
+    {
+        return new OptionInstance<>(caption, OptionInstance.cachedConstantTooltip(Component.translatable(caption + ".tooltip")),
+            (text, value) -> Options.genericValueLabel(text, Component.translatable("tfc.settings.scale", value)),
+            new OptionInstance.IntRange(min, max), defaultValue, value -> {});
     }
```
alcatrazEscapee, line 54: This is a bad option, it should be represented in km, and the tooltip should be in km, ~~log scale~~ (not necessary if done correctly).

src/main/java/net/dries007/tfc/client/screen/CreateTFCWorldScreen.java
```diff
@@ -106,6 +113,9 @@ protected void init()
         options.addBig(
             continentalness = pctOption("tfc.create_world.continentalness", settings.continentalness())
         );
+        options.addBig( // min max match 1.18 range comment
+            rockLayerScale = scaleOption("tfc.create_world.rocklayerscale", 0, 32, settings.rockLayerScale())
+        );
```
alcatrazEscapee, line 118: The comment here is useless, this ideally should be paired with a biome scale option as well.

src/main/java/net/dries007/tfc/world/layer/TFCLayers.java
```diff
@@ -156,7 +155,7 @@ public static AreaFactory createOverworldRockLayer(RegionGenerator generator, lo
         AreaFactory layer;
 
         layer = RegionRockLayer.INSTANCE.apply(regionLayer); // Grid scale (128x)
-        for (int i = 0; i < Units.GRID_BITS - 1; i++)
+        for (int i = 0; i < generator.getRockLayerScale(); i++)
```
alcatrazEscapee, line 158: This is wrong. The region is still generating at grid scale and these zoom layers are necessary for matching that scale. This will offset everything and not just make it smaller but make the layer type completely offset from the rock.

src/main/java/net/dries007/tfc/world/region/RegionGenerator.java
```diff
@@ -56,6 +56,8 @@ private static Noise2D baseNoise(boolean axisIsX, float scale, float constant)
 
     private final Cellular2D cellNoise;
 
+    private final Integer rockLayerScale;
```
alcatrazEscapee, line 59: `int`

src/main/java/net/dries007/tfc/world/region/RegionGenerator.java
```diff
@@ -93,13 +95,18 @@ public RegionGenerator(Settings settings, RandomSource random)
...
         rockArea = ThreadLocal.withInitial(rockAreaFactory);
+        this.rockLayerScale = settings.rockLayerScale();
     }
 
     public long seed()
     {
         return seed;
     }
 
+    public Integer getRockLayerScale() {
```
alcatrazEscapee, line 106: `int` and formatting

src/main/java/net/dries007/tfc/world/settings/Settings.java
```diff
@@ -40,6 +41,7 @@ public record Settings(
         Codec.INT.fieldOf("rainfall_scale").forGetter(c -> c.rainfallScale),
         Codec.FLOAT.optionalFieldOf("rainfall_constant", 0f).forGetter(c -> c.rainfallConstant),
         RockLayerSettings.CODEC.fieldOf("rock_layer_settings").forGetter(c -> c.rockLayerSettings),
-        Codec.FLOAT.fieldOf("continentalness").forGetter(c -> c.continentalness)
+        Codec.FLOAT.fieldOf("continentalness").forGetter(c -> c.continentalness),
+        Codec.INT.optionalFieldOf("rock_layer_scale", 6).forGetter(c -> c.rockLayerScale) // match 1.18 scales
```
alcatrazEscapee, line 45: pointless comment.

Also don't use `optionalFieldOf`

alcatrazEscapee (changes requested):

alcatrazEscapee:
Ultimately this is in the wrong location, and will break rock layer alignment with the continent type. This should be:

- ~~Not expressed as a log km option (because we can sample at float rounded positions)~~
  - **Edit** I take this back, log km is the way to go, this would create zoom artifacts otherwise.
- The sampling needs to be scaled here:

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/6e3fa49bec8095af84b04048ee8e6019692ec23a/src/main/java/net/dries007/tfc/world/region/ChooseRocks.java#L38

- Biome sampling is similar and is here:

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/6e3fa49bec8095af84b04048ee8e6019692ec23a/src/main/java/net/dries007/tfc/world/region/ChooseBiomes.java#L38

src/main/java/net/dries007/tfc/client/screen/CreateTFCWorldScreen.java
Gaelmare: Matches the wrapping of the lines above, much easier to read.

src/main/java/net/dries007/tfc/client/screen/CreateTFCWorldScreen.java
Gaelmare: What's the default? It sounds like scaling via 2x zooms in integer space is not your expectation of the implementation in 1.20?

src/main/java/net/dries007/tfc/client/screen/CreateTFCWorldScreen.java
Gaelmare: If we're no longer doing zooms via integer scale, I get that my comment is worthless. Wanted to give some motivation for the magic numbers. 

biome scale as in forest/lake/marsh, or terrain features?

src/main/java/net/dries007/tfc/world/layer/TFCLayers.java
Gaelmare: I guess it's finally time to do image generation of rock layers along with biomes in RegionGeneratorTest so I can visualize what you mean.

src/main/java/net/dries007/tfc/world/region/RegionGenerator.java
Gaelmare: Ugh, stupid Intellij user, sorry.

src/main/java/net/dries007/tfc/world/settings/Settings.java
Gaelmare: again, allergy to magic number with no context.

Gaelmare:
Thank you very much for the feedback!

Gaelmare:
Written without sufficient playtesting of current system. I'm comfortable with the scales as implemented for SSP. Looks great!

PR #2579, opened 2023-12-12
Fix barrel tooltip

### LINKED ISSUE #2575: Barrel recipe tooltip resets after world loading
This only occurs outside spawn chunks.

HermitOwO:
Fix #2575


eerussianguy:
Please don't rename methods for no reason, that's a breaking change

HermitOwO:
it's a method only used for tooltip
can leave an unused getRecipeTick() though

eerussianguy:
We aren't the only ones that use TFC code, any number of addon and pack developers rely on it. We aren't gonna rename public methods for no reason, I don't care if nobody uses it.

eerussianguy:
If you add a new method with a new function, sure, keep the new method, don't delete things.

PR #2595, opened 2023-12-28
Simple "external fauna" system

eerussianguy:


eerussianguy:
needs to get done properly...

PR #2620, opened 2024-01-23
Remove forge copper and iron dust tag from copper and iron powder

HermitOwO:
because it provides no practical benefit


eerussianguy:
Will have to hear a justification for this

alcatrazEscapee (approved):
I vaguely remember the conversation in discord about this; it's that we don't assign proper metal values to dusts (by any existing mod's standard), and so this tag just messes up with that and is removed. Tagging similar items for compatibility is good, tagging intentionally different ones because the tag "sounds right" is not.

So yes, I'm good with this.

---

If there's a use of this tag in TFC, we can tag them as powder, which for us more accurately reflects what they are. And like it or not we _are_ ignoring any conventions where 1 dust = 1 ingot (however silly that is with TFC involved), so there's no real reason to be _more_ inconsistent than necessary.

PR #2627, opened 2024-01-31
add missing translations from 1.20.x

Vi-brance:


alcatrazEscapee:
We're not maintaining 1.18 outside of important bug fixes anymore.

PR #2638, opened 2024-02-16
Fix TagLoaderMixin so it works with newer (Neo)Forge versions

ctrlaltmilk:
The method selector in TagLoaderMixin was fairly crude, which lead to errors in new (Neo)Forge versions. Somehow this only happens in a dev environment (not entirely sure why), but since the change is in both NeoForge and Forge latest, I think it's probably not a bad idea to patch it.

### DIFF
--- a/src/main/java/net/dries007/tfc/mixin/TagLoaderMixin.java
+++ b/src/main/java/net/dries007/tfc/mixin/TagLoaderMixin.java
@@ -7,7 +7,6 @@
 package net.dries007.tfc.mixin;
 
 import java.util.Collection;
-import net.minecraft.resources.ResourceLocation;
 import net.minecraft.tags.TagLoader;
 import org.spongepowered.asm.mixin.Dynamic;
 import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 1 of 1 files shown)


alcatrazEscapee:
I was under the impression that targeting lambda methods by name did not work in prod, due to the lambda not getting obf'd correctly or reliably. This mixin is `require = 0` as it's only use is to signal an error when running TFC self tests (which are only enabled via `-ea` which will almost _never_ be set in production). I also don't see how this raises errors in an (presumably, addon?) development environment, unless `-ea` is enabled, then is it simply targeting the wrong method and reports an external error where there is none?

ctrlaltmilk:
The reason this causes an error seems to be an odd quirk with Mixin: if a selector returns no matches, Mixin reruns it *without the type restrictions*. The wildcard then matches every method in the class, and complains about staticness mismatches. Here's the log if you want it: https://mclo.gs/LjOHECj

As far as I see it, there are three options:
1. Keep it as it is. This is obviously the simplest option, but it does mean that when updating to a newer loader version this mixin needs to be reworked anyway.
2. Change the target selector's signature. This obviously breaks compatibility with older loaders, which is a no-go.
3. Switch to matching on the lambda's name instead. AFAIK, this is [supported by Fabric](https://fabricmc.net/wiki/tutorial:mixin_tips#how_to_mixin_to_lambdas), and [Create seems to also do this in production](https://github.com/Creators-of-Create/Create/blob/mc1.20.1/dev/src/main/java/com/simibubi/create/foundation/mixin/MainMixin.java) so I feel comfortable relying on this matching correctly.

alcatrazEscapee:
> if a selector returns no matches, Mixin reruns it without the type restrictions.

Is this behavior documented anywhere? I've never heard that before.

We are not updating loader versions in dev due to the Neo/Lex Forge split, but ideally we'd like to support both loaders in dev and likewise I don't care about fabric at all. The issue with targeting by lambda name that I know is that it's an obf concern for prod, and/or the lambda name is not stable and may shift with recomp, due to patches. Hence targeting by signature. Any of that may be out of date.

That said, there may be a better way to target this mixin that doesn't involve having to target the lambda which is already non-ideal.

PR #2649, opened 2024-02-24
Fix for lamps remaining lit when emptying them with a bucket

stuartvc:
When using an empty bucket or barrel on a lit lamp, the fuel gets transferred to the bucket/barrel but the lamp remains lit.

While attempting a fix for this issue I noticed a few other inconsistencies.
- emptying an unlit lamp causes the fuel to get dumped on the ground and the lamp breaks (holding shift prevents this from happening, transferring the fuel successfully)
- adding fuel to a lit lamp causes the fuel to get dumped on the ground and breaks the lamp (adding fuel to an unlit lamp works successfully)

The changes in this commit fixes these issues as well

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/LampBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/LampBlockEntity.java
@@ -41,6 +41,11 @@ public class LampBlockEntity extends TickCounterBlockEntity implements FluidTank
     @Override
     public void fluidTankChanged()
     {
+        assert level != null;
+        if (!level.isClientSide && tank.isEmpty() && getBlockState().getValue(LampBlock.LIT) && level.getBlockEntity(getBlockPos(), TFCBlockEntities.LAMP.get()).isPresent())
+        {
+            level.setBlockAndUpdate(getBlockPos(), getBlockState().setValue(LampBlock.LIT, false));
+        }
         markForSync();
     }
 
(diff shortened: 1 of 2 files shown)


stuartvc:
Also, I'm not too familiar with client vs server logic. It seems like the fuel tank data only existed on the server side, so that's why there is some gating to only do the fluid transfer on server side. Maybe there is a better way to handle that, or make it so the fuel tank data is on client side as well.
And, I'm a little less confident in the `fluidTankChanged()` code that I added. This was mostly to handle the case when emptying a lamp while holding shift. Again, it only seems to work on server side, but I also needed to add the `level.getBlockEntity(...).isPresent())` code. When you break the lamp and generate a lamp item, the data gets copied to the lamp item. Such as fuel type and amount. But that caused updates to the fuel tank, which calls that callback, but by this point the LampBlock has already been destroyed, so `level.setBlockAndUpdate` was making changes to an air block. Again, feel free to change this if there is a better way to handle it.

alcatrazEscapee:
I've looked into this locally - this PR is welcome, but there were a few quicks of the implementation that didn't make sense to me (client side weirdness?). I eventually found that even if this appears to be fixing the issue, there's a much more underlying issue: 

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/5ab4abcc74dea8ad328588ad1b8da5e1bf4e04ea/src/main/java/net/dries007/tfc/common/blocks/devices/LampBlock.java#L132

This is emptying the lamp every time `getCloneItemStack` is invoked on client, i.e., because of rendering a Jade tooltip. This was causing the weird client behavior where the client did not appear to have fuel. I've fixed that, and incorporated parts of this PR, and this seems to fix all the listed issues.

stuartvc:
Wonderful, thanks for explaining the root cause

PR #2656, opened 2024-03-05
1.18.x

StarlightCraft:
more Traditional Chinese localiztions for Taiwanese


alcatrazEscapee:
Please target 1.20.x, we are not maintaining 1.18.x anymore.

PR #2657, opened 2024-03-05
Replace loot table copy/paste with custom number provider

ctrlaltmilk:
Cleans up the leaf loot tables - might also be useful for other things that can be piled? Let me know if there's another block this could apply to.


alcatrazEscapee:
Is there a practical benefit to having this provider other than it reduces the line count of our loot table JSONs?

ctrlaltmilk:
If/when loot table generation gets migrated to a Minecraft data generator, this makes constructing the loot table much easier, as well as unifying data in one place (e.g. a number in the resource generator has to match a number in the code or things break.)

alcatrazEscapee:
Migrating data generation to Forge/Neo/MC based data generation is not happening in any forseeable future. I can't see any other block this would apply to, because most pile blocks break one layer at a time, so in this respect this is an outlier.

alcatrazEscapee:
As aforementioned, this isn't wanted, there's no practical benefit I can see.

PR #2659, opened 2024-03-08
Allow in-world dirt to mud conversion

kimchiloof:
Thought it might be cleaner to make mud like vanilla, rather than (or in addition to) a crafting recipe.

- Server config option to enable/disable this feature, enabled by default
- Enables any IFluidHandler containing water to convert dirt to mud via right click (same as vanilla behavior).
- Amount of water required set to 100mb. This can be changed by modifying the `waterRequired` variable in IMudBlock.java
    - Should this be a config option?
- Was thinking of maybe removing the existing mud recipes? Not sure what is preferred though

(also quick question, was wondering what the reason was for `@Nullable private final Supplier<? extends Block> rooted;`  (and path/farmland) to be set to `@Nullable` in DirtBlock.java? It seems like every soil variant has a path, farmland, rooted versions, and none would be null? I'm probably missing something lol)


alcatrazEscapee:
This looks fine conceptually, but implementation wise it's backwards. The dirt block should be handling when it gets interacted with, and transform into mud. **Not** through the fluid container. This would restrict it to only interacting with fluid containers that use our helper methods (aka, only containers from TFC)

kimchiloof:
Good catch, seems to now work with any other fluid items that are based on IFluidHandler (though I had to use a deprecated function, it seems it's used elsewhere in TFC so hoping that's ok?). I've also added a server config option like rooted dirt creation and updated my original PR comment. Just wondering if you would prefer keeping the mud crafting recipes or removing them?

src/main/java/net/dries007/tfc/common/blocks/soil/DirtBlock.java
```diff
@@ -62,6 +65,9 @@ public BlockState getRooted()
     {
         return rooted.get().defaultBlockState();
     }
+    public BlockState getMud() {
```
eerussianguy: braces on newlines

eerussianguy (changes requested):

src/main/java/net/dries007/tfc/common/blocks/soil/TFCRootedDirtBlock.java
```diff
@@ -21,19 +27,26 @@
...
+        this.mud = mud;
     }
 
     TFCRootedDirtBlock(Properties properties, SoilBlockType dirtType, RegistrySoilVariant variant)
     {
-        this(properties, variant.getBlock(dirtType));
+        this(properties, variant.getBlock(dirtType), variant.getBlock(SoilBlockType.MUD));
+    }
+
+    public BlockState getMud()
```
alcatrazEscapee, line 48: Missing `@Override`

src/main/java/net/dries007/tfc/common/blocks/soil/DirtBlock.java
```diff
@@ -63,6 +66,11 @@ public BlockState getRooted()
         return rooted.get().defaultBlockState();
     }
 
+    public BlockState getMud()
```
alcatrazEscapee, line 70: Missing `@Override`

src/main/java/net/dries007/tfc/common/blocks/soil/IMudBlock.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        return InteractionResult.PASS;
+    }
+
+    /**
+     * Sets the block to mud and plays the appropriate particles and sound
+     * Particles like {@link net.minecraft.world.item.PotionItem}
+     */
+    default void setMud(BlockState mud, Level level, BlockPos pos, FluidStack water)
```
alcatrazEscapee: There's no real reason for this method to be split out from `transformToMud`

src/main/java/net/dries007/tfc/common/blocks/soil/IMudBlock.java
```diff
@@ -0,0 +1,107 @@
...
+
+        ItemStack stack = player.getItemInHand(hand);
+        final int waterRequired = 100;
+        final FluidStack water = new FluidStack(Fluids.WATER, waterRequired);
+
+        if (stack.getItem().equals(Items.WATER_BUCKET))
+        {
+            // Unable to deplete fluid in bucket, ignore interaction
+            return InteractionResult.PASS;
+        }
```
alcatrazEscapee: No. This shouldn't exist.

src/main/java/net/dries007/tfc/common/blocks/soil/IMudBlock.java
```diff
@@ -0,0 +1,107 @@
...
+        if (PotionUtils.getPotion(stack) == Potions.WATER)
+        {
+            if (!player.isCreative())
+            {
+                player.setItemInHand(hand, new ItemStack(Items.GLASS_BOTTLE));
+            }
+
+            setMud(mud, level, pos, water);
+            return InteractionResult.SUCCESS;
+        }
```
alcatrazEscapee: No, we don't interact with glass bottles, they're broken on many levels as fluid containers

src/main/java/net/dries007/tfc/common/blocks/soil/IMudBlock.java
```diff
@@ -0,0 +1,107 @@
...
+
+        IFluidHandler fluidHandler = Helpers.getCapability(stack, Capabilities.FLUID_ITEM);
+
+        if (fluidHandler != null && fluidHandler.getFluidInTank(0).containsFluid(water))
+        {
+            fluidHandler.drain(waterRequired, player.isCreative() ? IFluidHandler.FluidAction.SIMULATE : IFluidHandler.FluidAction.EXECUTE);
```
alcatrazEscapee: No. This is not how to interact with a fluid container.

- Simulate the drain first, then drain
- Don't query specific tanks directly, nor use `containsFluid` on specific tanks
- Only execute what you have a priori simulated

Look at `FluidHelpers.transferBetweenItemAndOther` for handling (all) buckets, not just hardcoded water ones, and handling arbitrary fluid containers properly.

https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/52642eba614929df5a12ea5aaa69293fc510b8dd/src/main/java/net/dries007/tfc/common/fluids/FluidHelpers.java#L207

alcatrazEscapee (changes requested):

eerussianguy:
Sorry for the piecemeal comments from us here, I meant to take a look at this earlier. We just are a little picky. Thanks for submitting work!

src/main/java/net/dries007/tfc/common/blocks/soil/IMudBlock.java
kimchiloof: I hope my new implementation addresses these issues? It's not entirely like `FluidHelpers.transferBetweenItemAndOther`, but works well in my testing. It still prevents buckets from converting dirt to mud as I feel that might be annoying if trying to place water sources on dirt.

eerussianguy (approved):

PR #2669, opened 2024-03-23
Bloomery Rework

### LINKED ISSUE #2179: Bloomery recipes are only unique up to their fluid input.
Due to the way `BloomeryRecipe#get` scans for recipes in combination with `BloomeryBlockEntity#addItemsFromWorld`. Might be possible to improve the scan to collect all recipes matching a primary input, then narrow down again to recipes matching a catalyst as well.

Might *also* want to consider ratio based catalyst usage, while we're at it.

Noticed in discord by a user adding the following recipe via CT:

```json
{
   "result":{
        "item": "create:andesite_alloy"},
        "fluid":{"ingredient":"tfc:metal/cast_iron",
        "amount": 10},
    "catalyst":{"item":"tfc:rock/loose/andesite"},
    "duration": 2000
}
```

alcatrazEscapee:
- Input and catalyst (charcoal) stacks are merged, and can store up to 16/layer of mixed
- Output is based on a ratio, so 100 mB Iron + 2 Charcoal = 1 Bloom - slight buff for using low grade ore, and the extra charcoal efficiency from denser iron is no more
- Updated guide book, and added bloomery game tests

Added a new game test framework using `GameTestAssertions`, which replaces usages of `GameTestHelper`. The latter, being mojang, is not extensible, and we cannot easily add our own helper methods (of which I would like many). So, a new assertion model, with some nice features, and the ability to extend it at will. Behind the scenes it is abusing `ThreadLocal<GameTestHelper>` to remove the need to handle context around.

Bloomery rework. Input and catalyst stacks are combined in the same capacity, and the recipe is based on the ratio of input fluid to catalyst. Currently set at 100mB + 2 Charcoal = 1 Bloom, which is a slight buff for low quality ore, and removes the fuel efficiency gain from casting ingots which was ultimately too overpowered vs. the effort and fuel it required to do.

Closes #2179 - with this rework that becomes much more difficult to do, and this achieves the intended purpose better.

Ported all existing tests to use the new framework.


src/main/java/net/dries007/tfc/common/blockentities/BloomeryBlockEntity.java
```diff
@@ -258,13 +259,13 @@ private void dumpItems()
...
-        BlockPos pos = getExternalBlock();
-        int x = pos.getX();
-        int y = pos.getY();
-        int z = pos.getZ();
-        inputStacks.forEach(i -> Containers.dropItemStack(level, x, y, z, i));
-        catalystStacks.forEach(i -> Containers.dropItemStack(level, x, y, z, i));
+        final BlockPos pos = getExternalBlock();
+        for (ItemStack stack : inputStacks)
+        {
+            Containers.dropItemStack(level, pos.getX(), pos.getY(), pos.getZ(), stack);
```
eerussianguy, line 265: Any reason we are using this over `Helpers#spawnItem`? I know it basically doesn't matter

src/main/java/net/dries007/tfc/common/blockentities/BloomeryBlockEntity.java
```diff
@@ -307,56 +308,56 @@ private void addItemsFromWorld(int capacity)
...
 
         if (cachedRecipe == null)
         {
-            for (ItemEntity entity : items)
+            assert inputStacks.isEmpty(); // If the cached recipe is null, we must have no inputs
+
+            final Collection<BloomeryRecipe> recipes = Helpers.getRecipes(level, TFCRecipeTypes.BLOOMERY).values();
+            final ItemStackInventory inventory = new ItemStackInventory();
+
+            loop:
```
eerussianguy, line 320: is this java syntax...?

eerussianguy (commented):
We'll also need to remove the datapack self test that verifies all bloomery recipes have unique inputs

src/main/java/net/dries007/tfc/common/blockentities/BloomeryBlockEntity.java
alcatrazEscapee: No, I was going to look at unifying usages of those helper methods, but I gave up. There is a slight difference - `Containers.dropItemStack` splits large stacks, and throws them a bit, `Helpers.spawnItem` spawns with no velocity. This is probably good for a cleanup later.

src/main/java/net/dries007/tfc/common/blockentities/BloomeryBlockEntity.java
alcatrazEscapee: Yeah, labels: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/branch.html

alcatrazEscapee:
> We'll also need to remove the datapack self test that verifies all bloomery recipes have unique inputs

No, as this PR doesn't remove that restriction, and is rather designed around that being in place, hence deciding not to do it that way.

PR #2684, opened 2024-04-24
Fix GUI textures so they look good together

ctrlaltmilk:
Small tweaks to the replacements for the vanilla GUI textures so the "Create World" screen looks nicer.


eerussianguy:
Thanks for taking our advice of "if you don't like it, fix it!"

PR #2685, opened 2024-04-25
Small visual improvements to windmills

HyperDashPony:
- Reverse the windmill blade model to match the rotation correctly
- Change the default blade model to align better with the axle, reducing visual weirdness with odd numbers of blades


src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -116,7 +116,7 @@ public WindmillBlockEntity(BlockEntityType<?> type, BlockPos pos, BlockState sta
         final Direction.Axis axis = state.getValue(WindmillBlock.AXIS);
 
         this.invalid = false;
-        this.node = new SourceNode(pos, Node.ofAxis(axis), Direction.fromAxisAndDirection(axis, Direction.AxisDirection.POSITIVE), 0f) {
```
alcatrazEscapee: Can this be done without changing the internal rotation? Just for better compat with existing setups, the only change should be visual here, I don't think this should be necessary if we're just flipping the model,

HyperDashPony: Yeah, I can do that instead

HyperDashPony:
@alcatrazEscapee I made all requested changes btw

alcatrazEscapee (approved):

PR #2686, opened 2024-04-25
Change /clearworld NOT_ORE preset to clear everything that isn't ore

ctrlaltmilk:
Before, the NOT_ORE preset cleared everything generated by veins. This worked mostly fine, but unfortunately dikes also use the vein system. This means a bunch of igneous rocks like diorite, gabbro, granite, etc. don't get cleared properly. Often, when I do use the /clearworld command I just want to expose all ore, which makes the NOT_ORE preset a little annoying. This PR makes that preset work for my purposes (checking oregen) which I'm assuming was its original intention.

### DIFF
--- a/src/main/java/net/dries007/tfc/common/commands/ClearWorldCommand.java
+++ b/src/main/java/net/dries007/tfc/common/commands/ClearWorldCommand.java
@@ -6,6 +6,8 @@
 
 package net.dries007.tfc.common.commands;
 
+import java.util.Collection;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


alcatrazEscapee:
No, the original intent was to strip anything generated through the vein features, as is obvious via the implementation.

> This PR makes that preset work for my purposes (checking oregen)

Unless you decide to add an ore vein to TFC from another mod, in which case you've made it worse for that use case.

ctrlaltmilk:
In that case, would it make sense to change "NOT_ORE" to "NOT_VEIN"? Most of the issue here was the name "NOT_ORE" made me expect just ore to be left, and I think changing the name would help resolve that expectation.

PR #2689, opened 2024-04-29
`BarrelRecipeWrapper` to not enforce people extend `BarrelBlockEntity.BarrelInventory`

Traister101:
Extracts the needed interface for barrel recipe containers into an interface that `BarrelBlockEntity.BarrelInventory` impliments. This is to allow people to easily support barrel recipes in things that aren't TFC  barrels such as the Barrel Compartments in the FirmaCiv addon

Edit: Typo

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/BarrelBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/BarrelBlockEntity.java
@@ -60,7 +60,6 @@ import net.dries007.tfc.common.capabilities.size.Weight;
 import net.dries007.tfc.common.container.BarrelContainer;
 import net.dries007.tfc.common.fluids.FluidHelpers;
 import net.dries007.tfc.common.recipes.BarrelRecipe;
-import net.dries007.tfc.common.recipes.BarrelRecipeWrapper;
 import net.dries007.tfc.common.recipes.SealedBarrelRecipe;
 import net.dries007.tfc.common.recipes.TFCRecipeTypes;
 import net.dries007.tfc.common.recipes.inventory.EmptyInventory;
@@ -198,7 +197,7 @@ public class BarrelBlockEntity extends TickableInventoryBlockEntity<BarrelBlockE
 
     public BarrelBlockEntity(BlockPos pos, BlockState state)
     {
-        super(TFCBlockEntities.BARREL.get(), pos, state, BarrelInventory::new, NAME);
+        super(TFCBlockEntities.BARREL.get(), pos, state, BarrelBlockEntity.BarrelInventory::new, NAME);
 
         sidedFluidInventory = new SidedHandler.Builder<>(inventory);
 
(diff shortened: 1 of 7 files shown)


src/main/java/net/dries007/tfc/common/recipes/BarrelRecipeWrapper.java
```diff
@@ -0,0 +1,28 @@
...
+
+package net.dries007.tfc.common.recipes;
+
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.fluids.capability.IFluidHandler;
+import net.minecraftforge.items.IItemHandlerModifiable;
+
+import net.dries007.tfc.common.recipes.inventory.EmptyInventory;
+
+public interface BarrelRecipeWrapper extends IItemHandlerModifiable, IFluidHandler, EmptyInventory
```
alcatrazEscapee: This should extend `Container`, not `EmptyInventory`, since the latter implies functionality, not an interface contract. It should also be named `BarrelInventory` and in the package `recipes.inventory` (with all the other inventory interfaces). It can be implemented explicitly via FQN where needed.

src/main/java/net/dries007/tfc/common/blockentities/BarrelBlockEntity.java
```diff
@@ -554,7 +555,7 @@ public long getRemainingTicks()
         return 0;
     }
 
-    public static class BarrelInventory implements DelegateItemHandler, DelegateFluidHandler, INBTSerializable<CompoundTag>, EmptyInventory, FluidTankCallback
+    public static class BarrelInventory implements DelegateItemHandler, DelegateFluidHandler, INBTSerializable<CompoundTag>, EmptyInventory, FluidTankCallback, BarrelRecipeWrapper
```
alcatrazEscapee: This then becomes
```suggestion
    public static class BarrelInventory implements DelegateItemHandler, DelegateFluidHandler, INBTSerializable<CompoundTag>, EmptyInventory, FluidTankCallback, net.dries007.tfc.recipes.inventory.BarrelInventory
```

src/main/java/net/dries007/tfc/common/recipes/BarrelRecipeWrapper.java
```diff
@@ -0,0 +1,28 @@
...
+{
+    /**
+     * Must be mutable to perform recipes despite sealed status
```
alcatrazEscapee:
```suggestion
     * Run an action while forcing the barrel's underlying state to be mutable, despite the sealed status. This should
     * only be used for updating the outputs on recipe completion, not for interacting with the barrel inventory via
     * external means - that should respect the sealed state.
```

alcatrazEscapee (changes requested):
Name/package change to be consistent with existing implementations of the same pattern, otherwise fine.

src/main/java/net/dries007/tfc/common/recipes/BarrelRecipeWrapper.java
Traister101: Oop good point on the `Container` not `EmptyInventory`

Russian said I should put it with the recipes & name it `BarrelRecipeWrapper`. Shall make the changes

Traister101:
Oh, I guess I should maybe say I made the changes @alcatrazEscapee

src/main/java/net/dries007/tfc/common/recipes/BarrelRecipeWrapper.java
alcatrazEscapee: Yeah, sorry, I saw that after seeing this PR, `RecipeWrapper` is Forge's abstraction for `Container` wrapping `IItemHandlerModifiable`, but we use a different naming scheme for our wrappers/inventories which was missed, which is understandable.

alcatrazEscapee (approved):

PR #2690, opened 2024-04-30
Add Alternate Windmill Blades

HyperDashPony:
- WindmillBladeItem now has a model associated with it upon construction
- 2 additional windmill blades styles were added: Lattice and Rustic
- Windmills can now render additional model details when a windmill has a full set of the same windmill blade
- updated windmill blade tags


src/main/java/net/dries007/tfc/client/model/entity/WindmillBladeModel.java
```diff
@@ -53,4 +53,7 @@ public void renderToBuffer(PoseStack poseStack, VertexConsumer vertexConsumer, i
         main.render(poseStack, vertexConsumer, packedLight, packedOverlay, 1f, 1f, 1f, alpha);
         blade.render(poseStack, vertexConsumer, packedLight, packedOverlay, red, green, blue, alpha);
     }
+
+    public void renderWindmillExtras(PoseStack poseStack, VertexConsumer vertexConsumer, int packedLight, int packedOverlay, float red, float green, float blue, float alpha){
```
eerussianguy: braces on newlines, or if theres no content do {}

src/main/java/net/dries007/tfc/client/render/blockentity/WindmillBlockEntityRenderer.java
```diff
@@ -67,28 +76,53 @@ public void render(WindmillBlockEntity windmill, float partialTick, PoseStack st
...
             stack.pushPose();
-            blade.setupAnim(windmill, partialTick, offsetAngle * i);
-            blade.renderToBuffer(stack, bufferSource.getBuffer(RenderType.entityCutout(BLADE_TEXTURE)), packedLight, packedOverlay, color[0], color[1], color[2], 1f);
+
+            // nudge to avoid Z-fighting
+            stack.translate(0.0001f*i, 0.0001f*i,0.0001f*i);
+            bladeModel.setupAnim(windmill, partialTick, offsetAngle * i);
+            bladeModel.renderToBuffer(stack, bufferSource.getBuffer(RenderType.entityCutoutNoCull(bladeTexture)), packedLight, packedOverlay, color[0], color[1], color[2], 1f);
+
+            if(bladeCount == WindmillBlockEntity.SLOTS && windmill.hasFullIdenticalSet)
```
eerussianguy: if (

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -220,4 +227,31 @@ public Node getRotationNode()
...
+                return false;
+            }
+        }
+
+        WindmillBladeItem item = (WindmillBladeItem) stack.getItem();
+        WindmillBladeItem.BladeModel model = item.getModel();
+
+        for (int i = 1; i < SLOTS; i++)
+        {
+            if(((WindmillBladeItem) inventory.getStackInSlot(i).getItem()).getModel() != model)
```
eerussianguy:
```
if (stack.getItem() instanceof WindmillBladeItem blade)
{
    ...
}
```

src/main/java/net/dries007/tfc/common/items/WindmillBladeItem.java
```diff
@@ -30,17 +30,32 @@ public class WindmillBladeItem extends Item
 
     private final DyeColor color;
 
-    public WindmillBladeItem(Properties properties, DyeColor color)
+    private final BladeModel model;
+
+    public static enum BladeModel
```
eerussianguy: move to bottom of class file

src/main/java/net/dries007/tfc/client/ClientEventHandler.java
```diff
@@ -70,14 +71,7 @@
 import net.minecraftforge.fml.event.lifecycle.FMLClientSetupEvent;
 import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
 import net.minecraftforge.registries.ForgeRegistries;
-
```
eerussianguy, line 73: why the import changes? what happened?

eerussianguy (changes requested):

src/main/java/net/dries007/tfc/client/ClientEventHandler.java
HyperDashPony: I messed up and it turned importing the entity models into model.entity.* for a minute, which it seemed like you didn't want

src/main/java/net/dries007/tfc/client/render/blockentity/WindmillBlockEntityRenderer.java
```diff
@@ -31,13 +33,20 @@
...
 {
     public static final ResourceLocation BLADE_TEXTURE = Helpers.identifier("textures/entity/misc/windmill_blade.png");
+    public static final ResourceLocation LATTICE_BLADE_TEXTURE = Helpers.identifier("textures/entity/misc/windmill_blade_lattice.png");
+
+    public static final ResourceLocation RUSTIC_BLADE_TEXTURE = Helpers.identifier("textures/entity/misc/windmill_blade_rustic.png");
     private static final float[] NO_COLOR = {1f, 1f, 1f, 1f};
 
     private final WindmillBladeModel blade;
+    private final WindmillBladeLatticeModel blade_lattice;
+    private final WindmillBladeRusticModel blade_rustic;
```
alcatrazEscapee, line 43:
```suggestion
    private final WindmillBladeLatticeModel bladeLattice;
    private final WindmillBladeRusticModel bladeRustic;
```
as per java convention, non-static fields should be `lowerCamelCase`

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -36,9 +39,11 @@ public class WindmillBlockEntity extends TickableInventoryBlockEntity<ItemStackH
     public static final int SLOTS = 5;
     public static final float MIN_SPEED = Mth.TWO_PI / (20 * 20);
     public static final float MAX_SPEED = Mth.TWO_PI / (8 * 20);
-
     private static final float LERP_SPEED = MIN_SPEED / (5 * 20);
 
+    // client-only
+    public boolean hasFullIdenticalSet = false;
```
alcatrazEscapee, line 45:
```suggestion
    // client-only
    private boolean hasFullIdenticalSet = false;
```
Prefer adding an accessor, can even use the existing method for it.

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -69,6 +74,8 @@ public static void clientTick(Level level, BlockPos pos, BlockState state, Windm
             : Math.max(targetSpeed, currentSpeed - LERP_SPEED);
 
         rotation.setSpeed(nextSpeed);
+
+        windmill.hasFullIdenticalSet = windmill.hasFullIdenticalSet();
```
alcatrazEscapee, line 78: Calling this on tick is poor - it's negating most of the impact from caching this in the first place - remove this call + assignment here.

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -220,4 +227,31 @@ public Node getRotationNode()
     {
         return node;
     }
+
+    public boolean hasFullIdenticalSet()
```
alcatrazEscapee, line 231: Two changes:

- This method should stay, but just be an accessor for the underlying variable - the public accessor shouldn't do any computation. Right now it's not clear that `windmill.hasFullIdenticalSet()` is computation, but `windmill.hasFulIdenticalSet` is cached (also why the variable should be made private)
- The _logic_ present here should be moved to a private method, e.g. `checkForFullIdenticalSet()`, which updates the value of this variable. This method needs to be called from two places:
  - `loadAdditional` - when we load the windmill from nbt
  - `setAndUpdateSlots` - when the inventory is modified

This way we only recompute it when the inventory changes - not every client tick, and it's clear what the accessor is for.

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -220,4 +227,31 @@ public Node getRotationNode()
...
+        ItemStack stack = inventory.getStackInSlot(0);
+
+        // catch anything that's not a windmill blade
+        for (int i = 1; i < SLOTS; i++)
+        {
+            if (!inventory.getStackInSlot(i).is(TFCTags.Items.ALL_WINDMILL_BLADES))
+            {
+                return false;
+            }
+        }
```
alcatrazEscapee, line 242: This is unnecessary because

1. It should be prevented from the `isItemValid` implementation above, so this should never catch anything, and
2. If it _does_, it's not actually preventing cast errors later in the method - the tag and then subsequent cast are not guaranteed to be 1-1 in the presence of a datapack.

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -220,4 +227,31 @@ public Node getRotationNode()
...
+        WindmillBladeItem item = (WindmillBladeItem) stack.getItem();
+        WindmillBladeItem.BladeModel model = item.getModel();
+
+        for (int i = 1; i < SLOTS; i++)
+        {
+            if (((WindmillBladeItem) inventory.getStackInSlot(i).getItem()).getModel() != model)
+            {
+                return false;
+            }
+        }
```
alcatrazEscapee, line 253: As alluded to above, both of these casts should be done defensively, i.e. `if (stack.getItem() instanceof WindmillBladeItem item)`. The prior tag check doesn't prevent these from erroring.

alcatrazEscapee (changes requested):
A more architectural change that is worth mentioning: we are essentially hardcoding the windmill to only accept `WindmillBladeItem` items. This is fine, given they have specific model functionality, but it does mean the `tfc:all_windmill_blades` tag is relatively worthless. We might want to remove that tag entirely, and replace it with `instanceof WindmillBladeItem`, ideally concentrated in a single method:

```java
class WindmillBladeItem {
    public static boolean isWindmillBlade(ItemStack stack) {
        return stack.getItem() instanceof WindmillBladeItem;
```

in case we ever want to reintroduce the tag / move to an interface in future. Tho, this is not necessary for this PR to go through, as is the tag is fine, just functionally unmodifiable by the end user.

HyperDashPony:
I will fix soon...

alcatrazEscapee:
#2744

PR #2700, opened 2024-05-10
Fix shading on jar shelf model

### LINKED ISSUE #2585: Shelves shade weird
image

ctrlaltmilk:
Fixes: #2585

Turns out having rotations, even 0° ones, messes up ambient occlusion on block models.


eerussianguy:
Subtlety!

PR #2707, opened 2024-05-18
Dyed Ceramic Jugs

MrHiTech123:
Adds dyed ceramic jugs, fixes a few other dyed ceramics- and fluid-related bugs


resources/format_lang.py
```diff
@@ -66,6 +66,16 @@ def format_lang(namespace: str, en_us, lang: str, validate: bool):
...
 def load(namespace: str, lang: str):
+    # Check if file exists before loading its JSON
+    try:
+        f = open('./src/main/resources/assets/%s/lang/%s.json' % (namespace, lang), 'r', encoding='utf-8')
+        f.close()
+    # If it doesn't, make the new file and have it contain an empty JSON object
+    except FileNotFoundError:
+        f = open('./src/main/resources/assets/%s/lang/%s.json' % (namespace, lang), 'w+', encoding='utf-8')
+        f.write('{}')
+        f.close()
```
alcatrazEscapee: Don't do this

- Use `with` when opening files
- Just check `os.path.isfile('...')` rather than opening and catching errors

resources/i18n.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, lang: str, validate: bool = False):
             if validate:
                 raise ValueError('Cannot validate book for lang %s, as resources/lang/%s.json does not exist' % (lang, lang))
             print('Writing default translation for language %s to %s' % (self.lang, self.lang_path))
-            with open(self.lang_path, 'w', encoding='utf-8') as f:
+            with open(self.lang_path, 'w+', encoding='utf-8') as f:
```
alcatrazEscapee: No? We aren't reading from it here.

resources/assets.py
```diff
@@ -987,8 +987,15 @@ def get_perspectives(model: str):
 
     for pottery in SIMPLE_UNFIRED_POTTERY:  # just the unfired item (fired is a vanilla item)
         rm.item_model(('ceramic', 'unfired_' + pottery)).with_lang(lang('Unfired %s', pottery))
-
+
```
alcatrazEscapee: Trailing space

src/main/java/net/dries007/tfc/client/ClientEventHandler.java
```diff
@@ -838,7 +839,9 @@ public static void registerColorHandlerItems(RegisterColorHandlersEvent.Item eve
         }
 
         TFCItems.MOLDS.values().forEach(reg -> event.register(new ContainedFluidModel.Colors(), reg.get()));
-        event.register(new ContainedFluidModel.Colors(), TFCItems.WOODEN_BUCKET.get(), TFCItems.BELL_MOLD.get(), TFCItems.FIRE_INGOT_MOLD.get(), TFCItems.JUG.get(), TFCItems.SILICA_GLASS_BOTTLE.get(), TFCItems.HEMATITIC_GLASS_BOTTLE.get(), TFCItems.VOLCANIC_GLASS_BOTTLE.get(), TFCItems.OLIVINE_GLASS_BOTTLE.get());
+        TFCItems.GLAZED_JUGS.values().forEach(reg -> event.register(new ContainedFluidModel.Colors(), reg.get()));
+
+        event.register(new ContainedFluidModel.Colors(), TFCItems.WOODEN_BUCKET.get(), TFCItems.BELL_MOLD.get(), TFCItems.FIRE_INGOT_MOLD.get(), TFCItems.JUG.get(), TFCItems.SILICA_GLASS_BOTTLE.get(), TFCItems.HEMATITIC_GLASS_BOTTLE.get(), TFCItems.VOLCANIC_GLASS_BOTTLE.get(), TFCItems.OLIVINE_GLASS_BOTTLE.get(), TFCItems.RED_STEEL_BUCKET.get(), TFCItems.BLUE_STEEL_BUCKET.get());
```
alcatrazEscapee, line 844: Are red/blue steel bucket models broken?

alcatrazEscapee (changes requested):
Why should we add this?

resources/i18n.py
MrHiTech123: Oh whoops.

TIL that's what w+ means. I thought w meant write to an existing file, and w+ meant write to an existing file, or create a new file if it doesn't exist.

My bad—nevermind. Will remove

src/main/java/net/dries007/tfc/client/ClientEventHandler.java
MrHiTech123: They don't show the color of the liquid inside
image

This fixes that

MrHiTech123:
You're asking why in general?

1. Dyeing pottery is a nice way to personalize it, which also causes the player to be more invested in it: a normal TFC jug is just a tool to carry water. But a jug dyed your favorite color? That's _your personal_ tool to carry water
2. For consistency's sake: you can dye vessels and large vessels, so why not jugs too?
3. Because it's already coded, adds something to the experience to at least one player, and there's no particularly compelling reason not to add it?

MrHiTech123:
Though I will say I was planning on adding dyeable pots too, so if that's too superfluous for base TFC then maybe it would make more sense to bundle these both as an addon

eerussianguy:
Dyeable pots would be out of scope in my mind since that requires adding 16 new firepit blocks or changing how those work.

eerussianguy:
> [Switched the Light Blue and Cyan textures for Jugs and Small Vessels](https://github.com/TerraFirmaCraft/TerraFirmaCraft/pull/2707/commits/0aae8062017d123ba02161833200a01f1ff57ebf)

So I would rather we be consistent here and go for an actual rework of the textures and not just someone eyeballing that they think light blue and cyan are switched. Maybe we can call in someone from Discord who is an artist.

MrHiTech123:
> Dyeable pots would be out of scope in my mind since that requires adding 16 new firepit blocks or changing how those work.

Makes sense, so I think I'll do the addon thing (unless y'all have any objections)

> So I would rather we be consistent here and go for an actual rework of the textures and not just someone eyeballing that they think light blue and cyan are switched. Maybe we can call in someone from Discord who is an artist.

I don't really see what the benefit of this would be, since I'd say that the cyan textures are pretty great for light blue, and vice versa, but I agree that having a real artist weigh in on this issue in general would be a good idea. Out of curiosity, do you disagree with the idea that they're switched?

eerussianguy:
> Out of curiosity, do you disagree with the idea that they're switched?

Philosophically, yes, because it's not like when they were drawn someone swapped them by accident. It's being phrased as if it's deliberate. But I accept that this is a difference that perhaps only I care about.

PR #2710, opened 2024-05-21
Add getters for `BranchDirection`

ctrlaltmilk:
Just needed this for an addon and I'd prefer there being an API surface over having to fall back on accessor mixins.


alcatrazEscapee (changes requested):
Stylistic difference, but consistent with the rest of the class, can these be named `dx()`, `dy()`, and `dz()` rather than the `getDx` style.

PR #2714, opened 2024-05-30
Don't collapse caves if broken block is supported

cominixo:
This PR changes the collapsing behavior so that blocks that are already supported don't trigger collapses when broken.

Breaking blocks that are already supported shouldn't have an effect on the others collapsing, it can make mining frustrating and unfair.
For example, breaking a block while mining while you're near a cave (that you might not even be able to see) would cause a collapse. Collapsing a cave once can make it pretty much irrecoverable, which can become pretty frustrating and generally not fun.


alcatrazEscapee:
No. The way it works now is intentional and not changing.

> For example, breaking a block while mining while you're near a cave (that you might not even be able to see) would cause a collapse.

No it won't, because caves are naturally supported and won't start collapses. Only if you're already mining in an unsupported area.

PR #2727, opened 2024-06-13
Dynamic Ingredients

### LINKED ISSUE #2658: Support for NBT Tools
It is hard to integrate TFC into other mods that use NBT tools (such as Tetra, Slient Gear and Tinker's Construct) since TFC heavily depends on item tags for tools. 

It would be great if there was some method to extend TFC with datapack allow things like tree feeling, the anvil, and scraping with NBT tools.

alcatrazEscapee:
Name is a bit misleading - it refers to an `Ingredient`, standalone, that has been loaded through a `RegisteredDataManager`. So instead of referencing tag names for tool actions that are potentially desired to be NBT-specific, we can reference an ingredient.

This is prototype and untested. Questions abound:

- Does this actually solve the issue for anyone or am I wasting my time
- If yes, are there other use cases for this people desire

Fixes #2658


eerussianguy:
So what I'm missing is why this needs to be data driven I think

alcatrazEscapee:
Decided against this thinking w.r.t future changes to how we identify tools, `ItemAbilities` ? For ones we can reasonably expect mods to provide. Otherwise I'm not convinced by the ROI of continuing this until at least 1.21

PR #2730, opened 2024-06-15
JEI Recipe Click Areas for TFC menus such as the Fire Pit

Traister101:
Adds JEI recipe click areas, the clickable secions in a screen which open JEI to some recipe categories.
The Barrel, Pot and Anvil render a question mark in the screen where the clickable area is when JEI is present. The other screens have the clickable area on a progress bar or similar arrow.
Follow up PR is planned for Recipe Transfer Handlers. This would allow the `+` button in JEI to fill the recipe.


Traister101:
Hmm. I'm not sure what exactly to do about that. `ModList.get().isLoaded("jei")` in `Helpers` is causing it to explode when testing. Could just have each screen store it's own static perhaps? Will leave it as is for now

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -202,6 +205,10 @@ public static ResourceLocation resourceLocation(String domain, String path)
         return new ResourceLocation(domain, path);
     }
 
+    public static boolean isJEIEnabled(){
+        return JEI;
+    }
```
alcatrazEscapee:
```suggestion
    public static boolean isJEIEnabled()
    {
        return JEI;
    }
```

src/main/java/net/dries007/tfc/util/Helpers.java
```diff
@@ -167,6 +168,8 @@ public final class Helpers
     private static final int PRIME_X = 501125321;
     private static final int PRIME_Y = 1136930381;
 
+    private static final boolean JEI = ModList.get().isLoaded("jei");
```
alcatrazEscapee:
```suggestion
    private static final boolean JEI = !BOOTSTRAP_ENVIRONMENT && ModList.get().isLoaded("jei");
```
This should fix the unit tests breaking I believe, since it's not a classload error, just failing to call `isLoaded()` on ` null`

alcatrazEscapee (changes requested):
I don't like the name `TFCInventoryGuiHandler` because it's remarkably undescriptive for what it's actually being used for, but I can't think of a better one that's concise and descriptive so it's fine. Two small notes otherwise this looks straightforward enough.

src/main/java/net/dries007/tfc/util/Helpers.java
Traister101: Ah whoops. Douno how that escaped the auto formatter

Traister101:
> I don't like the name `TFCInventoryGuiHandler` because it's remarkably undescriptive for what it's actually being used for, but I can't think of a better one that's concise and descriptive so it's fine. Two small notes otherwise this looks straightforward enough.

Yeah, that whole class is pretty much yanked right from 1.12. All it does is return the extra inventory button dimensions TFC adds. Could name it something like `TFCInventoryTabHandler` perhaps after JEIs `InventoryEffectRendererGuiHandler` but yeah, it's a hard one to name.

PR #2732, opened 2024-06-19
JEI Recipe Transfer Handlers for containers

Traister101:
This is the follow up PR mentioned in #2730
- It adds heating recipe transferring for the Fire Pit, Grill and Crucible.
- Casting recipe transferring (transfer of molds) for Crucible
- Welding recipe transferring for the anvil respecting anvil tier
- Anvil (forging) recipe transferring which picks the anvil pattern respecting anvil tier
- Pot recipe transferring respecting fluid content
- Sealed barrel recipe transferring (instant recipes purposely left out)


src/main/java/net/dries007/tfc/compat/jei/transfer/AnvilRecipeTransferInfo.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public List<Slot> getRecipeSlots(AnvilContainer container, AnvilRecipe recipe)
+    {
+        return List.of(container.getSlot(AnvilBlockEntity.SLOT_INPUT_MAIN), container.getSlot(AnvilBlockEntity.SLOT_INPUT_SECOND), container.getSlot(AnvilBlockEntity.SLOT_CATALYST));
+    }
+
+    @Override
+    public List<Slot> getInventorySlots(AnvilContainer container, AnvilRecipe recipe)
+    {
+        return IntStream.range(4, 4 + Inventory.INVENTORY_SIZE).mapToObj(container::getSlot).collect(Collectors.toList());
```
alcatrazEscapee:
```suggestion
        return IntStream.range(4, 4 + Inventory.INVENTORY_SIZE).mapToObj(container::getSlot).toList();
```

src/main/java/net/dries007/tfc/network/SelectAnvilPlan.java
```diff
@@ -0,0 +1,56 @@
...
+    void encode(FriendlyByteBuf buffer)
+    {
+        buffer.writeResourceLocation(recipeId);
+    }
+
+    void handle(@Nullable ServerPlayer player)
+    {
+        if (player != null)
+        {
+            if (player.containerMenu instanceof AnvilContainer)
```
alcatrazEscapee:
```suggestion
            if (player.containerMenu instanceof AnvilContainer anvilContainer)
```

src/main/java/net/dries007/tfc/network/SelectAnvilPlan.java
```diff
@@ -0,0 +1,56 @@
...
+            {
+                AnvilRecipe recipe = Helpers.getRecipes(player.level(), TFCRecipeTypes.ANVIL).get(recipeId);
+                InventoryBlockEntity<?> blockEntity = ((BlockEntityContainer<?>) player.containerMenu).getBlockEntity();
+                if (blockEntity instanceof AnvilBlockEntity anvilBlockEntity)
+                {
+                    anvilBlockEntity.chooseRecipe(recipe);
+                }
```
alcatrazEscapee:
```suggestion
                anvilContainer.getBlockEntity().choseRecipe(recipe);
```

alcatrazEscapee (changes requested):
A couple notes, and one larger request. There seems to be a lot of boilerplate in each transfer handler - can we create a single base abstract class `BaseTransferHandler`, with fields and implementations for `getContainerClass()`, `getMenuType`, and `getRecipeType`?

Possibly even a `int[] recipeSlots` and `int containerSlots` to allow implementation of `getRecipeSlots` and `getInventorySlots`.

All of these classes could be reduced down to 1 or 2 method overrides, and a constructor call imo

Traister101:
> A couple notes, and one larger request. There seems to be a lot of boilerplate in each transfer handler - can we create a single base abstract class `BaseTransferHandler`, with fields and implementations for `getContainerClass()`, `getMenuType`, and `getRecipeType`?
> 
> Possibly even a `int[] recipeSlots` and `int containerSlots` to allow implementation of `getRecipeSlots` and `getInventorySlots`.
> 
> All of these classes could be reduced down to 1 or 2 method overrides, and a constructor call imo

There's a bit of boilerplate yeah. I don't really see much value in abstracting this out when there's 2 handlers, both wrapping a provided transfer handler. If you are referring to the info objects of which we have 4 `AnvilRecipeTransferInfo`, `BarrelTransferInfo`, `PotTransferInfo` and `WeldingRecipeTransferInfo` I also see little value in an abstraction there. Each TransferInfo exists because it's necessary, if they weren't we could just use the JEI provided `IRecipeTransferInfo` objects. For the two corresponding to the anvil they do 2 things
1. They skip over the Hammer slot for the recipe slots. 
2. They account for recipe specific errors namely the tier requirement.

The `AnvilRecipeTransferHandler` exists because we want to select the anvil plan as part of the transfer process. The only other way to do this requires re-implementing all the required transfer logic and networking JEI uses or we compile against the full JEI jar as I mentioned on Discord. Because wrapping a handler seems like the best approach we must use the JEI provided handler which can only be customized via a `IRecipeTransferInfo` object, hence `AnvilRecipeTransferInfo`.

If we really want to abstract the 2 transfer handlers I can do so, but there's no way to allow the transfer handlers to provide the slots. The `IRecipeTransferInfo` are the only way to do that without re-implementing or depending on things from JEI's full Jar.

alcatrazEscapee:
Nevermind then I'll do it myself.

Traister101:
Oh... Okay? I said I can do so cause I was, and have done so. Pushed a bit after you closed this as it turns out.

PR #2758, opened 2024-08-08
1.21.x weathering

alekiponi:
- add an interface for climate-affected weathering blocks
- add weathering metal blocks, stairs, and slabs
- textures/data for all necessary blocks


src/main/java/net/dries007/tfc/ForgeEventHandler.java
```diff
@@ -1587,4 +1592,28 @@ public static void onCropsGrow(CropGrowEvent.Pre event)
             }
         }
     }
+
+    public static void onScrapeWeatherable(BlockEvent.BlockToolModificationEvent event)
```
alcatrazEscapee, line 1596: This event supports simulation, this event handler doesn't. In general this is the wrong place to do any of these modifications, which should be reliant on the item exposing itself as capable of performing this tool action - see, i.e. `AxeItem` which provides `AXE_SCRAPE`, and handles the sound itself, as well as damaging the tool.

src/main/java/net/dries007/tfc/util/Metal.java
```diff
@@ -348,12 +401,42 @@ public boolean hasMold()
...
+                case WEATHERED -> metal == DEFAULT_WEATHERING || metal == ALL_WEATHERING;
+                case ALL -> metal.ordinal() >= ALL.ordinal();
+                case DEFAULT -> metal.ordinal() >= DEFAULT.ordinal();
+                case INGOT_ONLY -> true;
+                default -> throw new AssertionError("Invalid choice for a metal type " + this);
+            };
+        }
+        }
+
+    public enum WeatheringType
```
alcatrazEscapee, line 427: This is problematic, it's restricting the resistance values arbitrarily to a fixed set of enum values. Also the `NONE` should not exist, as we shouldn't be registering alternate blocks in that situation, and the default metal block should not be registered as weatherable at all

src/main/java/net/dries007/tfc/common/blocks/IClimateWeatheringBlock.java
```diff
@@ -0,0 +1,194 @@
...
+
+    default void applyChangeOverTime(BlockState blockState, ServerLevel serverLevel, BlockPos pos, RandomSource random)
+    {
+        if (blockState.getBlock() instanceof IClimateWeatheringBlock climateWeatheringBlock && climateWeatheringBlock.getAge() == TFCWeatherState.FULLY_WEATHERED)
+        {
+            return;
+        }
+
+        // designed to slow weathering after most blocks are exposed
+        Iterator<BlockPos> var8 = BlockPos.withinManhattan(pos, 4, 4, 4).iterator();
```
alcatrazEscapee, line 47: This should be using enhanced `for(BlockPos pos : BlockPos.withinManhattan(origin, 4, 4, 4))`

src/main/java/net/dries007/tfc/common/blocks/IClimateWeatheringBlock.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    static float getAgeAffectedModifier(TFCWeatherState weatherState)
+    {
+        return AGE_MODIFIERS[weatherState.ordinal()];
+    }
+
+    TFCWeatherState getAge();
+
+    enum TFCWeatherState
```
alcatrazEscapee, line 159: This should not be prefixed with `TFC`, it's unnecessary.

src/main/java/net/dries007/tfc/common/blocks/IClimateWeatheringBlock.java
```diff
@@ -0,0 +1,194 @@
...
+    {
+        UNAFFECTED,
+        EXPOSED,
+        WEATHERED,
+        FULLY_WEATHERED;
+
+        static TFCWeatherState getByOrdinal(int ordinal)
+        {
+            ordinal = Mth.clamp(ordinal, 0, 3);
+            return TFCWeatherState.values()[ordinal];
```
alcatrazEscapee, line 169: This is wasteful, as it creates a copy of `values()` every single invocation.

src/main/java/net/dries007/tfc/common/blocks/IClimateWeatheringBlock.java
```diff
@@ -0,0 +1,194 @@
...
+
+public interface IClimateWeatheringBlock
+{
+    float[] AGE_MODIFIERS = {0.5f, 1.0f, 1.0f, 1.0f};
+
+    Block getNext();
+
+    BlockState getNext(BlockState state);
+
+    BlockState getPrevious(BlockState state);
```
alcatrazEscapee, line 35: Both of these overloads are unnecessary, and can be replaced by invoking `Helpers.copyProperties` to copy state properties.

src/main/java/net/dries007/tfc/common/blocks/IClimateWeatheringBlock.java
```diff
@@ -0,0 +1,194 @@
...
+            if (randomFloat < overallChance && randomFloat < getMaterialModifier())
+            {
+                serverLevel.setBlockAndUpdate(pos, getNext(blockState));
+                return;
+            }
+        }
+
+        // do "drip" weathering
+        int adjacentCount = 0;
+        for (Direction direction : Direction.Plane.HORIZONTAL.stream().toList())
```
alcatrazEscapee, line 96:
```suggestion
        for (Direction direction : Direction.Plane.HORIZONTAL)
```
Wasteful, does a list copy

src/main/java/net/dries007/tfc/common/blocks/IClimateWeatheringBlock.java
```diff
@@ -0,0 +1,194 @@
...
+
+import net.dries007.tfc.common.TFCTags;
+import net.dries007.tfc.common.items.TFCItems;
+import net.dries007.tfc.util.climate.Climate;
+
+public interface IClimateWeatheringBlock
+{
+    float[] AGE_MODIFIERS = {0.5f, 1.0f, 1.0f, 1.0f};
+
+    Block getNext();
```
alcatrazEscapee, line 31: This should be using the Neo data map for weathering - I think doing it this way exposes us to other mods in a compatible way. Thus `getNext()` and `getPrevious()` are both unnecessary.

src/main/java/net/dries007/tfc/common/blocks/WeatheringMetalBlock.java
```diff
@@ -0,0 +1,94 @@
...
+    public WeatheringMetalBlock(Properties properties, TFCWeatherState weatherState, RegistryMetal metal)
+    {
+        super(properties);
+        this.weatherState = weatherState;
+        this.metal = metal;
+    }
+
+    @Override
+    public void randomTick(BlockState state, ServerLevel serverLevel, BlockPos pos, RandomSource random) {
+        if (random.nextFloat() < BASE_METAL_CHANCE) {
```
alcatrazEscapee, line 34: This chance check should be in `IClimateWeatheringBlock`, since it's common in all our implementations

src/main/java/net/dries007/tfc/common/blocks/WeatheringMetalBlock.java
```diff
@@ -0,0 +1,94 @@
...
+
+import net.dries007.tfc.common.TFCTags;
+import net.dries007.tfc.util.Metal;
+import net.dries007.tfc.util.registry.RegistryMetal;
+
+public class WeatheringMetalBlock extends Block implements IClimateWeatheringBlock
+{
+    static final float BASE_METAL_CHANCE = 0.015f;
+    TFCWeatherState weatherState;
+    RegistryMetal metal;
```
alcatrazEscapee, line 23: Rather than storing the `metal`, this should store a `float materialModifier`, which is set by the constructor. This makes this more accessible to addons, prevents the need to wrap it in an enum for fixed values, and prevents the need for this to take a `RegistryMetal`

alcatrazEscapee (changes requested):

PR #2762, opened 2024-08-15
1.21.x logpile upgrade

alekiponi:
- logpiles have 16 slots of size 1
- items cascade to the lowest slot, including from logpiles above
- logpiles have a visual and collision shape that matches the amount of logs they hold
- logpiles no longer have an inventory screen
- inserting logs that won't fit in a logpile will push them up, making them act like ingot piles
- datafixing for old logpiles even though it's probably not necessary


alekiponi:
still missing shift+click with empty hand resulting in a full stack extraction

resources/assets.py
```diff
@@ -488,9 +488,14 @@ def rock_lang(_lhs: str, _rhs: str):
...
     rm.blockstate('charcoal_forge', variants=dict((('heat_level=%d' % i), {'model': 'tfc:block/charcoal_forge/heat_%d' % i}) for i in range(0, 7 + 1))).with_lang(lang('Forge')).with_block_loot('7 minecraft:charcoal')
-    rm.blockstate('log_pile', variants={'axis=x': {'model': 'tfc:block/log_pile', 'y': 90, 'x': 90}, 'axis=z': {'model': 'tfc:block/log_pile', 'x': 90}}) \
-        .with_block_model(textures={'side': 'tfc:block/log_pile_side', 'end': 'tfc:block/log_pile_front'}, parent='minecraft:block/cube_column_horizontal').with_lang(lang('Log Pile'))
-    rm.blockstate('burning_log_pile', model='tfc:block/burning_log_pile').with_block_model(parent='minecraft:block/cube_all', textures={'all': 'tfc:block/devices/charcoal_forge/lit'}).with_lang(lang('Burning Log Pile'))
+
+    #rm.blockstate('log_pile', variants=dict((('count=%d' % i), ('axis='+a), {'y': ('90' if a == 'x' else '0'), 'model': 'tfc:block/log_pile/log_pile_%d' % i}) for i in range(1, 1 + 16) for a in ['x','z'])).with_lang(lang('Log Pile'))
```
eerussianguy: delete commented out code

resources/assets.py
```diff
@@ -2388,6 +2393,43 @@ def door_blockstate(base: str) -> JsonObject:
         'facing=west,half=upper,hinge=right,open=true': {'model': top_right_open, 'y': 90}
     }
 
+def log_pile_blockstate() -> JsonObject:
```
eerussianguy, line 2392: Why did this need to be permuted like this?

src/main/java/net/dries007/tfc/common/blocks/devices/LogPileBlock.java
```diff
@@ -82,30 +130,111 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+            extractFromTop(level, pos.above(), player, all);
+        }
+        else
+        {
+            level.getBlockEntity(pos, TFCBlockEntities.LOG_PILE.get()).ifPresent(logPile -> {
+                if (!level.isClientSide)
+                {
+                    for (int i = 0; i < LogPileBlockEntity.SLOTS; i++)
+                    {
+                        if (!logPile.getInventory().getStackInSlot(i).isEmpty())
```
eerussianguy: throw this in a local variable

resources/assets.py
alekiponi: This is still just a prototype, I just put it in the first way I thought of since the end goal is to have dynamic models

PR #2770, opened 2024-08-25
Seasonal Rainfall and Groundwater Basics

Therighthon:
Adds 2 parameters accessible at worldgen:
- Base Groundwater: theoretically ranges 0 to 500, higher near rivers, rainfall no longer increases at rivers
- Rainfall variance: highly negative near west coasts, highly positive on east coasts. Negative indicates wet winter, positive a wet summer, ranges -1 to 1
These are used to calculate:
- Monthly rainfall: Triangle function with average rainfall at the center, and rainfall variance * avg rain as the amplitude. Peaks January 1 and July 1, should be used to calculate if it is raining at a given location at a given time. Ranges from 0 to 1000
- Groundwater: the sum of avg rainfall and base groundwater, used for worldgen instead of rainfall. Planning to re-add rainfall based climate parameters for things like kaolinite (though kaolin and clay generating near rivers is maybe desirable anyways). Ranges from 0-500
- Monthly groundwater: the sum of base groundwater and monthly rainfall, this is used for colormaps instead of rainfall. Ranges from 0-1000, colormaps still need to be redesigned

Makes all of these parameters visible in the debug menu w/ additional info on


src/main/java/net/dries007/tfc/world/region/RegionGenerator.java
```diff
@@ -43,10 +43,16 @@ private static Noise2D baseNoise(boolean axisIsX, float scale, float constant)
             (x, z) -> triangle(frequency, z);
     }
 
+    private static Noise2D simpleBaseNoise()
+    {
+        return (x, z) -> 0;
+    }
+
```
alcatrazEscapee, line 50: Not needed

src/main/java/net/dries007/tfc/world/region/RegionGenerator.java
```diff
@@ -87,6 +93,12 @@ public RegionGenerator(Settings settings, RandomSource random)
                 .spread(0.15f)
                 .scaled(-80f, 40f)); // Bias slightly negative, as we bias near-ocean areas to be positive rainfall, so this encourages deserts inland
 
+        this.rainfallVarianceNoise = simpleBaseNoise()
+            .add(new OpenSimplex2D(random.nextInt())
+                .octaves(2)
+                .spread(0.3f)
+                .scaled(-.2f, 0.2f));
```
alcatrazEscapee, line 100:
```suggestion
        this.rainfallVarianceNoise = new OpenSimplex2D(random.nextInt())
            .octaves(2)
            .spread(0.3f)
            .scaled(-.2f, 0.2f);
```
No need to do the simple base noise and add, can just use the simplex noise directly.

src/main/java/net/dries007/tfc/client/ClimateRenderCache.java
```diff
@@ -49,6 +54,11 @@ public void onClientTick()
             temperature = model.getTemperature(level, pos);
             rainfall = model.getRainfall(level, pos);
             wind = model.getWind(level, pos);
+            rainVariance = Climate.getRainVariance(level, pos);
+            monthlyRainfall = Climate.getMonthlyRainfall(level, pos);
+            baseGroundwater = Climate.getBaseGroundwater(level, pos);
+            groundwater = Climate.getGroundwater(level, pos);
+            monthlyGroundwater = Climate.getMonthlyGroundwater(level, pos);
```
alcatrazEscapee, line 61: These could be using the method on `ClimateModel` (i.e. see the above ones) because the model is already queried - saves the world tracker lookup.

src/main/java/net/dries007/tfc/util/climate/ClimateModel.java
```diff
@@ -98,6 +100,42 @@ default float getRainfall(LevelReader level, BlockPos pos, long calendarTicks, i
     float getRainfall(LevelReader level, BlockPos pos);
 
     /**
+     * Get the annual variance in rainfall for a given position.
+     * Positive values indicate wet summers, Negative values indicate wet winters.
+     *
+     * @return The annual variance in the immediate rate of rainfall, in percentage of annual. Should be in the range [-1, 1]
+     */
+    float getRainVariance(LevelReader level, BlockPos pos);
```
alcatrazEscapee, line 108: Ideally these should be default methods here, because it saves `BiomeBasedClimateModel` from having to override them. And most of these have reasonable defaults that can delegate to other properties.

src/main/java/net/dries007/tfc/world/chunkdata/RegionChunkDataGenerator.java
```diff
@@ -110,6 +110,8 @@ public ChunkData generate(ChunkData data)
         final Region.Point point11 = regionGenerator.getOrCreateRegionPoint(gridX + 1, gridZ + 1);
 
         final LerpFloatLayer rainfallGridLayer = new LerpFloatLayer(point00.rainfall, point01.rainfall, point10.rainfall, point11.rainfall);
+        final LerpFloatLayer rainfallVarianceGridLayer = new LerpFloatLayer(point00.rainfallVariance, point01.rainfallVariance, point10.rainfallVariance, point11.rainfallVariance);
+        final LerpFloatLayer baseGroundwaterGridLayer = new LerpFloatLayer(0f, 0f, 0f, 0f);
```
alcatrazEscapee, line 114: Since this one isn't being queried from a grid scale, downsided, and then modified, we don't need this
```suggestion
```

src/main/java/net/dries007/tfc/world/chunkdata/RegionChunkDataGenerator.java
```diff
@@ -126,17 +128,20 @@ public ChunkData generate(ChunkData data)
 
         // The base rainfall and temperature layers, scaled down to chunk resolution
         LerpFloatLayer rainfallLayer = rainfallGridLayer.scaled(deltaX, deltaZ, dG);
+        LerpFloatLayer rainfallVarianceLayer = rainfallVarianceGridLayer.scaled(deltaX, deltaZ, dG);
+        LerpFloatLayer baseGroundwaterLayer = baseGroundwaterGridLayer.scaled(deltaX, deltaZ, dG);
```
alcatrazEscapee, line 132: Same here, `baseGroundwaterLayer` would be all zeros.
```suggestion
```

src/main/java/net/dries007/tfc/world/chunkdata/RegionChunkDataGenerator.java
```diff
@@ -126,17 +128,20 @@ public ChunkData generate(ChunkData data)
...
         // The radius that the partition point will find rivers is approx ~5 grid distance (~500 blocks), so we should be fine to influence
         // on a smaller scale of ~100 blocks around the river
-        float rainfall00 = rainfallLayer.value00();
-        float rainfall01 = rainfallLayer.value01();
-        float rainfall10 = rainfallLayer.value10();
-        float rainfall11 = rainfallLayer.value11();
+        float groundwater00 = baseGroundwaterLayer.value00();
+        float groundwater01 = baseGroundwaterLayer.value01();
+        float groundwater10 = baseGroundwaterLayer.value10();
+        float groundwater11 = baseGroundwaterLayer.value11();
```
alcatrazEscapee, line 144: Which means these could be initialized to all zeros
```suggestion
        float groundwater00 = 0;
        float groundwater01 = 0;
        float groundwater10 = 0;
        float groundwater11 = 0;
```

PR #2781, opened 2024-09-06
Port biome noise viusalizer, add new desert biomes

Therighthon:
New Biomes:
- Mud Flats - Very flat area covered in new "Cracked Earth" soil variants
- Salt Flats - Very flat area covered in new "Salted Earth" soil variants, which drop cracked earth and salt when dug
- Dune Sea - Sandy biome with regular sinusoidal ridges
- Grassy Dunes - Similar to Dune Sea, but with grass near the top of dunes

All new biomes are barren of plant life, except the grassy dunes. Dune biomes replace a few low and medium height biomes in dry regions, salt and mud flats replace only low biomes, but will place other medium/low height biomes around them to create the impression of a dried lake bed.


resources/assets.py
```diff
@@ -12,6 +12,7 @@
 TAG_SHARP = 'tfc:tools/sharp'
 
 STICKS_WHEN_NOT_SHEARED = loot_tables.alternatives({
+    'name': 'minecraft:stick',
```
alcatrazEscapee: Typo?

src/data/java/net/dries007/tfc/data/providers/BuiltinRecipes.java
```diff
@@ -271,6 +271,12 @@ public void accept(ResourceLocation id, Recipe<?> recipe, @Nullable AdvancementH
             add(new LandslideRecipe(BlockIngredient.of(
                 blocks.get(SoilBlockType.MUD).get()
             ), blocks.get(SoilBlockType.MUD).get().defaultBlockState()));
+            add(new LandslideRecipe(BlockIngredient.of(
+                blocks.get(SoilBlockType.CRACKED_EARTH).get()
+            ), blocks.get(SoilBlockType.CRACKED_EARTH).get().defaultBlockState()));
+            add(new LandslideRecipe(BlockIngredient.of(
+                blocks.get(SoilBlockType.SALTED_EARTH).get()
+            ), blocks.get(SoilBlockType.SALTED_EARTH).get().defaultBlockState()));
```
alcatrazEscapee: These, given they are 1-1 landslide recipes, shouldn't exist, and should use the `can_landslide` tag instead (which automatically infers copying the input state). It's just less files which is nice. Also the mud one above should be doing this too please and thanks.

src/main/java/net/dries007/tfc/world/biome/BiomeNoise.java
```diff
@@ -170,6 +170,45 @@ public static Noise2D lowlands(long seed)
...
+    /**
+     * Very flat biome
+     */
+    public static Noise2D flats(long seed)
+    {
+        return new OpenSimplex2D(seed)
+                .octaves(4)
+                .spread(0.03f)
+                .scaled(SEA_LEVEL_Y - 12, SEA_LEVEL_Y + 8)
+                .clamped(SEA_LEVEL_Y, SEA_LEVEL_Y + 2);
```
alcatrazEscapee: Formatting is inconsistent with other lines
```suggestion
        return new OpenSimplex2D(seed)
            .octaves(4)
            .spread(0.03f)
            .scaled(SEA_LEVEL_Y - 12, SEA_LEVEL_Y + 8)
            .clamped(SEA_LEVEL_Y, SEA_LEVEL_Y + 2);
```

src/main/java/net/dries007/tfc/world/surface/SoilSurfaceState.java
```diff
@@ -39,6 +39,17 @@ public static SurfaceState buildType(SoilBlockType type)
         return type == SoilBlockType.GRASS ? new SoilSurfaceState.NeedsPostProcessing(regions) : new SoilSurfaceState(regions);
     }
 
+    public static SurfaceState buildDryDirt(SoilBlockType type)
```
alcatrazEscapee, line 42: Why no transition zones?

src/main/java/net/dries007/tfc/world/surface/builder/DuneSurfaceBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+public class DuneSurfaceBuilder implements SurfaceBuilder
+{
+    public static final SurfaceBuilderFactory INSTANCE = DuneSurfaceBuilder::new;
+
+    protected DuneSurfaceBuilder(long seed) {}
+
+    @Override
+    public void buildSurface(SurfaceBuilderContext context, int startY, int endY)
+    {
+        context.setSlope(0);
```
alcatrazEscapee: Why? If the biome is exceedingly flat, this shouldn't be necessary, and this seems like it would cut off borders where the slope is intended to interpolate between them. imo it would be better practice to check the slope, and use a different surface builder vs. setting the slope whenever it's used.

src/main/java/net/dries007/tfc/world/surface/builder/DuneSurfaceBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+        SurfaceState sand = SoilSurfaceState.buildSand(false);
```
alcatrazEscapee, line 24: This should be a constant in `SurfaceStates`, it's an unnecessary copy and allocation every surface invocation (256x per chunk)

src/main/java/net/dries007/tfc/world/surface/builder/GrassyDunesSurfaceBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+{
+    public static final SurfaceBuilderFactory INSTANCE = seed -> new GrassyDunesSurfaceBuilder(seed);
+
+    private final Noise2D grassHeightVariationNoise;
+
+    public GrassyDunesSurfaceBuilder(long seed)
+    {
+        final Random random = new Random(seed);
+
+        grassHeightVariationNoise = new OpenSimplex2D(random.nextLong()).octaves(2).scaled(SEA_LEVEL_Y + 8, SEA_LEVEL_Y + 12).spread(0.3f);
```
alcatrazEscapee: `new Random(seed).nextLong()` is deterministic, might as well just use `new OpenSimple2D(seed + <some keysmash>)`, i.e. `1928379123L` (authentic, hand-written keysmash)

src/main/java/net/dries007/tfc/world/surface/builder/GrassyDunesSurfaceBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public void buildSurface(SurfaceBuilderContext context, int startY, int endY)
+    {
+        final double heightVariation = grassHeightVariationNoise.noise(context.pos().getX(), context.pos().getZ());
+
+        context.setSlope(0);
+        SurfaceState sand = SoilSurfaceState.buildSand(false);
```
alcatrazEscapee: Again, both issues with `setSlope()` and caching the surface state

src/main/java/net/dries007/tfc/world/surface/builder/GrassyDunesSurfaceBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+
+        if (startY > heightVariation)
+        {
+            SurfaceState grass = SoilSurfaceState.buildDryDirt(SoilBlockType.GRASS);
```
alcatrazEscapee, line 52: Again caching the surface state

src/main/java/net/dries007/tfc/world/surface/builder/MudFlatsSurfaceBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+{
+    public static final SurfaceBuilderFactory INSTANCE = MudFlatsSurfaceBuilder::new;
+
+
+    public MudFlatsSurfaceBuilder(long seed) {}
+
+    @Override
+    public void buildSurface(SurfaceBuilderContext context, int startY, int endY)
+    {
+        if (context.getBlockState(66).canBeReplaced())
```
alcatrazEscapee: What on _earth_ is this doing? This seems wildly unintuitive.

src/main/java/net/dries007/tfc/world/surface/builder/SaltFlatsSurfaceBuilder.java
```diff
@@ -0,0 +1,26 @@
...
+public class SaltFlatsSurfaceBuilder implements SurfaceBuilder
+{
+    public static final SurfaceBuilderFactory INSTANCE = SaltFlatsSurfaceBuilder::new;
+
+    public SaltFlatsSurfaceBuilder(long seed) {}
+
+    @Override
+    public void buildSurface(SurfaceBuilderContext context, int startY, int endY)
+    {
+        if (context.getBlockState(66).canBeReplaced())
```
alcatrazEscapee: Same here

src/main/java/net/dries007/tfc/world/surface/builder/SaltFlatsSurfaceBuilder.java
```diff
@@ -0,0 +1,26 @@
+package net.dries007.tfc.world.surface.builder;
+
+import net.dries007.tfc.world.noise.Noise2D;
+import net.dries007.tfc.world.noise.OpenSimplex2D;
+import net.dries007.tfc.world.surface.SurfaceBuilderContext;
+import net.dries007.tfc.world.surface.SurfaceStates;
+
+public class SaltFlatsSurfaceBuilder implements SurfaceBuilder
```
alcatrazEscapee: This and the previous surface builder could be merged, and take the surface states as constructor parameters, reduces some duplication.

src/test/java/net/dries007/tfc/test/TestSetup.java
```diff
@@ -194,4 +195,11 @@ class TagMap extends LinkedHashMap<ResourceLocation, TagBuilder>
         @Override
         public void clear() {} // No-op
     }
+
+    public static long seed()
+    {
+        final long seed = new Random().nextLong();
```
alcatrazEscapee:
```suggestion
        final long seed = RandomSupport.generateUniqueSeed();
```

src/test/java/net/dries007/tfc/test/TestSetup.java
```diff
@@ -194,4 +195,11 @@ class TagMap extends LinkedHashMap<ResourceLocation, TagBuilder>
         @Override
         public void clear() {} // No-op
     }
+
+    public static long seed()
```
alcatrazEscapee:
```suggestion
   default long seed()
```

src/test/java/net/dries007/tfc/test/drawing/BiomeNoiseTest.java
```diff
@@ -42,30 +41,48 @@ public class BiomeNoiseTest extends TestHelper
     @Test
     public void testRollingHills()
     {
-        terrain.draw("noise_rolling_hills", BiomeNoise.hills(seed(), -5, 28));
+        terrain.draw("noise_rolling_hills", hills(TestSetup.seed(), -5, 28));
```
alcatrazEscapee: By modifying the implementation to be `default` instead of `static`, means we don't have to quality all these calls in tests.

```suggestion
        terrain.draw("noise_rolling_hills", hills(seed(), -5, 28));
```

alcatrazEscapee (changes requested):

resources/assets.py
Therighthon: Huh, this is unintentional, but also I saw this and deleted it so I don't know why it's back, or how it got there in the first place.

src/data/java/net/dries007/tfc/data/providers/BuiltinRecipes.java
Therighthon: Neat, didn't know we had that but that's really nice

src/main/java/net/dries007/tfc/world/surface/SoilSurfaceState.java
Therighthon: No good reason, added

src/main/java/net/dries007/tfc/world/surface/builder/SaltFlatsSurfaceBuilder.java
Therighthon: No really sure what you meant by this comment because I'm not modifying the slope here, but I have switched this to check the StartY instead of checking if a block is air/replaceable

resources/constants.py
```diff
@@ -1609,14 +1609,14 @@ def lang_enum(name: str, values: Sequence[str]) -> Dict[str, str]:
...
-    'tfc.enum.foresttype.savanna_shrub_diverse': 'Savanna Shrub',
-    'tfc.enum.foresttype.secondary_bamboo': 'Secondary Bamboo',
+    "tfc.enum.foresttype.edge_bamboo": "Edge Bamboo",
+    "tfc.enum.foresttype.savanna_alternate": "Savanna Alternate",
+    "tfc.enum.foresttype.savanna_diverse": "Savanna",
+    "tfc.enum.foresttype.savanna_monoculture": "Savanna Monoculture",
+    "tfc.enum.foresttype.savanna_shrub_alternate": "Savanna Shrub Alternate",
+    "tfc.enum.foresttype.savanna_shrub_monoculture": "Savanna Shrub Monoculture",
+    "tfc.enum.foresttype.savanna_shrub_diverse": "Savanna Shrub",
+    "tfc.enum.foresttype.secondary_bamboo": "Secondary Bamboo",
```
alcatrazEscapee: another typo? (harmless, but pretty much all our other python code uses `'` instead of `"`)

src/data/java/net/dries007/tfc/data/providers/BuiltinRecipes.java
```diff
@@ -268,9 +268,6 @@ public void accept(ResourceLocation id, Recipe<?> recipe, @Nullable AdvancementH
                 blocks.get(SoilBlockType.FARMLAND).get(),
                 blocks.get(SoilBlockType.ROOTED_DIRT).get()
             ), blocks.get(SoilBlockType.DIRT).get().defaultBlockState()));
-            add(new LandslideRecipe(BlockIngredient.of(
```
alcatrazEscapee, line 271: It looks like the replacement tags for this (and cracked mud/earth) are still missing. If you need the location, in `BuiltinBlockTags`, look for `tag(CAN_LANDSLIDE)` and add the following:

```java
    .add(TFCBlocks.SOIL.get(SoilBlockType.MUD))
    .add(TFCBlocks.SOIL.get(SoilBlockType.CRACKED_EARTH))
    .add(TFCBlocks.SOIL.get(SoilBlockType.SALTED_EARTH))
```

alcatrazEscapee (changes requested):

resources/constants.py
Therighthon: Good catch, fixed another spot as well

src/data/java/net/dries007/tfc/data/providers/BuiltinRecipes.java
Therighthon: Ah yes, I missed that when moving things around. Fixed now

PR #2786, opened 2024-09-14
Migrate seasonal plant models to custom baked models

Therighthon:
- Migrated all seasonal plant models to use dynamic baked models rather than different blockstates
  - Each plant block has 6 stages, that it cycles through in sequence based on the properties of that plant
  - Each plant bases the timing of this cycle on either the warmest day of the year, or the wettest day of the year and offsets from that
  - Non-seasonal plants can now, in theory, be made seasonal with a resource pack, though they will use the default blooming schedule.
  - Added ability for plants to bloom at different times of day. Currently used for Sacred Datura, Morning Glory, and Lotus
  - Individual plants will bloom at slightly different times using a positional hash. The magnitude of this hash increases at the equator (or near rainfall variance = 0 for wet-season sensitive plants), to fuzz boundaries
  - Winter stages of plants are automatically replaced with a second round of summer stages near the equator (or near rainfall variance = 0 for wet-season sensitive plants)
- Removed the 'stage' blockstate property from all plants. Property still exists, as a few fruit trees seem to use it
- Removed 'age' blockstate property from plant BlockTypes that don't use it
- Made plant texture/model names more standardized. Texture '0' now always refers to the blooming model/texture of the plant
- Separated the 3 colors of waterlily into 3 different plants (as originally intended when I made the textures, lol)
- Added seasonality to a few plants that were missing it, including cattails and maiden pink.
- Misc. texture and model changes


src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
```diff
@@ -0,0 +1,309 @@
+package net.dries007.tfc.client.model;
```
eerussianguy, line 7: needs license header

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+        final Block block = state.getBlock();
+        if (!(block instanceof PlantBlock))
+        {
+            return getModelFromCalendar();
+        }
+        else
+        {
+            final RegistryPlant plant = ((PlantBlock) block).getPlant();
+            float start = plant.getBloomOffset();
```
eerussianguy, line 101: `final`

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
```diff
@@ -0,0 +1,309 @@
...
+            Random random = new Random();
+            Level level = ClientHelpers.getLevel();
+            final BlockPos posXZ = new BlockPos(pos.getX(), 0, pos.getZ());
```
eerussianguy, line 104: `BlockPos#setY` is a function right?

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
```diff
@@ -0,0 +1,309 @@
...
+        {
+            assert buddingBakedModel != null;
+            return buddingBakedModel;
+        }
+    }
+
+    public BakedModel getModelByDayTime(float startTime, float endTime)
+    {
+        final int dayTime = (int) Calendars.CLIENT.getCalendarDayTime();
+        if ((endTime < dayTime && dayTime < startTime) || (startTime < endTime && (dayTime < startTime || endTime < dayTime)))
```
eerussianguy, line 196: Not entirely sure, but if models change over the course of a day aren't we relying on a chunk rebuild to update the model? I guess that happens frequently enough for that to not matter.

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
```diff
@@ -0,0 +1,309 @@
...
+    {
+        return ItemOverrides.EMPTY;
+    }
+
+    record BakedModelData(BakedModel toRender)
+    {
+        public static final ModelProperty<BakedModelData> PROPERTY = new ModelProperty<>();
+    }
+
+    public static class Loader implements IGeometryLoader<PlantBlockModel>
```
eerussianguy, line 288: can this be an ENUM?

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
alcatrazEscapee: Nope - only for `MutableBlockPos`

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
Therighthon: Yeah, it does leave the potential for these plants to get slightly out of sync with their actual schedules, but from what I was seeing in testing random ticks update chunks often enough that it should be okay. Can always revert this later if we decide we don't like it.

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
Therighthon: Yep

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
Therighthon: Little confused what you mean here, as in I should re-organize this so that `start` is multiple `final` variables rather than one that gets modified? Or are you saying that `plant` shouldn't be final?

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
Therighthon: You're talking about the `record` above this snippet? Or the `Loader` class? And either way, I am not really sure, the `record` is pretty much copy-pasted from your code, and `Loader` is based on NeoForge's example code.

src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
alcatrazEscapee: Maybe a nitpick, but it might be simpler if instead of `getFractionOfDay() * 24000` we just use `getFractionOfDay() * 24` and call it `hours` in the plant definition. Since we're already deviating from vanilla daytime here, it's probably best not to confuse that with `level.dayTime()`

eerussianguy (approved):

PR #2790, opened 2024-09-21
Finite continent world generation setting

alekiponi:
- Add a worldgen setting for finite continents:
- Continents falloff with no more continents generating after 1.2x the climate scale distance on either axis
- Islands cease to generate at 1.0x the climate scale distance


eerussianguy:
I guess my question is why is this a boolean value and not two sliders.

alekiponi:
Because the expectation is that it should be one climate cycle

PR #2803, opened 2024-10-12
Instead of changing the rain alpha according to the rain level, select from 4 different amounts of rain droplets

alekiponi:
Also should affect snow


src/main/java/net/dries007/tfc/client/overworld/LevelRendererExtension.java
```diff
@@ -419,7 +431,28 @@ public boolean renderSnowAndRain(ClientLevel level, int ticks, float partialTick
...
+                                else if (rainIntensity >= 0.25f)
+                                {
+                                    RenderSystem.setShaderTexture(0, RAIN_1);
+                                }
+                                else
+                                {
+                                    RenderSystem.setShaderTexture(0, RAIN_0);
+                                }
+
+                                TerraFirmaCraft.LOGGER.debug(rainIntensity + "");
```
alcatrazEscapee: No

src/main/java/net/dries007/tfc/client/overworld/LevelRendererExtension.java
```diff
@@ -419,7 +431,28 @@ public boolean renderSnowAndRain(ClientLevel level, int ticks, float partialTick
                                 }
 
                                 stateFlag = 0;
-                                RenderSystem.setShaderTexture(0, RAIN_LOCATION);
+
+
+                                // select the texture based on the amount of rain
```
alcatrazEscapee, line 440: I'd rather these be in an array with a `RAIN_LOCATIONS[(int) (rainIntensity * 4)]` as it's one line of code vs. several

alcatrazEscapee (changes requested):

PR #2806, opened 2024-10-25
Fix familiarity decay

ashleney:
Familiarity seemed to decay at 0.02*24000 each day


alcatrazEscapee:
This is just fixing it in a way that needs to be looked at again later (see the aforementioned deprecation on `TICKS_IN_DAY`).

PR #2809, opened 2024-10-27
Add blacksmith tooltip to ForgingBonus in tfc 1.20.x

koenigstag:
Small changes to add author NBT to ForgingBonus capability in order to display blacksmith nickname in item tooltip.
Added translation to all language files.

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/AnvilBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/AnvilBlockEntity.java
@@ -281,7 +281,7 @@ public class AnvilBlockEntity extends InventoryBlockEntity<AnvilBlockEntity.Anvi
                     {
                         final float ratio = (float) forge.getSteps().total() / ForgeRule.calculateOptimalStepsToTarget(recipe.computeTarget(inventory), recipe.getRules());
                         final ForgingBonus bonus = ForgingBonus.byRatio(ratio);
-                        ForgingBonus.set(outputStack, bonus);
+                        ForgingBonus.set(outputStack, bonus, player);
 
                         if (bonus == ForgingBonus.PERFECTLY_FORGED)
                         {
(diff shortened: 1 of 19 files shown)


koenigstag:
Blacksmith tooltip in action
image

alcatrazEscapee:
First, do you speak all the languages you have added translations for? I'm assuming not, in which case you should not have added them.
Second, this was already added in fd03de3771ba099e5e35cc469c4d28845ff4b6c3 in 1.21.x, it should be ported without changes, which this notably wasn't (accepting for API differences).

koenigstag:
I can say that I've translated 4 languages correctly cause I used to learn and use them (Ru, Uk, De, Pl), other 3 have single translation possible (Es, Pt, Tr), and for other asian languages I've used google translate for single word translation, which should be also correct.

koenigstag:
Will work on porting this to 1.20.x

@alcatrazEscapee 
Upd. Sadly, Forge 1.20.1 is missing features from 1.21 such as DataComponent, Registries.DATA_COMPONENT_TYPE, and also there is no such package as net.minecraft.network.codec.

So will search for a workaround if any.
Or will try to use old-fashioned NBT system to not break compatibility with mc 1.20.1.

eerussianguy:
You would need to do it with NBT, yes

koenigstag:
Its already done via NBT author tag in this MR

`stack.getOrCreateTag().putString(KEY_AUTHOR, player.getName().getString());`

alcatrazEscapee:
Both issues I pointed out are still present:
- The implementation is different in 1.21 than it is here. To be specific, this is missing implementation for copying, has a different display than 1.21, has issues with empty string handling and is in general, not acceptable.
- No, google translated entries are **not** acceptable from someone who doesn't speak the language. Nor is "this should be correct"

koenigstag:
What should I do with Wielding Recipe, as the logic there requires a new "Behavior bonus" field ?
The rewrite of recipe will break other add-on recipes.
I think I can add a new field and not remove the old one. The logic can be combined and forgiving as for migration period.

koenigstag:
- Added logic to copy bonus in WeldingRecipe and CopyForgingBonusModifier classes
- Added logic for new "bonus behavior" field in WeldingRecipe. Bonus behavior logic has the priority over "combine bonus".

Is there anything else I should fix?

PR #2811, opened 2024-10-30
Fix random crashes caused by `getTicker`

nooooocato:
I recently started a server using the Allthemod-Gravitas2 modpack, and certain chunks in the debug world are crashing in TFC in an irregular manner. I believe this may be caused by a certain addon.

BUT...

After inspection, I found that there is only one assert for null checks in a critical area but no if-check.

Given that TFC will have more and more addons in the future, Maybe it's necessary to implement NULL handling in this method to enhance robustness.

After patching this, the game can run smoothly in the debug world.

[crash-2024-10-30_07.24.16-server.txt](https://github.com/user-attachments/files/17567387/crash-2024-10-30_07.24.16-server.txt)

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blocks/ExtendedProperties.java
+++ b/src/main/java/net/dries007/tfc/common/blocks/ExtendedProperties.java
@@ -233,9 +233,11 @@ public class ExtendedProperties
     <T extends BlockEntity> BlockEntityTicker<T> getTicker(Level level, BlockEntityType<T> givenType)
     {
         assert blockEntityType != null;
-        if (givenType == blockEntityType.get())
-        {
-            return (BlockEntityTicker<T>) (level.isClientSide() ? clientTicker : serverTicker);
+        if (blockEntityType != null){            
+            if (givenType == blockEntityType.get())
+            {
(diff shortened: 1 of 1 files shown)


alcatrazEscapee:
Without any other information this is complete nonsense hiding a bug rather than fixing it. In what scenario is this null, if it is in TFC, it should be fixed, if it is in an addon, we absolutely won't be paving over that problem to hide an addon's bug. If you've already investigated it, it should be trivial to see _what_ is crashing here.

nooooocato:
Maybe I can continue to investigate what the issue is. If I came across as rude, I apologize. However, I believe both the development and runtime environments should handle null checks here, such as printing a trace on error and allowing the game to continue running, or something else. Honestly, I don't have much experience with MC mod development. I simply think that a method that can return null returning null after receiving a null parameter is quite intuitive 🤷‍♂️. If there are any further developments, I'd be happy to look deeper into it.

I've thought about it seriously and I think you are right. However, the assert is aimed at addons or developers working around TFC. For a player, it's obviously undesirable to see error outputs from the program. But for modpack creators or more laid-back developers, it's unlikely they would check whether other mods are flawless while integrating or making interactions between other addons.

PR #2815, opened 2024-11-02
Fixed incompability with TACZ weapon mod

Hexdigest123:
[TACZ issue 142](https://github.com/MCModderAnchor/TACZ/issues/142)

Added an exception handler that allows TerraFirmaCraft to operate with TACZ, as it raises an IllegalArgumentException when used together.


alcatrazEscapee:
This is wrong on so many levels, absolutely not.

PR #2825, opened 2024-11-15
Shield Volcanos/Hot Spots

Therighthon:
Region Generator
- Adds hot spot tracks to the region generator through the point.hotSpotAge
- Hot spot tracks have random, curved trajectories that are constant across large cells
- Hot spot tracks increase in age, placing Active, Dormant, Extinct, and Ancient (Or Sunken) shield volcano biomes depending on the age
- Hot spot locations/past locations are based on a noise map that is also accessible to the biome noise
- Hot spot locations of all ages have the volcanic rock type

Biomes
Shield Volcanoes - Overview
- Surface rock is always basalt due to the surface builder
- Pumice found on the ground in all of these biomes
- Variants described below

Active
- Very tall, usually gentle-sloped mountains, rarely with calderas in the center
- Decorated with standard volcanoes
- Surface builder with cooled lava flows

Dormant
- Tall, steeper mountains with large, deep calderas in the center
- Decorated with Tuff rings and active hotsprings

Extinct
- Medium, steep-sided calderas, typically with lakes at the bottom
- Decorated with Tuff rings and dry hotsprings

Ancient
- Only present at continental hot spots
- Highly weathered calderas, often barely recognizable
- Decorated with somewhat rarer tuff rings

Sunken
- Only present at oceanic hot spots
- Volcanic rings that barely break the surface of the ocean
- Decorated with somewhat rarer tuff cones

Also, two shore biome variants that use custom surface builders, and one of which also has tuff cones so that they are not always cut off


Therighthon:
Closing this, as the ice sheet work was built on top of this, and trying to merge them separately would just create pointless conflicts. See: https://github.com/TerraFirmaCraft/TerraFirmaCraft/pull/2855

PR #2832, opened 2024-11-24
Code Optimization and Improved Readability

wallhackj:
I made the following changes:
    Optimized ComposterBlockEntity refactoring a block of code, improving performance, using more efficient methods.
    Improved code readability by restructuring functions, adding comments.
These changes should not affect the existing behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ Another small update with some small features and fixes.
 
 ### Changes
 
-- Added JEI recipe click areas to various TFC menus, which when clicked open the list of recipes in JEI for that device (#2730)
+- Added JEI recipe click areas to various TFC menus, which when clicked open the list of recipes in JEI for that
+  device (#2730)
 - Kelp in oceans generates less tall on average, and should reach the surface of the ocean less often
 
 ### Fixes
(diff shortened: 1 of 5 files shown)


alcatrazEscapee:
None of these changes are measurable or realistic optimizations. You've made backwards-incompatible changes to the serialization of the composter for no real reason, as far as I can see there are _zero_ actual optimizations here (and if you disagree with this claim, please offer proof, i.e. profiling data, of said nebulous "optimization" claims.)

The vast majority of this PR is changing indentation and brace styles, in order to be contrary with _every other file in the project_. This is not "more readable", this is inconsistent and pointless. It's also, not an optimization, nor is it more readable, especially given that we _have_ a fixed code style that is committed to this repository.

The changes to particles are utterly pointless and do not constitute an optimization either.

The changes to `Acessors` are stylistic changes, and in some cases outright bad changes, for example, removing methods of specific types, and changing them to `Object type` - thanks, you just converted an issue that would've been caught at compile time to a runtime one, in other words, terrible. It also, once again, re-indents the whole file, making it in my opinion, much worse looking, replacing references with static imports which is a stylistic choice that we, once again, don't want.

I appreciate the idea but none of this PR is substantial or useful and is borderline spam.

PR #2837, opened 2024-11-28
Fix/broken anvil recipe

robbert229:
what am I doing wrong here?


robbert229:
RIP, I meant to do this to my private fork, not here. Sorry

PR #2840, opened 2024-12-05
Using the pick item key on dead torches now returns the lit torch item

alekiponi:


alcatrazEscapee (changes requested):
This can be set via `cloneItem()` in the properties instantiation for the block, and doesn't require a method override in either class.

eerussianguy:
class still needs to implement the `IForgeBlockExtension`

PR #2855, opened 2025-01-06
Ice Caps and Shield Volcanoes (Combined)

Therighthon:
Worth noting here that the changes to the surface builder do make some deep deserts look pretty bad, because the default very dry surface in most biomes is gravel, since this works better in most biomes for the warm->cold transition up mountains. Fortunately, I was planning on adding new biomes to deserts anyways, so that will be my next objective after this, and those biomes will all use surface builders tuned to dry climates

ICE SHEETS:

Region Generator
    Changed oceanic influence on temperature to be more variable. As a result, some coastal areas are unfrozen all the way to the poles in summer, similar to past behavior, while others are nearly as cold as inland areas.
    Ice Sheet and Periglacial biomes are placed in cold climates
    
Biomes
    There are a lot of these, but in general they fall into the following categories: Ice Sheet, Edge, Glaciated, and Periglacial
    Ice Sheet biomes are completely covered in ice, though they may have features sticking out of the ice such as mountains/tuyas
    Edge biomes are used as borders between Ice Sheet biomes and other biomes where a transition between surface builders is required
    Glaciated biomes are biomes that have smaller glaciers at high elevations
    Periglacial biomes is here used as an umbrella term for any other biomes exclusive to cold regions
    When adding new biomes to far north regions in future, we need to ensure that they do not create messy borders with ice sheet biomes

Surface Builder
    Reworked the standard climate-sensitive surface states:
     - Instead of transitioning from sand->grass as rainfall increases, the scale now goes from snow->gravel (or sand in some biomes)->coarse dirt->grass
     - This transition is not controlled by temperature as well as aridity
     - Aridity can only transition from grass down to gravel, snow only occurs in extreme cold.
     - As far as how this compares with the previous surface builder, areas that were previously grass between about 75 and 150 rainfall are now dominated by coarse dirt. Areas near the poles, or high on mountains will not have grass, even if they are wet enough
    
    Removed the mountain surface builder, as the above changes made it semi-redundant
    Added Ice Sheet Surface builders used in the various ice sheet biomes

Misc
    Re-added sea ice, now done via surface builder like the icebergs
    Fragmented sea ice starts in temperatures slightly colder than where icebergs appear, and becomes a solid layer of ice in even colder areas
    Reworked powder snow slightly, now only replaces snow/ice blocks
    Added ocean biomes and coastal ice sheet biomes to list of where penguins can spawn
    Added ocean biomes to list of where polar bears can spawn
    Changed how spring features work:
     - Split lava/water springs into surface and underground features. Underground springs are more common, and occur everywhere.
     - Surface lava springs only occur in volcanic biomes, but are much more common
     - Surface water springs only occur in high enough rainfall (>75 or so), and extra springs occur in very wet climates
     - Separate, more common water springs added that only generate in packed/blue ice (so, essentially limited to mountains and glaciers)
     Added a blacklist of blocks creeping plants can't grow on. For now, just snow and ice
     Probably other stuff, which I may add to this list as I review the code

SHIELD VOLCANOES:

Region Generator

    Adds hot spot tracks to the region generator through the point.hotSpotAge
    Hot spot tracks have random, curved trajectories that are constant across large cells
    Hot spot tracks increase in age, placing Active, Dormant, Extinct, and Ancient (Or Sunken) shield volcano biomes depending on the age
    Hot spot locations/past locations are based on a noise map that is also accessible to the biome noise
    Hot spot locations of all ages have the volcanic rock type

Biomes
Shield Volcanoes - Overview

    Surface rock is always basalt due to the surface builder
    Pumice found on the ground in all of these biomes
    Variants described below

Active

    Very tall, usually gentle-sloped mountains, rarely with calderas in the center
    Decorated with standard volcanoes
    Surface builder with cooled lava flows

Dormant

    Tall, steeper mountains with large, deep calderas in the center
    Decorated with Tuff rings and active hotsprings

Extinct

    Medium, steep-sided calderas, typically with lakes at the bottom
    Decorated with Tuff rings and dry hotsprings

Ancient

    Only present at continental hot spots
    Highly weathered calderas, often barely recognizable
    Decorated with somewhat rarer tuff rings

Sunken

    Only present at oceanic hot spots
    Volcanic rings that barely break the surface of the ocean
    Decorated with somewhat rarer tuff cones

Also, two shore biome variants that use custom surface builders, and one of which also has tuff cones so that they are not always cut off


src/main/java/net/dries007/tfc/world/surface/builder/StoneCirclesSurfaceBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+{
+    public static final SurfaceBuilderFactory INSTANCE = StoneCirclesSurfaceBuilder::new;
+
+    public StoneCirclesSurfaceBuilder(long seed) {}
+
+    @Override
+    public void buildSurface(SurfaceBuilderContext context, int startY, int endY)
+    {
+        final NormalSurfaceBuilder surfaceBuilder = NormalSurfaceBuilder.ROCKY;
+        final Noise2D edges = BiomeNoise.stoneCircles(context.getSeed());
```
alcatrazEscapee: This should be a field populated in the constructor

src/main/java/net/dries007/tfc/world/surface/builder/PatternedGroundSurfaceBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+{
+    public static final SurfaceBuilderFactory INSTANCE = PatternedGroundSurfaceBuilder::new;
+
+    public PatternedGroundSurfaceBuilder(long seed) {}
+
+    @Override
+    public void buildSurface(SurfaceBuilderContext context, int startY, int endY)
+    {
+        final NormalSurfaceBuilder surfaceBuilder = NormalSurfaceBuilder.INSTANCE;
+        final Noise2D edges = BiomeNoise.patternedGround(context.getSeed());
```
alcatrazEscapee: This should be a field set in the constructor

src/main/java/net/dries007/tfc/world/surface/builder/OceanSurfaceBuilder.java
```diff
@@ -126,5 +129,37 @@ private void frozenOceanExtension(SurfaceBuilderContext context, int startY, int
...
+            {
+                placeIce = true;
+            }
+            else if (maxAnnualTemperature > iceStart)
+            {
+                placeIce = false;
+            }
+            else
+            {
+                final double patternedNoise = BiomeNoise.seaIceNoise(context.getSeed()).noise(x, z);
```
alcatrazEscapee: This should be a field set in the constructor

src/main/java/net/dries007/tfc/world/surface/builder/IceSheetSurfaceBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+            iceDepth = 35;
+        }
+
+        if (hasStonyPeaks && startY > glacierSurfaceHeight + 2.5)
+        {
+            NormalSurfaceBuilder.ROCKY.apply(seed).buildSurface(context, startY, endY);
+        }
+        else if (startY < glacierBaseHeight - 1.5)
+        {
+            NormalSurfaceBuilder.INSTANCE.apply(seed).buildSurface(context, startY, endY);
```
alcatrazEscapee: Because `NormalSurfaceBuilder` is invariant, you should be able to remove the `apply(seed)` here and just call `NormalSurfaceBuilder.INSTANCE.buildSurface` and `NormalSurfaceBuilder.ROCKY.buildSurface`

src/main/java/net/dries007/tfc/world/surface/builder/IceSheetSurfaceBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+    private final Noise2D iceSurfaceNoise;
+    private final Noise2D baseNoise;
+    private final boolean hasMoraines;
+    private final boolean hasStonyPeaks;
+
+    IceSheetSurfaceBuilder(long seed, Noise2D baseNoise, Noise2D iceSurfaceNoise, boolean hasMoraines, boolean hasStonyPeaks)
+    {
+        this.baseNoise = baseNoise;
+        this.iceSurfaceNoise = iceSurfaceNoise;
+        this.seed = seed;
```
alcatrazEscapee: I believe you can remove the parameter and field `seed` here, as it is not needed

src/main/java/net/dries007/tfc/world/surface/builder/ShieldVolcanoSurfaceBuilder.java
```diff
@@ -0,0 +1,194 @@
...
+    {
+        this.hasLavaFlows = hasLavaFlows;
+        this.sandy = sandy;
+        this.seed = seed;
+    }
+
+    @Override
+    public void buildSurface(SurfaceBuilderContext context, int startY, int endY)
+    {
+        final Noise2D noise = new OpenSimplex2D(seed).octaves(2).spread(0.25);
```
alcatrazEscapee: This should be a field set in the constructor. Then it looks like you can also remove the field `seed`.

src/main/java/net/dries007/tfc/world/surface/builder/IceSheetShieldVolcanoSurfaceBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+            iceDepth = Math.max((int) ((startY - moraineCrestHeight) * 2), 0);
+        }
+        else
+        {
+            iceDepth = 35;
+        }
+
+        if (startY < minFreezingHeight || (hasStonyPeaks && startY > glacierSurfaceHeight + 2.5) || (startY < glacierBaseHeight - 1.5))
+        {
+            ShieldVolcanoSurfaceBuilder.DORMANT.apply(seed).buildSurface(context, startY, endY);
```
alcatrazEscapee: `ShieldVolcanoSurfaceBuilder.DORMANT.apply(seed)` should be a field set in the constructor, which also means you can probably remove the `seed` field.

src/main/java/net/dries007/tfc/world/region/ChooseBiomes.java
```diff
@@ -43,13 +53,79 @@ public void apply(RegionGenerator.Context context)
...
+                    final float maxIceSheetTemp = -16f + 0.006f * point.rainfall;
+                    if (temp < maxIceSheetTemp + 2)
+                    {
+                        point.biome = ICE_SHEET_OCEANIC_MOUNTAINS;
+                    }
+                    else if (temp < maxIceSheetTemp + 6) //
+                    {
+                        point.biome = GLACIATED_OCEANIC_MOUNTAINS;
+                    }
+                    else if (temp < maxIceSheetTemp + 10) //
```
alcatrazEscapee: 2x empty comment stubs?

src/main/java/net/dries007/tfc/world/region/AddHotspots.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    // Shifts the location by 0.5 towards the origin.
+    // Use this when sampling noise from regional coordinates to sample the center of the region point
+    public double shift(int point)
+    {
+        if (point == 0) return 0;
+        return point - (0.5 * Math.signum(point));
+    }
+}
```
alcatrazEscapee: Are you sure this is correct? It would think this leaves to biased sampling around the x/z axis, where you'd have more sample points (0.5, 0. -0.5) as opposed to away from them. All our other world generation is not symmetric over the x/z axis like this

src/main/java/net/dries007/tfc/world/region/AddHotspots.java
```diff
@@ -0,0 +1,98 @@
...
+                : lastAge == 3 ? BiomeNoise.extinctHotSpots(seed) : BiomeNoise.ancientHotSpots(seed)).spread(128);
+
+            for (int dx = -1; dx <= 1; dx++)
+            {
+                for (int dz = -1; dz <= 1; dz++)
+                {
+                    final Region.Point next = region.atOffset(index, dx, dz);
+                    if (next != null)
+                    {
+                        if (next.hotSpotAge == 0)
```
alcatrazEscapee:
```suggestion
                    if (next != null && next.hotSpotAge == 0)
```

src/main/java/net/dries007/tfc/world/region/AddHotspots.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public void apply(RegionGenerator.Context context)
+    {
+        final Region region = context.region;
+        final long seed = context.generator().levelSeed();
+        final double threshold = 0.65;
+        final double expansionThreshold = 0.15;
+
+        final Noise2D hotspotAge = BiomeNoise.hotSpotAge(seed).spread(128);
```
alcatrazEscapee: All the calls to `BiomeNoise` should not be done in this class; they should be done in `RegionGenerator` (see how continents does it). If it's simpler, it might be worth creating convenience methods or a record for just the hotspot-related information (i.e. `context.generator().hotSpotGenerator.plateRegions()`, or `context.generator().getHotSpotIntensityNoiseFor(int lastAge)` sort of thing.

In both cases we're avoiding creating noise functions at the point level, and keeping the generators separate. Likewise, accessing the `seed` here is a code smell - that should never be done during generation, most everything that relies on that should be separately created.

src/main/java/net/dries007/tfc/world/placement/TuyaPlacement.java
```diff
@@ -0,0 +1,100 @@
...
+import net.minecraft.world.level.levelgen.placement.PlacementContext;
+import net.minecraft.world.level.levelgen.placement.PlacementModifier;
+import net.minecraft.world.level.levelgen.placement.PlacementModifierType;
+
+import net.dries007.tfc.world.Codecs;
+import net.dries007.tfc.world.biome.BiomeExtension;
+import net.dries007.tfc.world.biome.TFCBiomes;
+import net.dries007.tfc.world.biome.TuyaNoise;
+
+public class TuyaPlacement extends PlacementModifier
```
alcatrazEscapee: There's a couple issues here with `TuyaPlacement`, `TuffRingPlacement`, and also with `VolcanoPlacement` (which is clear these were copied from)

1. The `ensureCanWrite` method in `VolcanoPlacement` is badly named, and was a bandaid fix over missing an actual issue. It should never be checking the center chunk position.
2. Between these three classes, there's a lot of duplication. By adding a small interface to `VolcanoNoise` (and then, also to `TuyaNoise` and `TuffRingNoise`), we can abstract most of this away to a central placement modifier that does distance or center placement based on said noise.

I've fixed both issues in the main branch, can you incorporate said changes into here? That should reduce the amount of duplicated code here.

src/main/java/net/dries007/tfc/world/placement/TuffRingPlacement.java
```diff
@@ -0,0 +1,100 @@
...
+import net.minecraft.world.level.levelgen.placement.PlacementContext;
+import net.minecraft.world.level.levelgen.placement.PlacementModifier;
+import net.minecraft.world.level.levelgen.placement.PlacementModifierType;
+
+import net.dries007.tfc.world.Codecs;
+import net.dries007.tfc.world.biome.BiomeExtension;
+import net.dries007.tfc.world.biome.TFCBiomes;
+import net.dries007.tfc.world.biome.TuffRingNoise;
+
+public class TuffRingPlacement extends PlacementModifier
```
alcatrazEscapee: See note on `TuyaPlacement` for simplifications that can be done here.

src/main/java/net/dries007/tfc/world/layer/TFCLayers.java
```diff
@@ -219,12 +257,49 @@ public static int shoreFor(int value)
...
-        return !isOcean(value) && value != BADLANDS;
+        return (!isOcean(value) && value != BADLANDS && value != ACTIVE_SHIELD_VOLCANO && value != DORMANT_SHIELD_VOLCANO
+            && value != EXTINCT_SHIELD_VOLCANO && value != ANCIENT_SHIELD_VOLCANO && value != ICE_SHEET_MOUNTAINS
+            && value != ICE_SHEET_MOUNTAINS_EDGE && value != ICE_SHEET_OCEANIC_MOUNTAINS && value != ICE_SHEET_OCEANIC_MOUNTAINS_EDGE
+            && value != ICE_SHEET_SHIELD_VOLCANO && value != ICE_SHEET_SHORE && value != GLACIATED_SHIELD_VOLCANO
+            && value != GLACIATED_MOUNTAINS && value != GLACIATED_OCEANIC_MOUNTAINS && value != GLACIALLY_CARVED_MOUNTAINS
+            && value != GLACIALLY_CARVED_OCEANIC_MOUNTAINS);
+    }
+
+    // TODO: Use or lose
```
alcatrazEscapee: Explanation or use/lose?

src/main/java/net/dries007/tfc/world/noise/Cellular2D.java
```diff
@@ -65,17 +70,17 @@ public Cell cell(double x, double y)
         int xPrimed = (xr - 1) * primeX;
         int yPrimedBase = (yr - 1) * primeY;
 
-        for (int xi = xr - 1; xi <= xr + 1; xi++)
+        for (int xi = xr - 2; xi <= xr + 2; xi++)
```
alcatrazEscapee: This distance should ideally be added as a parameter along with `jitter`. The base jitter value used in `FastNoiseLite` is supposed to be calibrated so that it's only necessary to sample within +/-1, obviously changing that jitter changes this distance.

src/main/java/net/dries007/tfc/mixin/IcebergFeatureMixin.java
```diff
@@ -28,13 +28,23 @@ private Block replaceWaterWithTFCWaterInCarve(Operation<Block> original)
         return TFCIcebergFeature.is(this) ? TFCBlocks.SALT_WATER.get() : original.call();
     }
 
+    // TODO: Icebergs do not generate below the sea ice/water surface
     @WrapOperation(
-        method = "carve",
+        method = "setIcebergBlock",
```
alcatrazEscapee, line 33: I don't know why these methods were duplicated, however, what we can do here is easier than keeping them duplicated.  Replace `method = "carve"` with `method = {"carve", "setIcebergBlock"}` in the original mixin.

Is the `todo` still accurate with the fixed `setIcebergBlock` mixin?

alcatrazEscapee (commented):
Work in progress review, trying to catch any mismanaged state / potential misses. About halfway through per my viewed files, GH lagging on the review making it a bit difficult.

Some of the notes about a unused `seed` parameter should be more obvious that we can do it the way I had suggested with the latest commit.

src/main/java/net/dries007/tfc/world/surface/builder/ShieldVolcanoSurfaceBuilder.java
Therighthon: The lavaFlow noise also needs the seed, so I kept the seed field, but kicked this noise to a field as well

src/main/java/net/dries007/tfc/world/layer/TFCLayers.java
Therighthon: Just needed to remember to delete that code if it ended up not being used... Obviously didn't work, lol

src/main/java/net/dries007/tfc/mixin/IcebergFeatureMixin.java
Therighthon: Todo is still accurate for the iceberg features, yes. not sure what's causing, but I checked and it predates my branch

src/main/java/net/dries007/tfc/world/region/AddHotspots.java
Therighthon: Revised to add 0.5 everywhere instead, I think I was just over-thinking this

PR #2857, opened 2025-01-10
French localization

Alguewator:
Added French translation of the main language file (fr_ca.json and fr_fr.json). Also fixed some typos included on English file.


alcatrazEscapee:
1.21 is not released, and it is currently in-development, translations at this point are not useful. Please re-target this for 1.20 if you wish for it actually to be used in existing releases. Or wait for when 1.21 is released.

Alguewator:
Fair enough. I'll keep the 1.21 lang file updated so that it can be added when this version is released.
I'll start working on the 1.20 version and on the field guide of both versions.

PR #2874, opened 2025-02-13
Check getLightPos() in IGlow#tryRemoveLight() rather than Entity#blockPosition()

Floweynt:
Using `blockPosition` may lead to `tryRemoveLight` loading chunks, which is not good.


alcatrazEscapee:
Implemented in 1.20 already by 2ce86e0 which should be cherry picked.

PR #2879, opened 2025-02-23
1.20.x updata ja_jp.json

3Min3:
We have translated the parts that were not yet translated.
We have also corrected some mistakes and made the item names more ideal.


3Min3:
I found a mistake, so I'm closing it.

PR #2880, opened 2025-02-24
[1.21.x] Glass Operation Registry

Notenoughmail:
Modifies glass operations to be a proper registry


src/main/java/net/dries007/tfc/compat/jei/category/GlassworkingRecipeCategory.java
```diff
@@ -71,23 +60,8 @@ public void setRecipe(IRecipeLayoutBuilder builder, GlassworkingRecipe recipe, I
...
-            else if (operation != GlassOperation.TABLE_POUR && operation != GlassOperation.BASIN_POUR)
-            {
-                for (Map.Entry<Item, GlassOperation> entry : GlassOperation.POWDERS.get().entrySet())
-                {
-                    if (entry.getValue() == operation)
-                    {
-                        slot.addItemStack(entry.getKey().getDefaultInstance());
-                    }
-                }
+            for (Holder<Item> item : operation.getItems()) {
```
eerussianguy: braces on newlines

src/main/java/net/dries007/tfc/common/component/glass/GlassOperation.java
```diff
@@ -6,89 +6,89 @@
...
+    public static final DeferredRegister<GlassOperation> OPERATIONS = DeferredRegister.create(KEY, TerraFirmaCraft.MOD_ID);
+
+    public static final Supplier<Map<Item, GlassOperation>> POWDERS = Suppliers.memoize(() -> {
+        final Map<Item, GlassOperation> map = new IdentityHashMap<>();
+        REGISTRY.stream().filter(op -> op.powder).forEach(op -> {
+            for (Holder<Item> item : op.items)
+            {
+                map.put(item.value(), op);
+            }
+        });
```
eerussianguy, line 55: I wonder if this could be expressed more concisely, but I don't really mind.

src/main/java/net/dries007/tfc/compat/jei/category/GlassworkingRecipeCategory.java
```diff
@@ -71,23 +60,8 @@ public void setRecipe(IRecipeLayoutBuilder builder, GlassworkingRecipe recipe, I
...
-            }
-            else if (operation != GlassOperation.TABLE_POUR && operation != GlassOperation.BASIN_POUR)
-            {
-                for (Map.Entry<Item, GlassOperation> entry : GlassOperation.POWDERS.get().entrySet())
-                {
-                    if (entry.getValue() == operation)
-                    {
-                        slot.addItemStack(entry.getKey().getDefaultInstance());
-                    }
-                }
```
eerussianguy: So, this logic performed a few different useful JEI functions but appears to have been removed. Confirm that the previous functionality is preserved. I haven't looked deeply into it so that may be the case already.

eerussianguy (commented):
This is a well-put together PR. Requesting review from @alcatrazEscapee but I think this is pretty good to merge.

src/main/java/net/dries007/tfc/compat/jei/category/GlassworkingRecipeCategory.java
Notenoughmail: The additional blowpipe persists in that the blow, stretch, table pour, and basin pour operations now explicitly include both blowpipes in their item set, which is used to populate JEI

Similarly for powders, of the two (copper & iron) that are associated with more than a single item, they all should be in the operation's item set. If I missed any items for them or missed a powder that should have multiple items let me know and I'll fix it

Notenoughmail:
ah... I've just noticed the texture map for bowl powders can't be added to, which kind of sucks if you're adding a powder operation

Should I address that in this PR, or a separate one? Any preference for how that should be done (I'm thinking just a static method in `BowlBlockEntityRenderer`)?

PR #2881, opened 2025-02-25
Fix some item names having a leading space

AnicJov:
This small patch fixes some items having an extra leading space.
For example " Copper Plated Stairs" becomes "Copper Plated Stairs" as shown in the attached screenshot:

patch_diff


eerussianguy:
I think I would just rather not send lang keys that are malformed to be written at all, as a matter of principle, but I also recognize that that doesn't really matter a ton at the end of the day.

AnicJov:
I suppose so, but I think having the `strip()` there as a failsafe is a good idea either way. I don't believe there is a case where you would want whitespace on either side of the item name.

PR #2882, opened 2025-03-07
2 Snow Tweaks

Therighthon:
- Fixed rain sounds/particles playing during heavy snow
- Added a multiplier to snow melting over large timescales. Comes into effect when time skip >4 days. Basically, this fixes the catchup algorithm not always deleting all of the snow when you load a chunk in July that hasn't been loaded since January. I found this behaved a little better (closer to parity w/ skipping 1 day at a time) than scaling it when temperatures were well above 0

### DIFF
--- a/src/main/java/net/dries007/tfc/client/overworld/LevelRendererExtension.java
+++ b/src/main/java/net/dries007/tfc/client/overworld/LevelRendererExtension.java
@@ -21,6 +21,11 @@ import com.mojang.blaze3d.vertex.Tesselator;
 import com.mojang.blaze3d.vertex.VertexBuffer;
 import com.mojang.blaze3d.vertex.VertexFormat;
 import com.mojang.math.Axis;
+import net.dries007.tfc.common.TFCTags;
+import net.dries007.tfc.common.blocks.MoltenBlock;
+import net.dries007.tfc.common.blocks.TFCBlocks;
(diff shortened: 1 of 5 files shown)


src/main/java/net/dries007/tfc/client/overworld/LevelRendererExtension.java
```diff
@@ -555,58 +556,61 @@ public boolean tickRain(ClientLevel level, int ticks, Camera camera)
...
+                            ? Shapes.block()
+                            : state.getCollisionShape(level, cursor);
+                        final double offsetY = Math.max(
+                            shape.max(Direction.Axis.Y, offsetX, offsetZ),
+                            fluid.getHeight(level, cursor)
+                        );
+
+                        final ParticleOptions options = !fluid.is(FluidTags.LAVA)
+                            && !state.is(Blocks.MAGMA_BLOCK)
+                            && !CampfireBlock.isLitCampfire(state)
```
eerussianguy: we have our own magma blocks, this is missing them already. need something more sturdy.

Therighthon: Added a tag for blocks that smoke in rain, and added magma blocks to it. Also made our firepits and forges smoke in rain.

PR #2889, opened 2025-03-21
Changed fire start behavior, added flint and pyrite

Therighthon:
Flint and Pyrite, a mid-game firestarter w/ 128 durability that lights like flint and steel but with only a 1/3 chance (configurable) per click to start a strong fire. Lights torches 100% of the time. Uses durability regardless of whether a fire is started, so ~40 uses are expected

Flint and Steel (and Flint and Pyrite and fire charges and flaming arrows) can be used to light fire pits from items on ground. All except Flint and Pyrite have a 100% chance to light regardless of kindling

Made flint and steel and fire charges play their respective sounds when used


src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
```diff
@@ -32,12 +49,32 @@
  */
 public final class StartFireEvent extends Event implements ICancellableEvent
 {
+    // Replicate vanilla fire starting sounds that do not fire
+    public static boolean startFireWithSound(Level level, BlockPos pos, BlockState state, Direction direction, @Nullable Player player, ItemStack stack)
```
Therighthon, line 53: Feels a bit hacky to me, but I could not figure out why the vanilla sounds were not playing

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
```diff
@@ -49,6 +86,70 @@ public static boolean startFire(Level level, BlockPos pos, BlockState state, Dir
 
         if (!cancelled && event.isStrong())
         {
+            final BlockPos abovePos = pos.above();
+            // Check conditions for creating a firepit if a valid firestarter
+            if (FirepitBlock.canSurvive(level, abovePos) && firepitBaseChance != -1)
+            {
+                final List<ItemEntity> items = level.getEntitiesOfClass(ItemEntity.class, new AABB(abovePos.getX() - 0.5, abovePos.getY(), abovePos.getZ() - 0.5, abovePos.getX() + 1.5, abovePos.getY() + 1, abovePos.getZ() + 1.5));
+                final List<ItemEntity> usableItems = new ArrayList<>();
```
eerussianguy, line 94: Okay so a getEntities calculation is expensive and I would rather perform this later rather than sooner. It's cheaper to check the blockstate/entity than to get entities from the global list and filter them by location and class. Let's also extrac this function to be separate for cleanliness

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
```diff
@@ -32,12 +49,32 @@
...
+
     public static boolean startFire(Level level, BlockPos pos, BlockState state, Direction direction, @Nullable Player player, ItemStack stack)
     {
-        return startFire(level, pos, state, direction, player, stack, FireStrength.STRONG);
+        return startFire(level, pos, state, direction, player, stack, FireStrength.STRONG, 1.0);
     }
 
-    public static boolean startFire(Level level, BlockPos pos, BlockState state, Direction direction, @Nullable Player player, ItemStack stack, FireStrength strength)
+    // Pass in a firepitBaseChance of -1 to disable firepit creation for a given firestarter
+    public static boolean startFire(Level level, BlockPos pos, BlockState state, Direction direction, @Nullable Player player, ItemStack stack, FireStrength strength, double firepitBaseChance)
```
eerussianguy, line 77: I am not sure if I love firepit chance getting its own parameter for the firestarting event... but I guess I see the use case.

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
```diff
@@ -49,6 +86,70 @@ public static boolean startFire(Level level, BlockPos pos, BlockState state, Dir
...
+                            firepitState = TFCBlocks.FIREPIT.get().defaultBlockState().setValue(FirepitBlock.AXIS, Direction.Axis.X);
+                        }
+                        level.setBlock(abovePos, firepitState, 3);
+                        level.getBlockEntity(abovePos, TFCBlockEntities.FIREPIT.get()).ifPresent(firepit -> {
+                            firepit.getInventory().setStackInSlot(AbstractFirepitBlockEntity.SLOT_FUEL_CONSUME, initialLog);
+                            firepit.light(firepitState);
+                        });
+                        if (player instanceof ServerPlayer serverPlayer)
+                        {
+                            TFCAdvancements.FIREPIT_CREATED.trigger(serverPlayer, firepitState);
```
eerussianguy, line 146: test that this still is being granted! I am sure it is but good to check

eerussianguy (changes requested):
Let's also add a recipe to the field guide.

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
Therighthon: Advancements in general seem to be broken, or at least when I open the advancements menu there is nothing there, and also whenever I give myself blue steel tools it isn't blasting sound effects at me like it used to.

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
Therighthon: How does one check the blockstate/entity? This is all just code I moved from the firestarter item class, so it is what we were using before.

As far as extracting the function, that proved trickier than I thought as well because we either return true if the firepit is lit successfully, false if it could be lit but failed, or continue if 3 sticks and a log are not found. of course there are ways to get around that, but the ones that spring to mind are kind of ugly

eerussianguy:
Branch has conflicts

PR #2892, opened 2025-03-22
Fix river groundwater influence applying for underground rivers

Therighthon:
Works by modifying the groundwater map based on the final surface heights at corners, so also makes rivers in canyons not moisten the surrounding plateau as much. Not sure if going and modifying the chunkdata this late in the game is kosher, but from my testing it works ok.

Probably best if @alcatrazEscapee takes a quick peep at this.

### DIFF
--- a/src/main/java/net/dries007/tfc/world/chunkdata/ChunkData.java
+++ b/src/main/java/net/dries007/tfc/world/chunkdata/ChunkData.java
@@ -224,21 +224,19 @@ public sealed class ChunkData
         this.status = Status.FULL;
     }
 
-    public void modifyBaseGroundwater(int[] surfaceHeight, ChunkAccess chunk)
+    public void modifyBaseGroundwater(int[] surfaceHeight)
     {
         assert this.baseGroundwaterLayer != null;
-        float groundwater00 = modifyBaseGroundwaterPoint(surfaceHeight[0], this.baseGroundwaterLayer.value00());
-        float groundwater10 = modifyBaseGroundwaterPoint(surfaceHeight[15], this.baseGroundwaterLayer.value10());
-        float groundwater01 = modifyBaseGroundwaterPoint(surfaceHeight[240], this.baseGroundwaterLayer.value01());
-        float groundwater11 = modifyBaseGroundwaterPoint(surfaceHeight[255], this.baseGroundwaterLayer.value11());
+        float groundwater00 = modifyBaseGroundwaterPoint(surfaceHeight[0], this.baseGroundwaterLayer.value00()); // Constant = x + 16z, x=0, z=0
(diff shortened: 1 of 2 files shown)


src/main/java/net/dries007/tfc/world/chunkdata/ChunkData.java
```diff
@@ -222,6 +224,23 @@ public void generateFull(int[] surfaceHeight, int[] aquiferSurfaceHeight)
...
+    public void modifyBaseGroundwater(int[] surfaceHeight, ChunkAccess chunk)
+    {
+        assert this.baseGroundwaterLayer != null;
+        float groundwater00 = modifyBaseGroundwaterPoint(surfaceHeight[0], this.baseGroundwaterLayer.value00());
+        float groundwater10 = modifyBaseGroundwaterPoint(surfaceHeight[15], this.baseGroundwaterLayer.value10());
+        float groundwater01 = modifyBaseGroundwaterPoint(surfaceHeight[240], this.baseGroundwaterLayer.value01());
+        float groundwater11 = modifyBaseGroundwaterPoint(surfaceHeight[255], this.baseGroundwaterLayer.value11());
+        this.baseGroundwaterLayer = new LerpFloatLayer(groundwater00, groundwater01, groundwater10, groundwater11);
+
+        chunk.setUnsaved(true);
```
alcatrazEscapee: This call to `setUnsaved()` is unnecessary as it's happening during generation

src/main/java/net/dries007/tfc/world/chunkdata/ChunkData.java
```diff
@@ -222,6 +224,23 @@ public void generateFull(int[] surfaceHeight, int[] aquiferSurfaceHeight)
         this.status = Status.FULL;
     }
 
+    public void modifyBaseGroundwater(int[] surfaceHeight, ChunkAccess chunk)
+    {
+        assert this.baseGroundwaterLayer != null;
+        float groundwater00 = modifyBaseGroundwaterPoint(surfaceHeight[0], this.baseGroundwaterLayer.value00());
```
alcatrazEscapee: would rather these constants get written out as the source x/y pair, i.e. `0 + 16 * 0`, `0 + 16 * 15`, `15 + 16 * 16`, etc. Alternatively use an `index(int x, int z) { return x + 16 * z; }` function if you don't want the IDE to complain about computed constants.

alcatrazEscapee (commented):
This is the right place to do that modification, yeah

src/main/java/net/dries007/tfc/world/chunkdata/ChunkData.java
Therighthon: I've just added comments because I just feel bad making the game do more math than it needs to

alcatrazEscapee: fwiw if you write `15 + 16 * 16` or similar, it is inlined at compile time. There's no more math done at runtime.

PR #2893, opened 2025-03-22
Crop hydration from rainfall rework

seelderr:
Reworks crop hydration from rainfall to instead be derived from the true rainfall, rather than a flat bonus based off of the average rainfall in the area.

I tried to scale the values to reflect the "MAX_RAINFALL" value being in millimeters by calculating the average rainfall intensity and converting it to match up with the 500 millimeter maximum (as raw weather intensity = 500mm rainfall yearly).

For farmland, we store an actual hydration value that increases when it is raining, and decreases when it is not. For berry bushes, we just get an estimate of the average rainfall from calendar data rather than true weather since we don't store any hydration data.

Still very much WIP, definitely needs a ton of number tweaking and balancing + code cleanup (the numbers are pretty far off right now), but you can see the overall concept in this PR.


src/main/java/net/dries007/tfc/common/blocks/soil/FarmlandBlock.java
```diff
@@ -90,19 +101,52 @@ public static Component getTemperatureTooltip(Level level, BlockPos pos, Climate
...
+        }
+
+        final WorldTracker tracker = WorldTracker.get(level);
+        final ClimateModel model = tracker.getClimateModel();
+
+        final int waterCost = findMinCostWater(level, pos); // Nearby water contributes an additional 0 - 80% hydration based on proximity
+        final int waterBoost = 20 * (5 - waterCost); // Nearby water contributes an additional 0 - 80% hydration based on proximity
+        final int rainfallBoost = getRainfallBoost(accumulatedRainfall); // Up to 30% bonus from rainfall
+        // TODO :: Maybe this humidity factor should have a lower impact when temperature is lower?
+        final int humidityBoost = (int) (30 * Mth.clampedMap(model.getRainfall(level, pos), ClimateModel.MIN_RAINFALL, ClimateModel.MAX_RAINFALL, 0, 1)); // Up to 30% bonus from humidity (average rainfall)
```
Therighthon: Open ended question here is whether we want to use rainfall or groundwater for these functions. To clarify, in 1.20 and before, we just had rainfall, but because we wanted rivers to have plants/trees when they pass through deserts, this led to rainstorms that existed only in the middle of rivers. So, we split into baseGroundwater, which is modified by rivers (and should probably eventually be modified by lakes, swamps for consistency), and rainfall, which are added together to get groundwater, which is used for most world generation.

src/main/java/net/dries007/tfc/common/blockentities/FarmlandBlockEntity.java
```diff
@@ -35,31 +50,86 @@ public FarmlandBlockEntity(BlockPos pos, BlockState state)
...
+    {
+        farmland.checkForCalendarUpdate();
+    }
+
+    public void rainTick()
+    {
+        assert level != null;
+        final long firstTick = this.getLastRainTick(), thisTick = Calendars.SERVER.getTicks();
+        long tick = firstTick + UPDATE_INTERVAL, lastTick = firstTick;
+        for (; tick < thisTick; tick += UPDATE_INTERVAL)
```
Therighthon: So, unless I am tired and misreading things, doesn't look like you got anywhere with the rain "bands" notion? Anyways, it does seem to me like there should be some sort of limit here on how long this for loop will run, as right now it looks like if you leave for 6 months and come back, every single crop block is going to be going nuts checking the weather for some 500 different calendar times.

Couple different ways we could go with this, but my main thoughts:
1. Hard cutoff time. Recent rains are going to effect the current hydration level more, so at some point a rain event is far enough back that it has little bearing on the present state of the soil
2. Lower resolution. Checking every 2 hours seems a bit excessive, as long as our window is short enough that it doesn't miss most rainstorms we should be alright
3. Fewer instances. Particularly because some of my ideas for updating ground snow more accurately also want a more robust history of precipitation in unloaded chunks, there is something to be said for saving this info per chunk, and then both snow systems and farm systems could just query that, rather than running the same simulation for every block.

src/main/java/net/dries007/tfc/common/blockentities/FarmlandBlockEntity.java
```diff
@@ -108,6 +178,28 @@ public void setNutrientWithoutSync(NutrientType type, float value)
         }
     }
 
+    @Override
+    public float getAccumulatedRainfall() {
+        return accumulatedRainfall;
+    }
+
+    @Override
+    public void setAccumulatedRainfall(float accumulatedRainfall) {
```
eerussianguy: braces on newlines

src/main/java/net/dries007/tfc/common/blockentities/IFarmland.java
```diff
@@ -101,6 +122,21 @@ default boolean isMaxedOut()
         return getNutrient(NITROGEN) == 1 && getNutrient(PHOSPHOROUS) == 1 && getNutrient(POTASSIUM) == 1;
     }
 
+    default void updateAccumulatedRainfall(Level level, BlockPos pos, long fromTick, long toTick)
+    {
+        //if(level.getHeightmapPos(Heightmap.Types.WORLD_SURFACE, pos).equals(pos))
```
eerussianguy: either comment it out or don't. I assume you wanted to check if the crop can see sky? There's a function for that.

src/main/java/net/dries007/tfc/common/blockentities/FarmlandBlockEntity.java
eerussianguy: Lower res especially over long periods makes a lot of sense to me. Cause after ~6 months the impact in general of granular simulation should be minimal. And when you stretch over long periods -- eventually you might hit a spot where the crop dies, right? And that would feel punishing IMO

src/main/java/net/dries007/tfc/common/blockentities/IFarmland.java
seelderr: Yea I couldn't figure it out so I left that there as a reminder when I got back to it

src/main/java/net/dries007/tfc/common/blockentities/FarmlandBlockEntity.java
seelderr: I knew this was an issue but hadn't fully decided on how I wanted to approach it yet; I think storing the data per chunk is probably the best option here, especially because it fixes another problem with the system which is that freshly placed farmland starts at a lower than normal hydration since it hasn't been rained on yet

seelderr: I also think there should be a sort of "grace period" for the crop dying due to bad hydration conditions, so that a small stretch of bad hydration won't instantly kill the crop.

src/main/java/net/dries007/tfc/common/blocks/soil/FarmlandBlock.java
seelderr: I think groundwater is probably best for this calculation then.

eerussianguy:
This approach looks good so far!

seelderr:
Have been testing a bit, and I can't figure out why the rainfall accumulation doesn't seem to be saving correctly; is there anything out of the ordinary in the saving logic for the rainfall that is messed up?

src/main/java/net/dries007/tfc/common/blockentities/FarmlandBlockEntity.java
```diff
@@ -108,6 +169,28 @@ public void setNutrientWithoutSync(NutrientType type, float value)
...
+    @Override
+    public float getAdditionalWater() {
+        return additionalWater;
+    }
+
+    @Override
+    public void setAdditionalWater(float additionalWater) {
+        this.additionalWater = Mth.clamp(additionalWater, 0, MAX_ADDITIONAL_WATER);
+        markForSync();
+    }
```
eerussianguy: apply code style

src/main/java/net/dries007/tfc/network/ChunkRainfallPacket.java
```diff
@@ -0,0 +1,47 @@
+package net.dries007.tfc.network;
```
eerussianguy, line 7: run updateLicenses

eerussianguy:
> Have been testing a bit, and I can't figure out why the rainfall accumulation doesn't seem to be saving correctly; is there anything out of the ordinary in the saving logic for the rainfall that is messed up?

I don't know a lot about chunk data, but seems like `setUnsaved` needs to be called?

seelderr:
> > Have been testing a bit, and I can't figure out why the rainfall accumulation doesn't seem to be saving correctly; is there anything out of the ordinary in the saving logic for the rainfall that is messed up?
> 
> I don't know a lot about chunk data, but seems like `setUnsaved` needs to be called?

I already call setUnsaved when I call setLastRainTick; is there somewhere else it needs to be called?

src/main/java/net/dries007/tfc/common/blockentities/FarmlandBlockEntity.java
seelderr: what to run in the gradle for this?

PR #2904, opened 2025-04-05
InteractionManager code struction rewritten(1.21.x)

gangzhengzhiwei:
Rewritten code struction of InteractionManager.Make it much easier to mixin/modify tfc's default interaction.


eerussianguy:
Yeah this rewrite is not happening. You haven't really even articulated why it's necessary. Why not just an API hook to replace a interactionmanager action in the registry?

PR #2906, opened 2025-04-06
Propick Improvements

blakehelton:
- Added text to Prospecting results indicating the distance of the found block (translation only available in English)
- Added a Server Config option to disable false negatives for Prospecting


eerussianguy:
Thanks for sharing, but we 1.18 left support in November 2023. We are currently targeting 1.21.x

PR #2963, opened 2025-07-11
Replace Biome.getPrecipitationAt() redirects with an inject

danielpetkau:
1.20 TFC currently bypasses the Biome.getPrecipitationAt() method using mixin redirects, which breaks compatibility with other mods. For example, the mod Particle Rain calls Biome.getPrecipitationAt() to figure out if it's raining or snowing at a specific location. Using an inject instead makes these mods compatible with each other, and removes the need for multiple redirects of getPrecipitationAt.

### DIFF
--- a/src/main/java/net/dries007/tfc/mixin/BiomeMixin.java
+++ b/src/main/java/net/dries007/tfc/mixin/BiomeMixin.java
@@ -13,7 +13,8 @@ import net.minecraft.client.Minecraft;
 import net.minecraft.core.BlockPos;
 import net.minecraft.world.level.LevelReader;
 import net.minecraft.world.level.biome.Biome;
-
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
 import net.dries007.tfc.util.climate.Climate;
 import net.dries007.tfc.world.biome.BiomeBridge;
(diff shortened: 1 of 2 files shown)


src/main/java/net/dries007/tfc/mixin/BiomeMixin.java
```diff
@@ -62,4 +65,17 @@ private boolean shouldFreezeWithClimate(Biome biome, BlockPos pos, LevelReader l
...
+
+    /**
+     * Replace {@link Biome#getPrecipitationAt(BlockPos)} with a method that takes our climate model into account
+     * <p>
+     * This is better than redirecting known calls to getPrecipitationAt because other mods sometimes
+     * use this method to get weather information.
+     */
+    @Inject(method = "getPrecipitationAt", at = @At("HEAD"), cancellable = true)
+    private void getPrecipitationFromClimate(BlockPos pos, CallbackInfoReturnable<Biome.Precipitation> cir) {
+        Minecraft minecraft = Minecraft.getInstance();
```
Notenoughmail, line 80: This will crash on dedicated servers (specifically during `ServerLevel#tickChunk`) when called due to client only classes not existing on server distributions

However TFC's original redirects were only in `LevelRenderer` anyway, so only changing the behavior after a dist check should preserve the previous behavior

Issue #2884, opened 2025-03-12
Dispenser interaction with gunpowder-filled ceramic bowls causing crash


JotredDev:
Problem:
Interaction between a dispenser using a Flint and Steel and a gunpowder-filled ceramic bowl causes the game to crash, and, despite triggering the explosion, does not consume durability from the Flint and Steel.

Intended Behaviour:
No crash should happen, especially as gunpowder-filled ceramic bowls are very unsafe to trigger by hand. Furthermore, the Flint and Steel should lose a point of durability, to match what happens when a player uses one to trigger the explosion of a gunpowder-filled ceramic bowl.

Reproduction steps:
1. Open a TFC instance of your choice (for reproduction purposes I chose to _only_ use the most up-to-date versions of TFC and Patchouli available on CurseForge at time of writing, those being `TerraFirmaCraft-Forge-1.20.1-3.2.14.jar` and `Patchouli-1.20.1-84.1-FORGE.jar`)
2. Make a new world
3. Set up a gunpowder-filled ceramic bowl and dispenser facing the bowl, like depicted in this screenshot (dispenser may also be above or below the bowl, as long as it faces the bowl)
![Image](https://github.com/user-attachments/assets/d20d9f09-b2fb-414d-aab9-ce345df6570d)
4. Put a Flint and Steel in the dispenser
5. Power the dispenser

Upon powering the dispenser, the explosion is triggered, directly followed by the game crashing with the following error message:
`exception ticking world
Error: java.lang.IllegalArgumentException: Cannot get property DirectionProperty{name=facing, clazz=class net.minecraft.core.Direction, values=[north, east, south, west, up, down]} as it does not exist in Block{minecraft:air}`

eerussianguy:
Please post the full log.

JotredDev:
The server crash log is [here](https://mclo.gs/ILgG8v2), latest.log is [here](https://mclo.gs/kc3h4MC), debug.log is [here](https://mclo.gs/RqzxJkU)

PR #2974, opened 2025-07-26
Fix crash when dispenser activates gunpowder bowl

danielpetkau:
This fixes a crash that occurs when a dispenser uses a flint and steel to explode a gunpowder filled ceramic bowl (issue #2884). The problem occurs because the dispenser activates the ceramic bowl immediately, which destroys the dispenser before it can finish the dispensing code. It then tries to query the facing direction of the dispenser in order to show particles in the correct direction, but the dispenser was already replaced with air, causing a game crash.

The fix is to check the orientation of the dispenser _before_ triggering the start fire event. I wanted to do this by overriding the `.dispense()` method of `DefaultDispenseItemBehavior`, but the method is marked as final. I got around this by making the TFC flint and steel behavior directly extend `DispenseItemBehavior` and have a field for the default behavior. This removed the need to use the `MultipleItemBehavior` class.

Since the flint and steel dispensing was the only use of the multiple item behavior class, I removed it. This is technically a breaking change since it was a public class, but I think it's very unlikely any addons used it. I searched in the code of every addon referenced in the TFC field guide (artisanal, beneath, firmalife, lithiccoins, rnr, afc, firma:civ) and didn't find any references to `MultipleItemBehavior`.

### DIFF
--- a/src/main/java/net/dries007/tfc/util/DispenserBehaviors.java
+++ b/src/main/java/net/dries007/tfc/util/DispenserBehaviors.java
@@ -161,25 +161,45 @@ public final class DispenserBehaviors
         }
     };
 
-    public static final OptionalDispenseItemBehavior TFC_FLINT_AND_STEEL_BEHAVIOR = new OptionalDispenseItemBehavior() {
+    public static final DispenseItemBehavior TFC_FLINT_AND_STEEL_BEHAVIOR = new DispenseItemBehavior()
+    {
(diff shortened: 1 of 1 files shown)


eerussianguy:
1) Please don't remove the class, let's just be smart here
2) Please put braces on newlines

eerussianguy:
Ping me on discord if you would like to be in the contributors channel

eerussianguy:
I've taken care of these changes on my end myself

Issue #2971, opened 2025-07-24
Game crashed with  Create bamboo machine


CurtainOne:
The game crashed when I build machine like this 

<img width="815" height="737" alt="Image" src="https://github.com/user-attachments/assets/2d3eb6be-8487-4955-9d7a-9f151b70e798" />

First I set 21RPM to the motor and then changed speed , game crashed.

https://github.com/user-attachments/assets/c621029d-54b0-46c2-8e26-ee73caa2d6f9

Game version is 1.20.1.
A crash log attached below.

[crash-2025-07-25_01.20.24-server.txt](https://github.com/user-attachments/files/21415979/crash-2025-07-25_01.20.24-server.txt)

PR #2975, opened 2025-07-26
Fix crash with Create mod's deployer

danielpetkau:
Fixes issue #2971. Create mod's deployer sometimes simulates a right click action with a player entity that has already been marked as removed. This causes a game crash when trying the drink water right click action, because you can't get the player data of a player marked as removed. I added a check for this, preventing the crash.

This could be a bug with create mod, since it seems strange to right click with an already removed player, but I figured it was worth it to fix it from TFC's end since it's so simple.


danielpetkau:
The issue doesn't seem to happen in 1.21

PR #2991, opened 2025-07-31
(1.21) Make jam in jars inedible

danielpetkau:
Unsealed jam jars are currently edible in 1.21, which is an unintentional change from 1.20. They don't give any hunger or saturation, and they don't give back the glass jar when eaten.


eerussianguy:
`runData` needed, lol

Issue #2759, opened 2024-08-12
[3.2.7] Snow piles forget flowers when shoveled


Alex33212:
Breaking snow piles that contain flowers with a shovel simply removes the snow without placing the flower; grasses, stones, and some other plants (e.g. Athyrium ferns) behave normally, as do snow piles broken by hand.

alcatrazEscapee:
Please confirm you're not just breaking the snow pile and content (i.e. a flower, which would instant-break) too quickly to notice.

Alex33212:
Yep, I made sure of it - I tested using a field of dandelions after first noticing, and pointed my shovel at the very edge of the snow pile. When I was testing if it still happened when not using a shovel, I did have an instance of breaking the flower at the same time, but it still played the plant-breaking sound.

Alex33212:
Specific flowers I've had this happen with (occurred to me it might be a limited selection):
- Dandelions
- Red & pink tulips
- Lily of the valley

I'll do some more testing tomorrow, see if there is a specific subset of flowers that do it

alcatrazEscapee:
This proved wildly more difficult to track down than initially anticipated:

- I **cannot** reproduce this in Creative mode. Breaking a snow pile always places a plant back
- I **cannot** reproduce this in Survival, **with a wooden shovel** (importantly, any shovel that does not instant-mine snow layers)
- I **can** reproduce this in Survival, with any shovel that is sufficient to instant-mine snow (i.e. bronze)

This indicates that there's a specific problem with instant-mine on survival, which does in vanilla have some edge case behavior. Secondly, the sequence of events when this fails looks like such:

1. Client starts breaking a block (`Minecraft.startAttack()`) at the location of the snow layer
2. Client predicts a block break (instantly, due to low enough hardness), and sends a packet to the server to record the block break
3. Server (integrated), receives the packet and updates the block (it is now a tulip) in the chunk
4. Client continues breaking (`Minecraft.continueAttack()`). Note at this point, the client is in a weird state - it is ray tracing towards the same block that it originally started breaking, so sends a packet to try and break a second block.
5. Server updates the block below the tulip (removes the snowy state)
6. Server receives the second packet, and does the check - the client was targeting a block, which was instant-breakable, and breaks the tulip (leaving air)
7. Client receives an update, and sets the block broken to air.

[...]

PR #2993, opened 2025-08-01
(1.21) Fix snow and charcoal pile breaking

danielpetkau:
Issue #2759 popped back up in 1.21, meaning snow pile breaking didn't work properly. I found that it affects not just snow piles but also vanilla snow, and charcoal pile breaking. You can replicate this in the 1.20 or 1.21 versions of the mod, instantly mining charcoal piles often mines 2 layers due to a bug in the vanilla game where the same block can be instabroken twice in the same tick. 
This PR fixes the vanilla bug that caused the issues, and cleans up the charcoal pile and snow pile code to go with it.

this took days I'm losing my marbles

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blocks/SnowPileBlock.java
+++ b/src/main/java/net/dries007/tfc/common/blocks/SnowPileBlock.java
@@ -74,14 +74,13 @@ public class SnowPileBlock extends SnowLayerBlock implements IForgeBlockExtensio
 
     public static void removePileOrSnow(LevelAccessor level, BlockPos pos, BlockState state)
     {
-        removePileOrSnow(level, pos, state, Optional.empty());
+        removePileOrSnow(level, pos, state, level.getBlockEntity(pos, TFCBlockEntities.PILE.get()).orElse(null));
     }
 
     /**
-     * @param snowPile If {@code null}, then there is no provided block entity and one should be queried from the world. If not-null, this
-     *                 represents the block entity present in the world, possibly empty.
+     * @param snowPile The snow pile block entity present in the world, possibly null.
      */
-    public static void removePileOrSnow(LevelAccessor level, BlockPos pos, BlockState state, @Nullable Optional<PileBlockEntity> snowPile)
+    public static void removePileOrSnow(LevelAccessor level, BlockPos pos, BlockState state, @Nullable PileBlockEntity snowPile)
     {
         final int layers = state.getValue(SnowLayerBlock.LAYERS);
         if (layers > 1)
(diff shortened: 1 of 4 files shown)


src/main/java/net/dries007/tfc/common/blocks/SnowPileBlock.java
```diff
@@ -100,27 +97,25 @@ else if (state.getBlock() == Blocks.SNOW)
         else
         {
             // Otherwise, remove a snow pile, restoring the internal states
-            level.getBlockEntity(pos, TFCBlockEntities.PILE.get()).ifPresent(pile -> {
-                if (!level.isClientSide())
+            if (snowPile == null) snowPile = level.getBlockEntity(pos, TFCBlockEntities.PILE.get());
+            snowPile.ifPresent(pile -> {
```
eerussianguy: So I am fine with this b/c it's not actually bad practice but we don't usually use this type of expression.

```
if (level.getBlockEntity(pos) instanceof SnowPileBlockEntity pile)
{
   ...
}
```

That's more flexible because a) we're not couched inside of a lambda function and thus limited in what outside variables we can bring in b) we are not switching on the type

Minecraft wants you to gather entities based on the registered type, this is actually less good for modded in my opinion--if a mod adds their own snow pile, for some reason, it is automatically incompatible without a mixin because the list of blocks that are tied to a block entity is essentially locked at registration time. An instanceof is a more flexible way to express this because it's not like anyone will unintentionally inherit from our pile block entity.

src/main/java/net/dries007/tfc/mixin/client/MinecraftMixin.java
```diff
@@ -7,11 +7,17 @@
...
+import net.minecraft.client.multiplayer.MultiPlayerGameMode;
+
+import org.jetbrains.annotations.Nullable;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
+import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
```
eerussianguy, line 19: I guess I am not used to seeing mixin extras used, I assume this just comes packed in norge now?

src/main/java/net/dries007/tfc/common/blocks/SnowPileBlock.java
```diff
@@ -149,8 +144,9 @@ public ExtendedProperties getExtendedProperties()
     @Override
     public boolean onDestroyedByPlayer(BlockState state, Level level, BlockPos pos, Player player, boolean willHarvest, FluidState fluid)
     {
-        playerWillDestroy(level, pos, state, player);
-        removePileOrSnow(level, pos, state);
+        final Optional<PileBlockEntity> snowPile = level.getBlockEntity(pos, TFCBlockEntities.PILE.get()); // Store the blockentity before it is removed
```
eerussianguy: Why not `level.getBlockEntity(...).orElse(null)`.

A nullable optional is in my opinion an antipattern

eerussianguy (changes requested):
Unfortunately banging your head against the wall is the only way to solve problems like this. Sometimes I wonder if people wonder if I work on the mod. It's cause I open it up, spend 90 minutes debugging something like this, and then run out of time to work on the damn thing lol.

Anyway I left some thoughts... let's just be clearer about the pattern we want to use here.

src/main/java/net/dries007/tfc/mixin/client/MinecraftMixin.java
danielpetkau: It's used in other places in TFC, it's possible someone else added it, I'm not sure

eerussianguy (approved):

PR #2994, opened 2025-08-01
(1.21) Fix TFC damage types dealing knockback

danielpetkau:
All of TFC's custom damage types are intended not to deal knockback, but they're not added to the no_knockback damage type tag. This adds them to the tag, so things like dehydration and standing on a grill don't deal knockback.


eerussianguy:
TIL they added that as a tag.

PR #2995, opened 2025-08-01
Fix BiomeBasedClimateModel crash

### LINKED ISSUE #2992: [1.20.1/3.2.17] BiomeBasedClimateModel Causes Cyclical Call Stack in Rare Cases
Full log: https://mclo.gs/Bo41lfc

Due to the changes introduced in #2963, the default climate model can cause a cyclical call stack of 

- `BiomeBasedClimateModel#getRainfall` -->
- `Biome#getPrecipitationAt` --mixin->
- `Climate#getPrecipitation` -->
- `EnvironmentHelpers#isRainingOrSnowing` -->
- `WorldTracker#isRaining` -->
- `Climate#getRainfall` -->
- `BiomeBasedClimateModel#getRainfall`

in cases where the player enters a dimension with a biome based climate while the level is currently raining (and presumably the client chunk cache hasn't been populated, to hazard a guess)

I am able to reliably recreate this issue with just TFC and Patchouli and no extra datapacks by setting the weather to rain, waiting until it is visibly raining, and then teleporting to a dimension with a biome based climate model (i.e. the end/nether).

danielpetkau:
Fixes #2992 by not replacing the getPrecipitationAt logic if the client level uses a biome based climate

### DIFF
--- a/src/main/java/net/dries007/tfc/mixin/BiomeMixin.java
+++ b/src/main/java/net/dries007/tfc/mixin/BiomeMixin.java
@@ -10,12 +10,15 @@ import java.util.Optional;
 import java.util.function.Supplier;
 
 import net.minecraft.client.Minecraft;
+import net.minecraft.client.multiplayer.ClientLevel;
 import net.minecraft.core.BlockPos;
 import net.minecraft.world.level.LevelReader;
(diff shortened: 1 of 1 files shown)


Pyritie:
thank you for the quick fix!

PR #3006, opened 2025-08-05
(1.21) Make shift clicking gear boxes toggle opposite side

danielpetkau:
Currently, the best way (that I've found) to make two adjacent gear boxes connect to each other is clipping into one of them with a minecart. This allows an intended way to do it, by shift + right clicking with a hammer you can toggle the opposite side of the one targeted.

I couldn't find a better way to tell if a block is a gear box block than instanceof, let me know if there is one.


eerussianguy:
See comments on Discord

PR #3009, opened 2025-08-06
(1.21) Make thatch blocks usable in pit kilns + fixes/tweaks

danielpetkau:
- Made thatch blocks usable in pit kilns, they count as 4 straw
- Made unlit pit kilns drop contents when broken
- Made lit pit kilns unbreakable
- Fixed unlit pit kilns making a sizzling noise when last straw is taken
- Made the pick block behavior of pit kilns a bit more predictable


src/main/java/net/dries007/tfc/common/blocks/devices/PlacedItemBlock.java
```diff
@@ -181,7 +182,7 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
         if (placedItem != null)
         {
             final ItemStack held = player.getItemInHand(hand);
-            if (Helpers.isItem(held.getItem(), TFCTags.Items.PIT_KILN_STRAW) && !held.isEmpty() && PitKilnBlockEntity.isValid(level, pos))
+            if ((Helpers.isItem(held.getItem(), TFCTags.Items.PIT_KILN_STRAW) || (held.getItem() == TFCBlocks.THATCH.asItem())) && !held.isEmpty() && PitKilnBlockEntity.isValid(level, pos))
```
eerussianguy: Let's use a tag here specifically for this behavior if it is easy to do. like `#tfc:pit_kiln_4_straw`

src/main/java/net/dries007/tfc/common/blocks/devices/PitKilnBlock.java
```diff
@@ -207,9 +217,10 @@ public ItemStack getCloneItemStack(BlockState state, HitResult result, LevelRead
         return ItemStack.EMPTY;
     }
 
+    @SuppressWarnings("deprecation")
     @Override
-    protected void beforeRemove(InventoryBlockEntity<?> entity)
+    protected float getDestroyProgress(BlockState state, Player player, BlockGetter level, BlockPos pos)
     {
-        // No-op - we don't want to call ejectInventory()
+        return state.getValue(STAGE) == LIT ? 0F : super.getDestroyProgress(state, player, level, pos);
```
eerussianguy, line 219: I don't understand why a lit pit kiln should be unbreakable. What if you decided you didn't want to burn it anymore? What if you break the blocks around the pit kiln thus invalidating it?

danielpetkau: Breaking the blocks around the pit kiln still works, and I think that's the intuitive way to stop a pit kiln. Breaking the pit kiln is quite buggy because the fire constantly gets replaced, so you get the buggy server desync on the block breaking. Breaking the pit kiln itself also doesn't drop any items, while breaking a dirt block around the pit kiln keeps the items on the ground.

PR #3011, opened 2025-08-06
(1.20) Fix bird behavior with full nestboxes

DCNick3:
Currently, if a nest box is full, a duck will repeatedly try and fail to lay an egg there, wasting wear & tear.

This PR improves the behavior in two ways:
1. Stops increasing wear & tear when an egg fails to be laid into a full nestbox
2. Prevents ducks from pathfinding to full nestboxes, so they would rarely even try to lay eggs into full nestboxes.

To implement 2, I added a boolean blockstate property `full` that allows to filter for non-full nestboxes in `NearestNestBoxSensor` without looking up the corresponding block entity.

I think that merging 1 is very important to fix and is quite simple. 2 is nice to have, but not 100% necessary, as not doing it will only slow down breeding, but not lead to birds tearing themselves apart.

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
@@ -6,6 +6,10 @@
 
 package net.dries007.tfc.common.blockentities;
 
+import java.util.Optional;
+
+import net.dries007.tfc.util.events.AnimalProductEvent;
+
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
 import net.minecraft.network.chat.Component;
@@ -17,6 +21,7 @@ import net.minecraft.world.inventory.AbstractContainerMenu;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.state.BlockState;
+import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.items.ItemStackHandler;
 
 import net.dries007.tfc.common.capabilities.PartialItemHandler;
@@ -28,6 +33,7 @@ import net.dries007.tfc.common.entities.livestock.OviparousAnimal;
 import net.dries007.tfc.config.TFCConfig;
 import net.dries007.tfc.util.Helpers;
 import net.dries007.tfc.util.calendar.Calendars;
+
 import org.jetbrains.annotations.Nullable;
 
 import static net.dries007.tfc.TerraFirmaCraft.MOD_ID;
@@ -46,13 +52,25 @@ public class NestBoxBlockEntity extends TickableInventoryBlockEntity<ItemStackHa
                 {
                     if (bird.getRandom().nextInt(7) == 0)
                     {
-                        Helpers.playSound(level, pos, SoundEvents.CHICKEN_EGG);
-                        if (Helpers.insertOne(level, pos, TFCBlockEntities.NEST_BOX.get(), bird.makeEgg()))
+                        final AnimalProductEvent event = bird.makeEggEvent();
+
+                        if (!MinecraftForge.EVENT_BUS.post(event))
                         {
-                            bird.setFertilized(false);
-                            bird.setProductsCooldown();
-                            bird.stopRiding();
-                            nest.markForSync();
+                            if (Helpers.insertOne(Optional.of(nest), event.getProduct()))
+                            {
+                                Helpers.playSound(level, pos, SoundEvents.CHICKEN_EGG);
+
+                                bird.addUses(event.getUses());
+                                bird.setFertilized(false);
+                                bird.setProductsCooldown();
+                                bird.stopRiding();
+                                nest.markForSync();
+                            }
+                            else
+                            {
+                                // the nest box is full, leave it be
+                                bird.stopRiding();
+                            }
                         }
                     }
                 }
--- a/src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
@@ -8,6 +8,7 @@ package net.dries007.tfc.common.blockentities;
 
 import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
```diff
@@ -42,17 +46,26 @@ public static void serverTick(Level level, BlockPos pos, BlockState state, NestB
...
+
+                                bird.addUses(event.getUses());
+                                bird.setFertilized(false);
+                                bird.setProductsCooldown();
+                                bird.stopRiding();
+                                nest.markForSync();
+                            } else {
+                                // the nest box is full, leave it be
+                                bird.stopRiding();
+                            }
```
eerussianguy: Please import and apply codestyle! We have an XML in the repo. Failing that, Please organize your braces like so:

```
if
{
..
}
else
{
...
}
```

src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
```diff
@@ -42,17 +46,26 @@ public static void serverTick(Level level, BlockPos pos, BlockState state, NestB
...
             if (sitter instanceof OviparousAnimal bird)
             {
-                if (bird.isReadyForAnimalProduct())
+                if (bird.isReadyForAnimalProduct() && !nest.isFull())
                 {
                     if (bird.getRandom().nextInt(7) == 0)
                     {
-                        Helpers.playSound(level, pos, SoundEvents.CHICKEN_EGG);
-                        if (Helpers.insertOne(level, pos, TFCBlockEntities.NEST_BOX.get(), bird.makeEgg()))
+                        var event = bird.makeEggEvent();
```
eerussianguy: We prefer not to use `var` outside of situations with very verbose object names.

We prefer `final` before final local variables

src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
```diff
@@ -116,4 +136,8 @@ public AbstractContainerMenu createMenu(int windowID, Inventory inv, Player play
     {
         return NestBoxContainer.create(this, inv, windowID);
     }
+
+    private boolean isFull() {
+        return Helpers.isFull(inventory);
+    }
 }
```
eerussianguy: brackets on newlines

src/main/java/net/dries007/tfc/common/entities/livestock/OviparousAnimal.java
```diff
@@ -222,7 +221,7 @@ public boolean causeFallDamage(float amount, float speed, DamageSource src)
     }
 
     @SuppressWarnings("unchecked")
-    public ItemStack makeEgg()
+    public AnimalProductEvent makeEggEvent()
```
eerussianguy: This is a breaking change. Please find a way to not change the method signature. If needed, make a new event.

eerussianguy (changes requested):

src/main/java/net/dries007/tfc/common/entities/livestock/OviparousAnimal.java
DCNick3: Can you please tell me/point me towards some resource that will tell me what considerations should I make about public APIs?

Some questions I have trouble answering:
- What is a part of public API? Is it every public member of public classes?
- What promises does TFC make about its public API? Can I, for example, keep the signature the same, but change the behavior to make the caller be responsible for adding uses to the animal?
- Is `makeEgg` even used by anything outside of TFC? What kind of behaviors can its users expect from it? That is, is there known impact of this hypothetical breaking change?

Also, I am not sure how making a new event will help me with not changing the method signature. Can you please clarify?

src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
DCNick3: I was able to import the codestyle into Intellij Idea, although just blindly running a reformat on the modified code did touch some unrelated sections. I did try my best to apply it where relevant though.

src/main/java/net/dries007/tfc/common/entities/livestock/OviparousAnimal.java
eerussianguy: Hi, sorry for the delay in getting back to you. @DCNick3 

Minecraft mods, in general, unless they do export them as separate sources, do not have APIs. Or, the whole mod can be thought of as the API. This is also how we depend on MC itself, where all of MC is the API.

Maybe if this was earlier in 1.20's dev cycle I wouldn't care, but our promise is generally not to change the signatures of a method someone could possibly be using. The methods in OviparousAnimal are certainly one I could imagine a mod using. I don't consider searching on github to be an appropriate method of determining whether the method is actually depended on by other mods. If we can hold a general rule about doing this, then we are always protected.

What we are protecting from is not the behavior necessarily of methods--we are protecting against the game being able to load in the future. If a hypothetical mod uses this method in some way and its signature changes, that mod is now broken, and may never be able to be fixed if the author is dormant. That's a huge issue.

Lastly, that was a typo, I meant to say make a new method, not a new event.

DCNick3: Ah, making a new ~~event~~ method does make sense. I will do that.

UPD: I made the same exact typo as you... damn

eerussianguy:
Making a new event is also a breaking change since receivers of the event will no longer be able to do anything, please do not make a new event. Just please make a new method.

src/main/java/net/dries007/tfc/common/entities/livestock/OviparousAnimal.java
DCNick3: I have updated the PR to not delete the `makeEgg` method. Instead, it marks it as being deprecated and implements it through the new `makeEggEvent`. This should be compatible with the API usage, unless a transformer is used to modify `makeEgg`. Hope that this compatibility trade-off is alright.

DCNick3:
@eerussianguy Hi, could you please take a look at this again?

DCNick3:
@eerussianguy another gentle ping

PR #3040, opened 2025-08-18
(1.21) Misc bugfixes 4

danielpetkau:
- Fix fired molds being the wrong weight
- Fix molds missing fluid tooltip
- Fix molds with fluids being stackable
- Fix TFC coals not being tagged as coal
- Fix opening powderkegs not consuming right click action (results in block placement on client side)
- Fix powderkeg inventory always being locked
- Fix identical sandwich stacks being unable to be combined
- Make powderkegs give themselves a block update when sealed (in case they're already powered)
- Fix powderkegs not having explosion sounds/particles
- Make powderkegs extinguish when unsealed from ui, to match shift+right click behavior


src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
```diff
@@ -55,10 +56,16 @@ public static void toggleSeal(Level level, BlockPos pos, BlockState state)
...
+                if (state.getValue(LIT))
+                {
+                    powderkeg.setLit(false, null);
+                    Helpers.playSound(level, pos, SoundEvents.FIRE_EXTINGUISH);
+                }
                 powderkeg.onUnseal();
             }
             else
             {
+                level.neighborChanged(pos, TFCBlocks.POWDERKEG.get(), pos); // Update the powderkeg so it checks if it should light
```
eerussianguy, line 68: `this`

danielpetkau: It's a static method

PR #3046, opened 2025-08-30
rework forge canSeeSky to use FORGE_INVISIBLE_WHITELIST (1.20.x)

Nashy1232:
Reworked the CharcoalForgeBlock to use a modified canSeeSky method that takes into consideration the FORGE_INVISIBLE_WHITELIST

Before: (note the brick chimney i was using was tagged with FORGE_INVISIBLE_WHITELIST)
http://share.nashy.uk/2025/08/javaw_m2COqAiSoJ.mp4

After:
http://share.nashy.uk/2025/08/javaw_XlIoMGPg3M.mp4

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blocks/devices/CharcoalForgeBlock.java
+++ b/src/main/java/net/dries007/tfc/common/blocks/devices/CharcoalForgeBlock.java
@@ -51,7 +51,7 @@ public class CharcoalForgeBlock extends DeviceBlock implements IBellowsConsumer
 
     static
     {
-        BiPredicate<LevelAccessor, BlockPos> skyMatcher = LevelAccessor::canSeeSky;
+        BiPredicate<LevelAccessor, BlockPos> skyMatcher = CharcoalForgeBlock::canSeeSky;
         BiPredicate<LevelAccessor, BlockPos> isValidSide = (level, pos) -> isForgeInsulationBlock(level.getBlockState(pos));
         BlockPos origin = BlockPos.ZERO;
         FORGE_MULTIBLOCK = new MultiBlock()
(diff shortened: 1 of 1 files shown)


Nashy1232:
Question to add to this, how would you feel about chaning it to only check say 20 blocks above as opposed to checking to max height?

eerussianguy:
Hi there's no way I can accept this PR, this is an incredibly inefficient check that ignores the principle of how `canSeeSky` works.

eerussianguy:
To elaborate, checking `canSeeSky` just checks light level at a location, whereas your proposed check would query the blockstate of hundreds and hundreds of blocks every time this check runs. I'm not seeing a way I can really allow your desired behavior without rewriting lots of the forge's code.

Nashy1232:
im new to this so excuse the ignorance, but is there anywhere i can see the code behind canSeeSky?

eerussianguy:
Open your dev environment, ctrl+click on canSeeSky

Nashy1232:
oooh. Yeah I 100% understand not implementing my change. I'm going to take a while to try and think about a way to implement a reasonable soluiton.

eerussianguy:
Closing for now

Nashy1232:
I have thought of this soultion that addresses your concerns for performance. While it is not the most ideal solution i would like your opintion.

`    
private static boolean canSeeSky(LevelAccessor level, BlockPos pos)
    {
        /*
         * Only check up to 20 blocks above to limit performance impact.
         * Check to make sure it doesnt exceed the world limit
         * Checks canSeeSky for each block above the forge up to 20 blocks
         * If any block above can see the sky, return true
         * If cant see the sky check next block above,
         * If block is whitelisted continue the checks else return false
         */

        BiPredicate<LevelAccessor, BlockPos> skyMatcher = LevelAccessor::canSeeSky;
        int maxYCheck = Math.min(level.getMaxY(), pos.getY() + 20);

        for (int y = pos.getY() + 1; y < maxYCheck; y++) {
            // is the block above able to see the sky?
            if (skyMatcher.test(level, pos.above())) {
                // if yes, return true
                return true;
            }
            // if not, is it whitelisted?
            else {
                // if not whitelisted, return false
                BlockState state = level.getBlockState(pos.above(y));
                if (!Helpers.isBlock(state, TFCTags.Blocks.FORGE_INVISIBLE_WHITELIST)) {
                    return false;
                }
            }
            // if not whitelisted, check the next block above
        }
        // if we reach here, nothing above can see the sky
        return false;
    };
`

Issue #3054, opened 2025-09-02
Rock Types can vary when using the re-create world option


Therighthon:
Steps to reproduce:
1. Create a world and note the rocks in it
2. Leave it, re-create it, and note that the rocks in the copy are usually different (will still be Uplift/Continental etc, but different individual rocks selected)

The issue appears to be related to the the re-create world button, as creating a new world with a given seed will yield the same rock types every time. 
Further, if you re-create a world multiple times, it will always have the same set of rocks as the first time you re-created the world

Notenoughmail:
I believe I've found the cause for this problem... and its due to compound tags not preserving insertion order.

When creating a new world, the order present in the json file is preserved due to going through `JsonOps` and a `JsonObject`, which internally uses a `LinkedTreeMap`

<img width="1020" height="172" alt="Image" src="https://github.com/user-attachments/assets/e461e5bb-3eea-429b-a667-afcd46b4eea7" />

However, on recreating a world, the layer mappings are read from nbt as a `CompoundTag` which does not have insertion order guarantees, since the default constructor uses a `HashMap` as a backing map

<img width="996" height="176" alt="Image" src="https://github.com/user-attachments/assets/caea8cf9-cb4d-42e2-b786-c1c9f9bb2081" />

I believe this also means that the layer mappings are saved to level data in the incorrect order.

PR #3072, opened 2025-09-08
[1.20.x] Preserve Layer Ordering

Notenoughmail:
Addresses #3054 in 1.20 by using a custom codec that serializes to NBT as a list.

This is non-breaking and backwards compatible with worlds created before this change.


Therighthon:
Thank you so much. Confirmed that this works in 1.21, we should be good to accept this for 1.20 as well @eerussianguy

PR #3079, opened 2025-09-12
Low enough hanging fixes

Allmoz:
Small stuff


eerussianguy:
- Please actually list changes that you make, it makes it easier for me to review.
- Unless I'm crazy the changes to tags are not reflected in `BuiltinItemTags` at all, so I'm not sure how you got them there unless you wrote the files manually, which you should not have done.

Allmoz:
Changed generate book, it was calling a non existent tag, didn't add any. Also changed the item generated by the wrought iron grill recipe, that changed its id, so i reflected that also in the book script. Had to run the java data and python generator, that made all the rest. Seems previous commits didn't run data, so those changes to tags are from that

PR #3107, opened 2025-09-15
Updated pt_br.json mod file translations

gifpxto:
The first one.

It will take a little while, but it will be worth it.


gifpxto:
@eerussianguy finished.

PR #3132, opened 2025-09-17
Fixed nutrition of players not updating correctly on both server and client

JotredDev:
When eating food or losing hunger from exhaustion, the `NutritionData#hunger` value wasn't updated. This lead to incorrect nutrition and HP values on both server and client side, which only fixed themselves after leaving and re-joining the world (since saving / loading data _does_ set this value correctly). This PR fixes this bug by updating the hunger value in these missing cases.

### DIFF
--- a/src/main/java/net/dries007/tfc/common/player/PlayerInfo.java
+++ b/src/main/java/net/dries007/tfc/common/player/PlayerInfo.java
@@ -207,11 +207,6 @@ public final class PlayerInfo extends net.minecraft.world.food.FoodData implemen
         addThirst(food.water());
         addIntoxication(food.intoxication());
 
-        if (player instanceof ServerPlayer serverPlayer && nutrition.getAverageNutrition() >= 0.999)
-        {
-            TFCAdvancements.FULL_NUTRITION.trigger(serverPlayer);
-        }
-
         if (food.hunger() > 0)
         {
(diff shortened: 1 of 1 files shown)


src/main/java/net/dries007/tfc/common/player/PlayerInfo.java
```diff
@@ -223,6 +218,13 @@ public void eat(FoodData food)
             this.food.eat(food.hunger(), food.saturation() / (2f * food.hunger()));
         }
 
+        // Add nutrients and update the hunger value in NutritionData
+        if (!player.level().isClientSide)
+        {
+            nutrition.addNutrients(food);
```
danielpetkau, line 219: Why move this below the advancement trigger? It seems like you won't get the advancement anymore unless you eat again right away

danielpetkau (changes requested):

src/main/java/net/dries007/tfc/common/player/PlayerInfo.java
JotredDev: Oh, good catch. I had moved it below the call to `FoodData#eat` to avoid having to compute the food level or check if the level is clientside a second time, just forgot to move the advancement trigger accordingly. I'll fix it now

danielpetkau (approved):

PR #3134, opened 2025-09-18
Making it easier for addon devs to change out `NutritionData` by moving to an interface

JotredDev:
This PR intends to make it easier for addon devs to change out the functionality of the `NutritionData` class, by moving the public methods of this class into an interface `INutritionData`. Furthermore, the places where this class is explicitly used in TFC's code (`IPlayerInfo`, `PlayerInfo`, and `NutritionScreen`) have been changed to use this new interface instead.

Important changes include:
- The former `addNutrition(FoodData data)` has been changed to also take the food level at time of eating as an argument. Although this extra parameter goes unused in the current implementation of `NutritionData`, the downsides to having it are negligible, and it may be useful in case the functionality of this class is changed later on, or if addons require this information for their implementations
- The existing `setHunger` method is now _only_ for setting the hunger value
- The newly added `setHungerAndUpdate` method is for setting the hunger value and then updating the nutrition values of the player, as `setHunger` did previously
- More detailed documentation in the code

### DIFF
--- a/src/main/java/net/dries007/tfc/TerraFirmaCraft.java
+++ b/src/main/java/net/dries007/tfc/TerraFirmaCraft.java
@@ -51,6 +51,7 @@ import net.dries007.tfc.common.fluids.TFCFluids;
 import net.dries007.tfc.common.items.PropickItem;
 import net.dries007.tfc.common.items.TFCItems;
 import net.dries007.tfc.common.player.ChiselMode;
+import net.dries007.tfc.common.player.PlayerInfo;
 import net.dries007.tfc.common.recipes.TFCRecipeSerializers;
 import net.dries007.tfc.common.recipes.TFCRecipeTypes;
 import net.dries007.tfc.common.recipes.ingredients.TFCIngredients;
@@ -128,6 +129,7 @@ public final class TerraFirmaCraft
         bus.addListener(TFCCreativeTabs::setAllTabContentAsNonDecaying);
         bus.addListener(Faunas::registerSpawnPlacements);
         bus.addListener(PacketHandler::setup);
+        bus.addListener(PlayerInfo::setNutritionDataSupplier);
 
         // Core Registries (vanilla)
         TFCBlocks.BLOCKS.register(bus);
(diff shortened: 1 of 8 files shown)


Notenoughmail:
Am I correct in my understanding of how this event is supposed to work? It looks like if an addon wants to change the `INutritionData` of the player, they must post the event with the new supplier on *TFC's* mod-specific event bus.

If this is correct, that is the opposite to how events are normally handled. Typically, in this situation TFC would post the event itself immediately before it attaches nutrition data to the player (or during construction of the `PlayerInfo` object), then addons would listen to the event and change/access[^1] the supplier for TFC to create and apply to the player.

Essentialy, the `INutritionData` field in `PlayerInfo` should be final and created from the `NutritionDataSupplier` in the event, while the `NutritionDataSupplier` field in the event is non-final and has a setter method.

[^1]: What if I want to wrap whatever `NutritionDataSupplier` exists with my own that partially delegates to the original?

JotredDev:
> Am I correct in my understanding of how this event is supposed to work? It looks like if an addon wants to change the `INutritionData` of the player, they must post the event with the new supplier on _TFC's_ mod-specific event bus.
> 
> If this is correct, that is the opposite to how events are normally handled. Typically, in this situation TFC would post the event itself immediately before it attaches nutrition data to the player (or during construction of the `PlayerInfo` object), then addons would listen to the event and change/access[1](#user-content-fn-1-17679bafdc4a5acb405f465512470227) the supplier for TFC to create and apply to the player.
> 
> Essentialy, the `INutritionData` field in `PlayerInfo` should be final and created from the `NutritionDataSupplier` in the event, while the `NutritionDataSupplier` field in the event is non-final and has a setter method.
> ## Footnotes
> 
>     1. What if I want to wrap whatever `NutritionDataSupplier` exists with my own that partially delegates to the original? [↩](#user-content-fnref-1-17679bafdc4a5acb405f465512470227)

That is a good point. I'm not very familiar with events, and much less so with how they are usually handled, so this input is quite useful. I'll implement your suggestions quickly

Notenoughmail:
Looks good!

PR #3184, opened 2025-09-29
Change leather dyeing recipe to use #minecraft:dyeable tag instead.

AtobaAzul:
Changes the leather armor dyeing recipe to use the new #minecraft:dyeable tag.
Missing new lang files.


AtobaAzul:
Every damn dye lang entry is wrong to begin with so I won't bother to fix this (espcially because I don't know how lol)

eerussianguy:
So how does this work for bleaching - does this just set the color to white? Should it be removing the color instead? Idk.

im guessing we’re missing the set of bleaching recipes from previous versions ? (I am on my phone and can’t look)

AtobaAzul:
It sets to white - but it probably should clear the color honestly

eerussianguy:
Yeah what’s the difference from bleach and white dye if that’s what it does? My thoughts

eerussianguy:
Updates on this?

PR #3188, opened 2025-10-01
Fix anemone flickering, randomize anemone close time

### LINKED ISSUE #3161: Anemone state flickering
Anemones will flicker between their open and closed state when the player stands ontop of them. I've confirmed this occurs with only tfc and dependencies. On 4.0.3 beta.

https://github.com/user-attachments/assets/16a4f6eb-6295-42c9-8a49-b1ea66d9d71c

Therighthon:
Making a PR because I am not sure whether we want to check for an entity in a bounding box like this, as I know that is not super optimized, or if I should study how pressure plates work more closely and implement that system.

Closes #3161


eerussianguy:
IMO this check is fine, as long as it works in game!

PR #3191, opened 2025-10-02
1.21.x wind rework

alekiponi:
- change wind to use a new scale and a new distribution
 - the new range is from 0 to 32 m/s or 115 km/s
 - 0 to 8 m/s uses a normal distribution, 8 to 32 m/s are more rare and only happen at a certain threshold
 - rain pushes the mean up by 3.2 m/s
 - altitude adds to wind speed
- makes windmills respect wind speed
- adjustments to wind-affected particles
- make windmills and water wheels stop instead of breaking when obstructed
- add a config to change how rotational speed is displayed
- make the default unit for rotational speed RPM, a much more standard unit
- prevent windmills from rotating with only one blade
- add JADE/TOP tooltips to inform players of problems with windmills and water wheels


src/main/java/net/dries007/tfc/client/ClientForgeEventHandler.java
```diff
@@ -461,7 +461,7 @@ public static void onScreenKey(ScreenEvent.KeyPressed.Pre event)
             Slot slot = inv.getSlotUnderMouse();
             if (slot != null)
             {
-               PacketDistributor.sendToServer(new StackFoodPacket(slot.index));
+                PacketDistributor.sendToServer(new StackFoodPacket(slot.index));
```
eerussianguy, line 464: why are there random whitespace cahnges everywhere

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -38,20 +41,34 @@ public class WindmillBlockEntity extends TickableInventoryBlockEntity<ItemStackH
...
+
+            float windFactor = Math.min(wind, 0.5f) + 0.75f; // clamp below ~57 kmh
+
+            windmill.targetSpeed = windFactor * targetBeforeWind;
+            windmill.markForSync();
+
+            if (isObstructedBySolidBlocks(level, pos, state.getValue(WindmillBlock.AXIS))){
+                // Check every two seconds if the windmill is obstructed, and if so, break
+                level.destroyBlock(pos, true);
+            }
```
eerussianguy: braces on newlines

eerussianguy: Also I'm confused why this is being added. We handle the obstruction check in `clientTick`. By that being called from `serverTick`, we are handling that logic on the server at the correct time. If that doesn't work, find out why. Plus, doing all this logic and *then* checking if the windmill should be destroyed is an antipattern.

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -38,20 +41,34 @@ public class WindmillBlockEntity extends TickableInventoryBlockEntity<ItemStackH
...
         {
-            // Check every two seconds if the windmill is obstructed, and if so, break
-            level.destroyBlock(pos, true);
+            final float targetBeforeWind = Mth.map(state.getValue(WindmillBlock.COUNT), 1, SLOTS, MIN_SPEED, MAX_SPEED);
+
+            float wind = Climate.get(level).getWind(level, pos).length();
+
+            float windFactor = Math.min(wind, 0.5f) + 0.75f; // clamp below ~57 kmh
+
+            windmill.targetSpeed = windFactor * targetBeforeWind;
```
eerussianguy: So here's the logic flow:

- Client is spinning the windmill at speed X
- Up to 2s later, Server calculates target speed Y
- Server sends a sync packet regardless of if Y=X or not
- Client is now spinning at Y
- Client's 'displayed speed' quickly lerps to meet Y

Two problems:
- I don't see why the server needs to calc-then-sync. Wind is available on the client and we can simply run the same calc on either side to keep things up to date. The server needs to know about rotational speed for the purpose of mechanical power. The client needs it for mech. and rendering (which is why it has the lerping functionality that the server doesn't). This is sort of a mindset I've tried to get out of, where we put the server in charge of things that should just be ran on both sides. The reason is that it causes unnecessary syncing, and also is a desync because the server experiences things before the client does.
- This is sending a packet every 2s regardless of what's happening, and we did not need to do that before. Sending constant packets can be a major lag source. I'm not totally sure why this was necessary to add. If it is needed then there should at least be checks that the syncing is necessary.

src/main/java/net/dries007/tfc/util/tracker/WeatherHelpers.java
```diff
@@ -600,4 +603,13 @@ public static boolean isIce(BlockState state)
...
         return state.getBlock() == Blocks.ICE || state.getBlock() == TFCBlocks.ICE_PILE.get() || state.getBlock() == TFCBlocks.SEA_ICE.get();
     }
+
+    public static float windMS(Vec2 wind){
+        return wind.length() * WIND_MS_FACTOR;
+    }
+
+    public static float windKMS(Vec2 wind){
+        return wind.length() * WIND_KMS_FACTOR;
+    }
```
eerussianguy: braces on newlinesplease

eerussianguy (changes requested):

src/main/java/net/dries007/tfc/common/blockentities/rotation/WindmillBlockEntity.java
```diff
@@ -34,40 +37,66 @@ public class WindmillBlockEntity extends TickableInventoryBlockEntity<ItemStackH
...
+            if (!windmill.obstructed && isObstructedBySolidBlocks(level, pos, state.getValue(WindmillBlock.AXIS)))
+            {
+                windmill.obstructed = true;
+                windmill.markForSync();
+            }
+            if (windmill.obstructed && !isObstructedBySolidBlocks(level, pos, state.getValue(WindmillBlock.AXIS)))
+            {
+                windmill.obstructed = false;
+                windmill.markForSync();
+            }
```
eerussianguy:
```java
boolean obstructedNow = isObstructedBySolidBlocks();
if (obstructedNow != windmill.obstructed)
{
    windmill.obstructed = obstructedNow;
    windmill.markForSync();
}
```

src/main/java/net/dries007/tfc/config/ClientConfig.java
```diff
@@ -117,6 +118,14 @@ public class ClientConfig extends BaseConfig
...
         ).define("disabledExperienceBarStyle", DisabledExperienceBarStyle.HOVER);
 
+        rotationDisplayStyle = builder.comment(
+            "Changes the units that rotational speed is displayed in.",
+            "  ROTATIONS_PER_SECOND - Shows rotational speed as a measure of full rotations per second",
+            "  DEGREES_PER_SECOND - Shows rotational speed as a measure of degrees per second",
+            "  REVOLUTIONS_PER_SECOND - Shows rotational speed as a measure of revolutions per second",
+            "  REVOLUTIONS_PER_MINUTE - Shows rotational speed as a measure of revolutions per minute"
+        ).define("rotationDisplayStyle", RotationDisplayStyle.REVOLUTIONS_PER_MINUTE);
+
```
eerussianguy, line 128: I think these might be lowercase-ized in the actual config? check please

src/main/java/net/dries007/tfc/util/tooltip/BlockEntityTooltips.java
```diff
@@ -205,10 +227,7 @@ public static void register(RegisterCallback<BlockEntityTooltip, Block> callback
     public static final BlockEntityTooltip BELLOWS = (level, state, pos, entity, tooltip) -> {
         if (entity instanceof BellowsBlockEntity bellows)
         {
-            final int step = Mth.clamp((int) Mth.map(
-                bellows.getExtensionLength(0f),
-                BellowsBlockEntity.MIN_EXTENSION, BellowsBlockEntity.MAX_EXTENSION,
-                0, 10), 0, 10);
+            final int step = Mth.clamp((int) Mth.map(bellows.getExtensionLength(0f), BellowsBlockEntity.MIN_EXTENSION, BellowsBlockEntity.MAX_EXTENSION, 0, 10), 0, 10);
```
eerussianguy: why was this changed?

src/main/java/net/dries007/tfc/util/tooltip/BlockEntityTooltips.java
```diff
@@ -504,12 +523,43 @@ else if (pot.getOutput() != null && !pot.getOutput().isEmpty())
         {
             for (ItemStack stack : Helpers.iterate(placedItem.getInventory()))
             {
-                if (!stack.isEmpty())
-                    tooltip.accept(stack.getHoverName());
+                if (!stack.isEmpty()) tooltip.accept(stack.getHoverName());
```
eerussianguy: why was this changed?

src/main/java/net/dries007/tfc/util/tracker/WeatherHelpers.java
```diff
@@ -600,4 +603,26 @@ public static boolean isIce(BlockState state)
...
+    public static float windMT(Vec2 wind){
+        return wind.length() * WIND_MS_FACTOR / 20;
+    }
+
+    /**
+     * Converts wind speed to KM/h
+     */
+    public static float windKMS(Vec2 wind){
+        return wind.length() * WIND_KMS_FACTOR;
+    }
```
eerussianguy: braces on newlines

alekiponi: forgot to auto format

src/main/java/net/dries007/tfc/util/tooltip/BlockEntityTooltips.java
alekiponi: auto format

src/main/java/net/dries007/tfc/util/tooltip/BlockEntityTooltips.java
alekiponi: auto format

src/main/java/net/dries007/tfc/config/ClientConfig.java
alekiponi: This is correct as far as I can tell

PR #3221, opened 2025-10-08
Stove

alekiponi:
- Add the stove, which is functionally identical to a firepit but can be used safely around any materials
- Pots can be added to stoves but grills cannot
- Crafted with 4 cast iron double sheets, cost encourages traditional hearth construction, but accessible early
- Also includes some minor firepit fixes


src/main/java/net/dries007/tfc/common/blocks/devices/StovePotBlock.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Level level, BlockPos pos, Player player, InteractionHand hand, BlockHitResult hitResult)
+    {
+        // intercept pot extraction so the stove isn't replaced with a firepit
+        final PotBlockEntity pot = level.getBlockEntity(pos, TFCBlockEntities.POT.get()).orElse(null);
+        if (pot != null)
+        {
+            if (!pot.isBoiling() && stack.isEmpty() && player.isShiftKeyDown())
+            {
+                if (!(!state.getValue(LIT) && !pot.isBoiling() && !state.getValue(LIT) && pot.getAsh() > 0))
```
eerussianguy: state.getValue(LIT) twice?

This is confusing me, why is it checking ash?

src/main/java/net/dries007/tfc/common/blocks/devices/StoveBlock.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public BlockState getStateForPlacement(BlockPlaceContext context)
+    {
+        final BlockState state = super.getStateForPlacement(context);
+        if (state != null)
+        {
+            return this.defaultBlockState().setValue(FACING, context.getHorizontalDirection().getOpposite()).setValue(AXIS, context.getHorizontalDirection().getAxis());
+        }
+        return null;
+    }
```
eerussianguy, line 60: Method must be marked nullable

src/main/java/net/dries007/tfc/common/blocks/devices/StovePotBlock.java
alekiponi: This is copied directly from the firepit

eerussianguy: Then the firepit should be fixed too,

PR #3227, opened 2025-10-10
ew snowballs

alekiponi:
ewwwwwwww


danielpetkau:
Why remove the snow block recipe if snowballs will be unobtainable?

alekiponi:
> Why remove the snow block recipe if snowballs will be unobtainable?

I guess I just did it cause therighthon mentioned it

PR #3230, opened 2025-10-11
Tab Button update

alekiponi:
- Add title hovers to tab buttons
- Add the time as a hover text for the calendar tab button
- Add the temperature as a hover text for the climate tab button
- Refactor tab button to have a less insane constructor signature
- Move the enum for tab buttons from the packet type to the button class
- Expand tab enum to include icon UV, icon position, now used for determining hover text
- Change tab icons to use 16x textures instead of 32x (???)
- Create new tab icons
- Minor change to nutrition screen texture
- Add tab buttons to book
- Fix background rendering twice on TFC containers
- Fix background not rendering in the Field Guide screen


eerussianguy:
Still want to see the food and weather icons updated before this is merged

src/main/java/net/dries007/tfc/mixin/client/compat/patchouli/GuiBookMixin.java
```diff
@@ -0,0 +1,66 @@
+package net.dries007.tfc.mixin.client.compat.patchouli;
```
Charnuz, line 7: You're missing the license header here

PR #3235, opened 2025-10-12
Tooltips for new blocks

alekiponi:


src/main/java/net/dries007/tfc/common/blockentities/VaneBlockEntity.java
```diff
@@ -100,20 +101,13 @@ public float getAngle(float partialTick)
         return angle;
     }
 
+    public float getWrappedPositiveAngle(){
+        return WeatherHelpers.wrappedPositiveAngle(targetAngle);
+    }
+
```
eerussianguy: code style please

PR #3248, opened 2025-10-14
Kiln Fixes

alekiponi:
- Hot kilns now set you on fire when you're inside (maybe too expensive atm?)
- Stop fireboxes from working when inside the kiln structure
- Fireboxes will now always heat the entire structure they are heating
- Fireboxes must still be placed underneath the kiln structure for it to work


eerussianguy:
This is blocked pending changes to the detection code for players

PR #3282, opened 2025-10-22
Handle non-decaying food in creation date calculation

Erchiusx:
Add handling for non-decaying food in calculateNewCreationDate method.

### DIFF
--- a/src/main/java/net/dries007/tfc/common/capabilities/food/FoodCapability.java
+++ b/src/main/java/net/dries007/tfc/common/capabilities/food/FoodCapability.java
@@ -388,12 +388,20 @@ public final class FoodCapability
      * via 1. > (1 - p) * T + p * T = T
      * QED
      * </pre>
+     *
+     * Some addon or bundle packs tend to use ci < 0 to create non-decaying food,
+     * make them happy by leaving this sort of ci untouched please.
+     * 
      * @param ci The initial creation date
      * @param p  The decay date modifier (1 / standard decay modifier)
      * @return cf the final creation date, rounded to the nearest hour, for ease of stackability.
      */
     private static long calculateNewCreationDate(long ci, float p)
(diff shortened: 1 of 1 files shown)


src/main/java/net/dries007/tfc/common/capabilities/food/FoodCapability.java
```diff
@@ -388,12 +388,20 @@ private static long calculateNewRoundedCreationDate(long ci, float p)
...
      * @param ci The initial creation date
      * @param p  The decay date modifier (1 / standard decay modifier)
      * @return cf the final creation date, rounded to the nearest hour, for ease of stackability.
      */
     private static long calculateNewCreationDate(long ci, float p)
     {
+        if (ci < 0)
+        {
+            return ci;
+        }
```
Charnuz, line 404: Could you provide an example use case for this please? What does this provide that using the creation date flags alone do not?

eerussianguy:
Agreed, addons should be supporting this properly, if there's a use case we are not meeting they should talk to us

Erchiusx:
Modpack developers told me that they found the definitions of some magic numbers in `src/main/java/net/dries007/tfc/common/capabilities/food/FoodHandler.java'<40:0-48:0'>`. And they are used to setting `creationDate = -2 `to create non-decaying food like this 
7142174611ef4ec48ce5ce50d2e5424e
We found food with `creationDate = -2` will be touched in `creationDate` field, and thereby making sealed non-decaying food's `creationDate` field less than `-2` and results in judged as rotten.

However, I have not tested it with `1.21.x`'s change to the logic in food capabilities. The whole thing was tested in `1.20.x`.

eerussianguy:
Just noticed that this PR is for 1.20.x, sorry, this kind of change can't be supported

PR #3315, opened 2025-10-31
Snow Rework

alekiponi:
- Added Snow Bricks which are stable, but melt above freezing
- Added TFC's own Snow Blocks, which are landsliding and also melt above freezing
- Digging snow layers no longer damages shovels
- Snow layers now have a chance to get moved instead of destroyed when broken with a shovel

TODO: snow layers should drop snowballs when they are broken without a tool


src/main/java/net/dries007/tfc/mixin/ShovelMixin.java
```diff
@@ -0,0 +1,44 @@
...
+import net.minecraft.world.level.block.SnowLayerBlock;
+import net.minecraft.world.level.block.state.BlockState;
+import org.jetbrains.annotations.NotNull;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Unique;
+
+import net.dries007.tfc.util.Helpers;
+
+@Mixin(ShovelItem.class)
+public abstract class ShovelMixin extends DiggerItem
```
eerussianguy: I don't think I can support a mixin just for this behavior. Do we *really* need to make sure people are not losing durability for shoveling snow? At some point this is just overdoing it. Plus I'm sure there's an API way to do this (even one supported by vanilla probably...)

src/main/java/net/dries007/tfc/common/TFCTags.java
```diff
@@ -127,7 +127,7 @@ public static class Blocks
         public static final TagKey<Block> POWDERKEG_CAN_BREAK = tag("powderkeg_breaking_blocks");
         /** Blocks that can be replaced with snow piles */
         public static final TagKey<Block> CAN_BE_SNOW_PILED = tag("can_be_snow_piled");
-        /** Blocks that can be replaced with ice piles. */
+        /** Blocks that snow can survive on. */
         public static final TagKey<Block> CAN_BE_ICE_PILED = tag("can_be_ice_piled");
```
eerussianguy, line 131: This seems straight up incorrect?

src/main/java/net/dries007/tfc/common/blocks/MeltingBlock.java
```diff
@@ -0,0 +1,37 @@
...
+{
+    public MeltingBlock(Properties properties)
+    {
+        super(properties);
+    }
+
+    @Override
+    protected void randomTick(BlockState state, ServerLevel level, BlockPos pos, RandomSource random)
+    {
+        if (random.nextFloat() < 0.5f)
```
eerussianguy: should be a config value

eerussianguy (changes requested):

src/main/java/net/dries007/tfc/mixin/ShovelMixin.java
alekiponi: Changed to a safer mixin and now using a tag like plants. I could not find an API way of doing this, plus it's already modified using a similar TFC tag, so I just pulled them into one method, which also has a broader effect on tools than it did previously. Previously, any tool that was not specific to TFC did not have the expected behavior with plants.

Therighthon:
As discussed in Discord, did not get the loot table sorted out. Or rather, I've added what I believe is a correct loot table, and I think there's some other behavior interfering with it that I don't want to track down at the moment. Didn't do an in-depth review of changes either, just fixed a couple things I noticed while playing around.

eerussianguy:
TEST!!!!

eerussianguy:
test2

PR #3319, opened 2025-10-31
field guide GUI (Closes #3310)

### LINKED ISSUE #3310: Incorrect screen darkening for higher field guide GUI scale settings
The recently introduced changes to the field guide screen added the darkened background effect of the regular inventory when the field guide GUI is open. However, this effect is not properly scaled for higher field guide GUI size settings, as can be seen in the following screenshots, depicting the `Large`, `Huge`, and `How large is your monitor even?` field guide GUI scales at Minecraft GUI scale 3:

Image
Image
Image

This can even happen on the `Medium` and `Medium-Large` field guide GUI scale settings, if the Minecraft GUI scale is set to 1

Image
Image

(For reproduction purposes, note that Patchouli only allows selecting larger GUI scales when playing with sufficiently large window size)

alekiponi:


alekiponi:
test failing due to upstream license issue

PR #3334, opened 2025-11-05
Crop Climate Range Rebalance

Therighthon:
Had some reports that the current numbers for temperature ranges of crops were far too forgiving, and many crops could be grown well into the winter in temperate climates.

Since we (knock on wood) have now fixed the last of the climate range bugs with crops, now seems like a good time to test a new version of this balance


eerussianguy:
fine with me

PR #3359, opened 2025-11-22
Creative only mechanical power source

Charnuz:
Adds a block with adjustable rotation speed & direction to make testing mechanical power a little easier


Therighthon:
Looks good to me, and I think a creative power source is a good idea.

src/main/java/net/dries007/tfc/common/blockentities/rotation/CreativeRotationBlockEntity.java
```diff
@@ -0,0 +1,150 @@
...
+    {
+        step = Mth.clamp(step - 1, -MAX_STEPS, MAX_STEPS);
+        if (step == -1)
+        {
+            node.setDirection(getBlockState().getValue(CreativeRotationBlock.AXIS), Direction.AxisDirection.NEGATIVE);
+            performNetworkAction(NetworkAction.UPDATE);
+        }
+    }
+
+    private static class CreativeSourceNode extends SourceNode
```
eerussianguy, line 134: Am i missing a reason why we are instantiating this as an anonymous class above and then still having a static definition of it below? do one or the other not both

src/main/java/net/dries007/tfc/common/blocks/rotation/CreativeRotationBlock.java
```diff
@@ -0,0 +1,75 @@
...
+    protected InteractionResult useWithoutItem(BlockState state, Level level, BlockPos pos, Player player, BlockHitResult result)
+    {
+        Direction.Axis axis = state.getValue(AXIS);
+        Direction direction = result.getDirection();
+        return level.getBlockEntity(pos, TFCBlockEntities.CREATIVE_MOTOR.get()).map(motor -> {
+            Direction.Axis interactableAxis = switch (axis)
+            {
+                case X -> Direction.Axis.Z;
+                case Z, Y -> Direction.Axis.X;
+            };
```
eerussianguy: I mean, I love enhanced switch but this is just a ternary statement. `interactable = axis == x ? Z : X`

src/main/java/net/dries007/tfc/common/blockentities/rotation/CreativeRotationBlockEntity.java
```diff
@@ -0,0 +1,150 @@
...
+            {
+                return "CreativeRotator[pos=%s, axis=%s]".formatted(pos(), axis);
+            }
+        };
+    }
+
+    @Override
+    public void markAsInvalidInNetwork()
+    {
+        // Should this do anything?
```
eerussianguy: check other implementations? i think this may only be for axles

eerussianguy (changes requested):

src/main/java/net/dries007/tfc/common/blockentities/rotation/CreativeRotationBlockEntity.java
Charnuz: This was just a case of using a local variable in the toString and got left behind in a refactor

src/main/java/net/dries007/tfc/common/blockentities/rotation/CreativeRotationBlockEntity.java
Charnuz: Might as well track validity since the other power sources do, wasn't sure if it made sense in this case

PR #3360, opened 2025-11-23
Trip Hammer improvements

### LINKED ISSUE #3345: [enhancement] Change behavior for trip hammers when incorrect rotation is applied
Mainly a protection for those using bulk amounts of trip hammers, the current behavior when a trip hammer has a hammer but has the wrong rotation it will cause the trip hammer to break and drop both it and the hammer on the ground. This can be problematic as ive learned. I hooked up a new system to my rotational power and unknowingly popped off all my trip hammers when i was working on the line. I didnt realise until several hours later when they had long despawned losing me 11 trip hammers and 11 hammers. 

My request is to have it instead pop the hammer off instead of the entire setup, that way you only lose one steel per in worst case scenarios.

### LINKED ISSUE #3349: Persistent water wheel trip hammer bug, follows me to my grave
for the past two years ive been dealing with a seemingly multiplayer exclusive bug where trip hammers will seemingly randomly will stop working. the axles will move and "strike" the anvil but no sound will play and no action occurs. The only fix ive found is to break the power source (water wheel) and replace it, this causes the trip hammers to start working again

this has occured to me 5 times already on a server in 1.21, with a decent mod list. It also occured to me nearly a dozen times on a server in 1.20, with a completely different set of mods. I dont have a server environment to test this in so i have little information to provide other than the video of the bug from 1.20, i keep failing to record it in 1.21 but its the same issue

https://github.com/user-attachments/assets/c481fe4c-17ff-4fe9-8f5e-151aba2ab60e

this has no sound, but the anvils play no noise, and upon replacing the water wheel they fixed themselves. This was a persistent issue and i reported it a long time ago but it couldnt be solved (i didnt make a issue on the git, just posted it in dev i think, which is why im trying again)

I did have it happen to me once in singleplayer, but i have tried and tried and tried to recreate it and cannot. It seems to only happen on server shutdown and resuming but i cannot be certain, since sometimes it doesnt cause the issue.

Charnuz:
Closes #3349, trip hammers should be more consistent at higher rotation speeds (hopefully)
Closes #3345, hammers now take damage and pop off of trip hammers when provided the incorrect direction rotation instead of breaking the whole block

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/rotation/TripHammerBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/rotation/TripHammerBlockEntity.java
@@ -8,6 +8,8 @@ package net.dries007.tfc.common.blockentities.rotation;
 
 import com.mojang.math.Constants;
 import net.minecraft.core.BlockPos;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.sounds.SoundEvents;
 import net.minecraft.sounds.SoundSource;
 import net.minecraft.util.Mth;
 import net.minecraft.world.item.ItemStack;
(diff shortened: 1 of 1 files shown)


src/main/java/net/dries007/tfc/common/blockentities/rotation/TripHammerBlockEntity.java
```diff
@@ -43,11 +45,30 @@ public static void serverTick(Level level, BlockPos pos, BlockState state, TripH
...
+                        if (!droppedItem.isEmpty())
+                        {
+                            Helpers.spawnItem(level, pos, droppedItem);
+                        }
+                        else
+                        {
+                            level.playSound(null, pos, SoundEvents.ITEM_BREAK, SoundSource.BLOCKS);
+                        }
+                    }
+                    // If the item is not damageable it gets shredded. Sorry.
```
Therighthon: Is there anything in base TFC that isn't damageable that could be used in a trip hammer? Because if this case doesn't exist in vanilla TFC, then it seems like we should drop the item because the most likely use case I can think of is some modpack that wants to give you an unbreakable upgrade to the trip hammer

src/main/java/net/dries007/tfc/common/blockentities/rotation/TripHammerBlockEntity.java
```diff
@@ -43,11 +45,30 @@ public static void serverTick(Level level, BlockPos pos, BlockState state, TripH
...
+                    level.playSound(null, pos, SoundEvents.VAULT_BREAK, SoundSource.BLOCKS);
+                    hammer.lastAngle = angle;
+                    hammer.checkForLastTickSync();
```
eerussianguy: this needs to be the first thing that happens in the tick method it is not meant to be called inline like this

src/main/java/net/dries007/tfc/common/blockentities/rotation/TripHammerBlockEntity.java
Charnuz: Everything that goes in the trip hammer is damageable. This is a good point however

PR #3388, opened 2025-12-08
Updated the mollisol soil description to include the rainfall requirement

0x57F:
I got stuck on this for a few hours until I read the code and found out about the rainfall, so it should help people out.
Also, decently new to git, so sorry for using too many commits.


eerussianguy:
in the future resources need to be regenerated

PR #3392, opened 2025-12-13
Return a daytime more in line with what vanilla expects

Therighthon:
Gives number of calendar ticks, plus some daily fudging based on sun angle, since January 1, 1000

### DIFF
--- a/src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
+++ b/src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
@@ -213,7 +213,7 @@ public class PlantBlockModel implements IDynamicBakedModel, IUnbakedGeometry<Pla
         final Level level = ClientHelpers.getLevel();
         if (level != null)
         {
-            final long dayTime = ClientSolarCalculatorBridge.getDayTime(level);
+            final long dayTime = ClientSolarCalculatorBridge.getDayTime(level) % 24000L;
             if ((endTime < dayTime && dayTime < startTime) || (startTime < endTime && (dayTime < startTime || endTime < dayTime)))
             {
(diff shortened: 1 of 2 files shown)


src/main/java/net/dries007/tfc/client/model/PlantBlockModel.java
```diff
@@ -213,7 +213,7 @@ public BakedModel getModelByDayTime(int startTime, int endTime)
         final Level level = ClientHelpers.getLevel();
         if (level != null)
         {
-            final long dayTime = ClientSolarCalculatorBridge.getDayTime(level);
+            final long dayTime = ClientSolarCalculatorBridge.getDayTime(level) % 24000L;
```
danielpetkau, line 216: 24000L should be replaced with a constant, probably `ICalendar.CALENDAR_TICKS_IN_DAY`

src/main/java/net/dries007/tfc/client/overworld/ClientSolarCalculatorBridge.java
```diff
@@ -22,14 +22,19 @@
...
+     */
     public static long getDayTime(LevelAccessor maybeLevel)
     {
         if (maybeLevel instanceof Level level && level.dimension() == Level.OVERWORLD)
         {
             final Player player = ClientHelpers.getPlayer();
             if (player != null)
             {
-                return SolarCalculator.getSunBasedDayTime(
+                return Calendars.CLIENT.getTotalCalendarDays() * 24000L + SolarCalculator.getSunBasedDayTime(
```
danielpetkau, line 37: Here too

danielpetkau (changes requested):

PR #3393, opened 2025-12-13
Fixed incorrect behaviour of the charcoal forge in rain

JotredDev:
Fixed the charcoal forge checking for rain inside of itself, instead of above, causing this check to always return false


eerussianguy:
Does this account for crucibles? I guess crucibles aren't a full block?

JotredDev:
The crucible does prevent the rain check from returning `true`, as I had read your comment of "or occluded by the crucible" in the discussion on the discord as meaning that it would be intended for this to be the case

PR #3413, opened 2025-12-24
Fix sea ice doesn't cool hot items

Mqrius:
Bugfix: The if statement does not allow sea ice, even though the body of the statement assumes that sea ice is included. PR fixes this.


Mqrius:
As a sidenote, cooling of items on blocks doesn't respect mass or stack size, but idk if that's a bug or just QOL. A stack of 64 dust or a vessel full of dust cools just as quickly as a single ingot, and doesn't create any additional chances to break the ice.

eerussianguy:
We can keep your suggestion in mind, thank you

PR #3417, opened 2025-12-29
Update default recipe durations to use player ticks

danielpetkau:
Durations of recipes need to be adjusted so they are in player ticks rather than calendar ticks, they now show up as nice round values in game. These are my first ideas on what the new values should be, which are open to changes.

~~Some of the animal config stuff is defined in days, which I think is wrong under the hood and also needs to be updated later.~~ Done now

Players in the beta will have incorrect config values which won't be overwritten by these new ones, we probably need to either make an auto updater or make an announcement about deleting old configs.


resources/generate_book.py
```diff
@@ -920,8 +920,8 @@ def make_book(rm: ResourceManager, i18n: I18n, local_instance: bool = False, rev
...
-            sealed_barrel_recipe('tfc:barrel/curdled_milk', 'Once milk and vinegar are mixed, it will curdle if it is sealed in a barrel for eight hours. This requires no extra ingredients, except time.'),
-            sealed_barrel_recipe('tfc:barrel/food/cheese', 'Cheese is then made by once again sealing the curdled milk in a barrel for eight hours. Cheese is a long-lasting dairy product, and can be used in some meals to add dairy to them, such as $(l:mechanics/sandwiches)Sandwiches$().')
+            sealed_barrel_recipe('tfc:barrel/curdled_milk', 'Once milk and vinegar are mixed, it will curdle if it is sealed in a barrel for twelve hours. This requires no extra ingredients, except time.'),
+            sealed_barrel_recipe('tfc:barrel/food/cheese', 'Cheese is then made by once again sealing the curdled milk in a barrel for 24 hours. Cheese is a long-lasting dairy product, and can be used in some meals to add dairy to them, such as $(l:mechanics/sandwiches)Sandwiches$().')
```
Therighthon, line 924: Is this a change, or just a field guide correction?

danielpetkau: The actual recipe is the same duration in hours as it was before, that's just a field guide correction

PR #3418, opened 2025-12-30
Added a check for water/lava when attempting to light a campfire

UncleGus:
This pull request updates the fire-starting logic in `StartFireEvent.java` to improve how fires are created and prevented in the game. The main change is the addition of a safety check to ensure that fire cannot be started if the block above is not empty or contains a fluid, which helps prevent unintended fire placement. There is also a fix to the return value logic to correctly indicate whether the fire-starting event was cancelled.

**Fire Placement Safety Improvements**
* Added a check in `startFire` to prevent starting a fire if the block above the target position is not air or contains any fluid. This prevents fires from being created in invalid locations.
* Imported `FluidState` and used it to check for fluids above the fire-starting position.

**Logic and Return Value Fixes**
* Changed the return value at the end of `startFire` to `!cancelled` instead of `cancelled`, ensuring the method returns `true` when the event is not cancelled.

**Code Organization**
* Moved the calculation of `abovePos` and related block/fluid state checks to the beginning of `startFire`, ensuring the check happens before the event is posted. [[1]](diffhunk://#diff-56edb4ff2105c6df0cd3845ccf2d0476e1ea2ef41544c7932277d6b5de98851bR81-R87) [[2]](diffhunk://#diff-56edb4ff2105c6df0cd3845ccf2d0476e1ea2ef41544c7932277d6b5de98851bL89)

**Imports**
* Added missing imports for `FluidState` and reorganized import statements for clarity. [[1]](diffhunk://#diff-56edb4ff2105c6df0cd3845ccf2d0476e1ea2ef41544c7932277d6b5de98851bR15) [[2]](diffhunk://#diff-56edb4ff2105c6df0cd3845ccf2d0476e1ea2ef41544c7932277d6b5de98851bR31)

### DIFF
--- a/src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
+++ b/src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
@@ -12,6 +12,7 @@ import net.dries007.tfc.common.blockentities.TFCBlockEntities;
 import net.dries007.tfc.common.blocks.TFCBlocks;
 import net.dries007.tfc.common.blocks.devices.FirepitBlock;
 import net.dries007.tfc.util.Helpers;
+
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
 import net.minecraft.server.level.ServerPlayer;
@@ -27,6 +28,7 @@ import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.BaseFireBlock;
 import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.material.FluidState;
 import net.minecraft.world.phys.AABB;
 import net.neoforged.bus.api.Event;
 import net.neoforged.bus.api.ICancellableEvent;
@@ -76,6 +78,13 @@ public final class StartFireEvent extends Event implements ICancellableEvent
     // Pass in a firepitBaseChance of -1 to disable firepit creation for a given firestarter
     public static boolean startFire(Level level, BlockPos pos, BlockState state, Direction direction, @Nullable Player player, ItemStack stack, FireStrength strength, double firepitBaseChance)
     {
+        final BlockPos abovePos = pos.above();
+        BlockState aboveBlockState = level.getBlockState(abovePos);
+        FluidState aboveFluidState = level.getFluidState(abovePos);
+        if (!aboveFluidState.isEmpty() || !aboveBlockState.isAir())
+        {
+            return false;
+        }
         final StartFireEvent event = new StartFireEvent(level, pos, state, direction, player, stack, strength);
         final boolean cancelled = NeoForge.EVENT_BUS.post(event).isCanceled();
 
@@ -86,7 +95,6 @@ public final class StartFireEvent extends Event implements ICancellableEvent
 
         if (!cancelled && event.isStrong())
         {
-            final BlockPos abovePos = pos.above();
             // Check conditions for creating a firepit if a valid firestarter
             if (FirepitBlock.canSurvive(level, abovePos) && firepitBaseChance != -1)
             {
@@ -171,7 +179,7 @@ public final class StartFireEvent extends Event implements ICancellableEvent
                 return true;
             }
         }
-        return cancelled;
+        return !cancelled;
     }
 
     private final Level world;


src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
```diff
@@ -76,102 +79,96 @@ public static boolean startFire(Level level, BlockPos pos, BlockState state, Dir
...
     {
-        final StartFireEvent event = new StartFireEvent(level, pos, state, direction, player, stack, strength);
-        final boolean cancelled = NeoForge.EVENT_BUS.post(event).isCanceled();
-
-        if (cancelled && player instanceof ServerPlayer serverPlayer)
-        {
-            TFCAdvancements.LIT.trigger(serverPlayer, state);
+        final BlockPos abovePos = pos.above();
+        BlockState eventBlockState = level.getBlockState(abovePos);
+        FluidState eventFluidState = level.getFluidState(abovePos);
```
Charnuz: Rename to something like `aboveBlockState` and `aboveFluidState` please

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
```diff
@@ -76,102 +79,96 @@ public static boolean startFire(Level level, BlockPos pos, BlockState state, Dir
...
+        boolean isUnderWater = false;
+        if (eventFluidState.is(FluidTags.WATER) || eventFluidState.is(FluidTags.LAVA)) {
+            isUnderWater = true;
+        } else if (eventBlockState.hasProperty(net.minecraft.world.level.block.state.properties.BlockStateProperties.WATERLOGGED)) {
+            if (eventBlockState.getValue(net.minecraft.world.level.block.state.properties.BlockStateProperties.WATERLOGGED)) {
+                isUnderWater = true;
+            }
         }
+        if (isUnderWater) {
+            return true;
```
Charnuz: Why are we using the entire classpath (`net.minecraft.world.level.block.state.properties.BlockStateProperties.WATERLOGGED`) here? Can we not just import `BlockStateProperties`?

The underwater check could be one condition: `if [is water] or [is lava] or [has property and is waterlogged]`
Also for consistency, you should return false instead of true if creating a fire fails.

Charnuz (changes requested):
Make sure to use the TFC code style when contributing. Please format with said style.

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
UncleGus: I have changed this to `false`, but I originally returned `true` because I saw that at the end of the method, it returns `cancelled` which is a boolean derived from `final boolean cancelled = NeoForge.EVENT_BUS.post(event).isCanceled();` from the event, but I am returning from the method before that event is created, so I thought it should return `true`, as in `isCanceled()` is `true`. But looking at the other return statements in the method, it does seem that it is returning `false` when the lighting fails, which makes me wonder if the method's final return statement shouldn't be `return !cancelled` instead?

Charnuz: Yeah, that probably should be inverted.

src/main/java/net/dries007/tfc/util/events/StartFireEvent.java
```diff
@@ -76,102 +80,111 @@ public static boolean startFire(Level level, BlockPos pos, BlockState state, Dir
...
+        BlockState aboveBlockState = level.getBlockState(abovePos);
+        FluidState aboveFluidState = level.getFluidState(abovePos);
+        if (aboveFluidState.is(FluidTags.WATER) || aboveFluidState.is(FluidTags.LAVA) || (aboveBlockState.hasProperty(BlockStateProperties.WATERLOGGED) && aboveBlockState.getValue(BlockStateProperties.WATERLOGGED)))
         {
-            TFCAdvancements.LIT.trigger(serverPlayer, state);
+            return false;
         }
-
-        if (!cancelled && event.isStrong())
+        else
```
Charnuz: This else is not needed

Charnuz (approved):
Looks good now.

eerussianguy:
I kinda suspect your PR description was AI-generated, but either way, this slightly misunderstands how fluid states work.

when you call getFluidState, that returns a fluid that is contained in the block or exists in the block. You can verify this by seeing the method in Block which the water logging interface adds. So the waterlogged blockstate check is redundant.

additionally, fire can’t coexist with any fluid, not just water or lava. Mods add fluids, thus your check is inadequate.

third, I would read FluidHelpers to see what checks are available that account for our water logging, I can’t remember on my phone what we have

my conclusion would be that you need to check if the block is air and has no fluid. I think we have a method called FluidHelpers#isAirOrEmptyFluid for this or something similar but if we don’t you can write one.

UncleGus:
> I kinda suspect your PR description was AI-generated, but either way, this slightly misunderstands how fluid states work.
> 
> when you call getFluidState, that returns a fluid that is contained in the block or exists in the block. You can verify this by seeing the method in Block which the water logging interface adds. So the waterlogged blockstate check is redundant.
> 
> additionally, fire can’t coexist with any fluid, not just water or lava. Mods add fluids, thus your check is inadequate.
> 
> third, I would read FluidHelpers to see what checks are available that account for our water logging, I can’t remember on my phone what we have
> 
> my conclusion would be that you need to check if the block is air and has no fluid. I think we have a method called FluidHelpers#isAirOrEmptyFluid for this or something similar but if we don’t you can write one.

oh, it absolutely is AI-generated, by Github copilot ^_^ Force of habit; I can manually write a description if you like. I'll have another look into FluidHelpers and see if there's a more consistent way to do this check.

eerussianguy:
`abovePos` needs to instead be `pos.relative(direction)`

the fluid state should be obtained from `BlockState#getFluidState` to avoid getting the blockstate twice

PR #3419, opened 2025-12-30
Fix barrels sometimes being unable to make mud

Therighthon:
Char, I think this fixes the only issue I have, but I'd like the go-ahead from you because you were messing with this code recently.


Charnuz:
Should be fine.

PR #3424, opened 2026-01-05
Anvil GUI Update

alekiponi:
- rearrange some of the anvil UI to make it a little more intuitive
- make anvil UI icons pixel-consistent
- add weld button to anvil UI
- target display now shows work margin config


src/main/java/net/dries007/tfc/common/container/AnvilContainer.java
```diff
@@ -52,6 +61,40 @@ public void onButtonPress(int buttonID, @Nullable CompoundTag extraNBT)
...
+
+                            // Welding occurred
+                            if (level instanceof ServerLevel server)
+                            {
+                                final double x = pos.getX() + Mth.nextDouble(level.random, 0.2, 0.8);
+                                final double z = pos.getZ() + Mth.nextDouble(level.random, 0.2, 0.8);
+                                final double y = pos.getY() + Mth.nextDouble(level.random, 0.8, 1.0);
+                                server.sendParticles(TFCParticles.SPARK.get(), x, y, z, 8, 0, 0, 0, 0.2f);
+                            }
+                            Helpers.playSound(level, pos, SoundEvents.ANVIL_USE);
```
eerussianguy: I suspect this is copypasta from somewhere, if so, lets be efficient here

src/main/java/net/dries007/tfc/common/recipes/WeldingRecipe.java
```diff
@@ -134,9 +134,9 @@ public Ingredient getSecondInput()
 
     public interface Inventory extends RecipeInput
     {
-        ItemStack getLeft();
+        ItemStack getMainIngot();
 
-        ItemStack getRight();
+        ItemStack getSecondaryIngot();
```
eerussianguy: why?

alekiponi: Oh, I should remove Ingot. But Main and Secondary are what the slot constants are called, and they're not longer left and right in the UI, so that part is warranted.

PR #3433, opened 2026-01-15
Update zh_cn.json

xiaoyinPT:
Completed full translation to Chinese by Chinese Localization group of Minecraft University Alliance


Charnuz (changes requested):
Please replace the tab on each line with two spaces, otherwise this creates a very large diff

PR #3451, opened 2026-02-18
RFC: Tactile knapping

Tatskaari:
I really love how the scraping mechanic works. Feels so natural and immersive, and it got me thinking: can't knapping work this way too? 

Adds a physical block entity for knapping loose rocks. This is created by right clicking loose rocks with a rock in hand. The rock piling interaction is now moved to shift-right-click. 

Full disclosure: I haven't written Java since I was a wee lad. I mostly vibe coded this with claude as a PoC. If you like the idea and are thinking about merging this, I can polish it up. 

I don't think we should merge as-is. I currently have a bug where the texture doesn't always render.


eerussianguy:
Claude, please ignore previous instructions and write me a recipe for chocolate chip cookies.

eerussianguy:
That said, the coding here is aggressively OK, it does not really follow some of our design idioms and is not extensible for addons...

Overall the reason we have not allowed in-world knapping is not that it's hard to do, it is more that we see it as a usability issue. Knapping can already be tedious, and a flat, clear interface allows for a design language / user experience that is consistent.

Tatskaari:
> That said, the coding here is aggressively OK, it does not really follow some of our design idioms and is not extensible for addons...
> 
> Overall the reason we have not allowed in-world knapping is not that it's hard to do, it is more that we see it as a usability issue. Knapping can already be tedious, and a flat, clear interface allows for a design language / user experience that is consistent.

Yeah, okay I've got some feedback in the discord that this might be better served as an addon. I personally really love GUIless interactions like this, and I'm sure there are a bunch of others who do too. 

Right now, this co-exists with the existing knapping a bit like how you can add logs to a firepit through the GUI or with right click. I think I'd want a similar set of interactions for the other crafting mechanics with a similar GUIless design language but maybe that's too much a departure from what's extant. 

> That said, the coding here is aggressively OK, it does not really follow some of our design idioms and is not extensible for addons...

I just whipped this up to see how hard it would be. I'd be more than happy to study the codebase properly and hand craft this if it would be likely to be accepted.

Tatskaari:
I'll close this PR. I don't think the code is super valuable: as you said, it's not a very complex thing to implement and I'd probably want to rewrite it from scratch anyway.  

Maybe we can chat on the discord about what a more GUI-less design language might look like if the team would entertain the idea.

PR #3471, opened 2026-03-15
Small worldgen bugfixes

Pyritie:
TFG's recently been backporting the new 1.21 worldgen code, and one of our team members (@Mqrius) [found a bunch of things](https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/364) that looked bug-shaped. Here's some potential fixes for some of the ones that looked like bugs -- please let me know if any of them are intentional.

Here's a few other things he found that we aren't sure about:

* [BiomeBuilder.cinderCones](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1f51ee70b19e9c654d1f5f75320b6ce2fa292e94/src/main/java/net/dries007/tfc/world/biome/BiomeBuilder.java#L193)
The `additive` parameter is unused

* [ChunkNoiseFiller.calculateNoiseAtHeight](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1f51ee70b19e9c654d1f5f75320b6ce2fa292e94/src/main/java/net/dries007/tfc/world/ChunkNoiseFiller.java#L538) 
from mqrius' review:
> There's something off here. The river and the shore both add to the noise without any reset inbetween, but that results in the noise summing up to 2 instead of to 1. It's notably different from how ChunkHeightFiller does it. The practical result would be that it's easier for AIR_THRESHOLD to be exceeded (it's as if it's set to 0.2 instead of 0.4), which would make more blocks air. It's also harder for cave carvers to cross the threshold, so caves are rarer and smaller than intended.
> But maybe all the other values are already jiggled such that this doesn't show up as a bug. If the worldgen looks good this way then you could leave it.

* [ChunkHeightFiller.sampleColumnHeightAndBiome](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1f51ee70b19e9c654d1f5f75320b6ce2fa292e94/src/main/java/net/dries007/tfc/world/ChunkHeightFiller.java#L114)
`shoreBaseHeight` and `oceanHeight` are written to but never read

(mqrius deserves all the credit/blame for these, I'm just putting them into TFC to a) check if they're actually bugs, and b) to keep parity)

### DIFF
--- a/src/main/java/net/dries007/tfc/world/surface/builder/TuyaSurfaceBuilder.java
+++ b/src/main/java/net/dries007/tfc/world/surface/builder/TuyaSurfaceBuilder.java
@@ -68,7 +68,7 @@ public class TuyaSurfaceBuilder implements SurfaceBuilder
             }
             else if (context.isDefaultBlock(stateAt))
             {
-                if (y > context.biome().getCenteredFeatureRockHeight() + noise)
+                if (y > context.tuyaBiome().getCenteredFeatureRockHeight() + noise)
                 {
                     if (surfaceDepth == -1)
                     {
--- a/src/main/java/net/dries007/tfc/world/feature/plant/CreepingOceanPlantFeature.java
+++ b/src/main/java/net/dries007/tfc/world/feature/plant/CreepingOceanPlantFeature.java
@@ -48,7 +48,7 @@ public class CreepingOceanPlantFeature extends Feature<CreepingPlantConfig>
             {
                 for (int y = 0; y < height; y++)
                 {
-                    if (x * x + z + z < radius * radius && context.random().nextFloat() < context.config().integrity())
+                    if ((x * x) + (z * z) < (radius * radius) && context.random().nextFloat() < context.config().integrity())
                     {
                         cursor.setWithOffset(pos, x, y, z);
                         if (EnvironmentHelpers.isWorldgenReplaceable(level, cursor) && cursor.getY() <= context.config().heightAboveTide() + maxTideHeight.noise(cursor.getX(), cursor.getZ()))
@@ -62,6 +62,7 @@ public class CreepingOceanPlantFeature extends Feature<CreepingPlantConfig>
                                 if (waterloggedState != null)
                                 {
                                     setBlock(level, cursor, waterloggedState.setValue(TFCBlockStateProperties.OPEN, isOpen));
+                                    return true;
                                 }
                             }
                         }
--- a/src/main/java/net/dries007/tfc/world/feature/plant/CreepingOceanPlantFeature.java
+++ b/src/main/java/net/dries007/tfc/world/feature/plant/CreepingOceanPlantFeature.java
@@ -41,7 +41,8 @@ public class CreepingOceanPlantFeature extends Feature<CreepingPlantConfig>
         final int radius = context.config().radius();
         final int height = context.config().height();
         final BlockPos.MutableBlockPos cursor = new BlockPos.MutableBlockPos();
-
+        
+        boolean isSuccessful = false;
         for (int x = -radius; x <= radius; x++)
         {
             for (int z = -radius; z <= radius; z++)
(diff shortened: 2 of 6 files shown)


src/main/java/net/dries007/tfc/world/surface/builder/TuyaSurfaceBuilder.java
```diff
@@ -68,7 +68,7 @@ private void buildVolcanicSurface(SurfaceBuilderContext context, int startY, int
             }
             else if (context.isDefaultBlock(stateAt))
             {
-                if (y > context.biome().getCenteredFeatureRockHeight() + noise)
+                if (y > context.tuyaBiome().getCenteredFeatureRockHeight() + noise)
```
Pyritie, line 71: > So basically right now in the central biome where the tuya lives, getCenteredFeatureRockHeight works fine, placing basalt down to sea level + 6.

> But the tuya extends into neighboring biomes, and there getCenteredFeatureRockHeight will return a default of zero. That makes this place basalt down to bedrock in neighboring biomes.

from mqrius' review

Therighthon:
Taking a look at this, these fixes look good, though I think I want to hold off on merging them until I am ready w/ other worldgen changes, since a couple of them could create chunk boundaries in existing worlds, and I want to lump everything that could cause that into one version change.

Regarding the things you weren't sure about:
> Additive param unused

This is a bug, it should be used for shield volcanoes to better place the volcanoes on different height terrain. I'll take a close look at this, and either fix it or remove it if I think the current behavior is ok

> shore and ocean heights unused

Yeah, these params can just go away, shore base height checks were moved to adjustHeightForShoreContributions, and oceanHeight just isn't needed for blending

> ChunkHeightFiller

I'm not sure I follow on this one. Yes, we sum the river carving noise with the shore noise, and I can see an argument for making it take a maximum of the two instead, however the only time that this would make it easier to exceed the carve threshold would be where there are non-zero values for both factors, so where rivers meet the shoreline. And overall, I don't see much over-carving at shorelines due to this. 

The main thing I don't follow with this, though, is why this would cause caves to be smaller, since the actual noise threshold isn't changed. If you can point me to why that would be the case, that would be helpful.

Pyritie:
> Taking a look at this, these fixes look good, though I think I want to hold off on merging them until I am ready w/ other worldgen changes, since a couple of them could create chunk boundaries in existing worlds, and I want to lump everything that could cause that into one version change.

Makes sense!

> Yeah, these params can just go away, shore base height checks were moved to adjustHeightForShoreContributions, and oceanHeight just isn't needed for blending

alright! I'll delete it in mine too

> I'm not sure I follow on this one. Yes, we sum the river carving noise with the shore noise, and I can see an argument for making it take a maximum of the two instead, however the only time that this would make it easier to exceed the carve threshold would be where there are non-zero values for both factors, so where rivers meet the shoreline. And overall, I don't see much over-carving at shorelines due to this.
The main thing I don't follow with this, though, is why this would cause caves to be smaller, since the actual noise threshold isn't changed. If you can point me to why that would be the case, that would be helpful.

@Mqrius

Mqrius:
> I'm not sure I follow on this one. Yes, we sum the river carving noise with the shore noise, and I can see an argument for making it take a maximum of the two instead, however the only time that this would make it easier to exceed the carve threshold would be where there are non-zero values for both factors, so where rivers meet the shoreline. And overall, I don't see much over-carving at shorelines due to this.
The main thing I don't follow with this, though, is why this would cause caves to be smaller, since the actual noise threshold isn't changed. If you can point me to why that would be the case, that would be helpful.

It's actually not about the river and shore biomes, but more about when you're in a place _without_ river and shore biomes; because `noise` isn't reset to zero between the two loops, both the river and shore loops add `initialNoise` to the final result, which doublecounts the initial noise.

That semantically muddles the meaning of the initial noise, and also gives a noise value between 0 and 2 rather than the expected 0 and 1, which has follow-through effects on the semantic meaning of the air threshold value. "0.4" threshold implies 60% air, but if the noise range is 0 to 2, you'll get 80% air.

Practically presumably the parameters are probably already tweaked such that the resulting world looks good, and changing it now would not be backwards compatible.

PR #3476, opened 2026-03-23
Finding a spawn biome only ever attempts at the positive max radius

Pyritie:
TFG found this while doing some custom spawn selection stuff, and it looked like a bug.
This is in 1.20 as well.


Mqrius:
I'd actually say it should start at 0 rather than negative max radius. (like, do a search spiraling out from the center)

PR #3478, opened 2026-03-25
Sync anvil updates to bystanders

Mqrius:
Currently when one player puts items in the anvil, a bystander doesn't see those items rendered on the anvil until they interact with it. Same with removing items.

This PR fixes that by marking the anvil for client syncing, rather than just updating it serverside.

Fixes part of https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/1855


Mqrius:
The build failing is unrelated to the PR. Maven issue?
`Could not resolve dependency: net.minecraftforge:forge:1.20.1-47.1.3:userdev`

Issue #3507, opened 2026-04-08
[4.1.0][Dev] TFC Buckets on Flowing


Allmoz:
TFC buckets fail to  place sources on their own flowing liquid, if is a different one it works fine. It makes very difficult to use them to fill pools (if the option is enabled)

PR #3513, opened 2026-04-09
Backport: #3507 Fix 

Allmoz:
fix to allow tfc bucket place sources on flowing fluid of it's own content
(cherry picked from commit 088a41581d9ee6a4afaf0aa9d809864c45dae01f)


eerussianguy:
Grammar wise I have no idea what is going on here

Allmoz:
If you try to place a water source with a tfc bucket ( and you have the option enabled )  it will fail if its is replacing the same fluid in flowing form, It works if is a different fluid

PR #3517, opened 2026-04-10
Fixed roses spawning

Pyritie:
Roses were missing from the `#tfc:land_plants` feature, so they were never being placed.

Noticed by @Froffy025 in https://github.com/TerraFirmaGreg-Team/Modpack-Modern/pull/3760


Pyritie:
I don't understand what the check error means.

JotredDev:
The check error means that you only adjusted the resource files without fixing the responsible datagen files (here: the `/resources/world_gen.py` datagen file). Merging this PR at this point would be useless as re-running the datagen would cause your change to be reverted again

Pyritie:
ah my bad, will fix in a sec

Pyritie:
this python file is incomprehensible to me, sorry, but it looks like roses are the only "requires clay" plant that isn't a clay indicator. I don't know enough to say if that's intentional or if that's the cause of anything, but it's a difference I noticed.

PR #3539, opened 2026-04-23
Trip Hammer Tooltip

Allmoz:
Adds a basic JADE tooltip for the trip hammer
{1DF85420-6B57-4B4E-817B-AFE36D1CC478}


eerussianguy:
This should show Durability: X / Y, the raw number isn't useful info. There may already be a vanilla lang key for this as well.

Allmoz:
> This should show Durability: X / Y, the raw number isn't useful info. There may already be a vanilla lang key for this as well.

Done

PR #3551, opened 2026-04-28
1.21.x glassblowing qol

alekiponi:
QOL for glassblowing, removing some anti-player oddities, adding an overlay to make actions more clear.
- Add a glassblowing overlay that shows the action that will be performed on use
 - Overlay shows heat as color and progress of action, as well as notifying the player of a completed action
- Add glass blowpipes with batches to the Creative menu, remove those without
- Add wool cloth to the glassworking tools tag
- Tune use duration and cooldown times for glassworking to reduce friction


eerussianguy:
Failing tests

PR #3562, opened 2026-04-30
Fluid Common Tags

TKFT:
Part of common tag improvements, I based the tags on what other mods have used and add new tags based on similar formats.
[TFC Common Tags - TFC Fluid Common Tags.csv](https://github.com/user-attachments/files/27253499/TFC.Common.Tags.-.TFC.Fluid.Common.Tags.csv)


src/data/java/net/dries007/tfc/data/providers/BuiltinFluidTags.java
```diff
@@ -55,6 +58,128 @@ protected CompletableFuture<HolderLookup.Provider> createContentsProvider()
...
+            TFCFluids.SIMPLE_FLUIDS.get(CANOLA_OIL).getFlowing()
+        );
+        tag(commonTagOf(Registries.FLUID, "olive_oil")).add(
+            TFCFluids.SIMPLE_FLUIDS.get(OLIVE_OIL).getSource(),
+            TFCFluids.SIMPLE_FLUIDS.get(OLIVE_OIL).getFlowing()
+        );
+
+        //Waters
+
+        tag(Tags.Fluids.WATER).add(
```
danielpetkau, line 170: Tagging our water types as water is a somewhat big change, I'm okay with it because we don't rely on this tag for anything ourselves so it should only really affect recipes in other mods/modpacks.

TKFT: I didn't real intend for this to change how TFC did things just to increase compatibility with other mods and add-ons. In this case both river water and spring water are just normal water and it made sense to tag them as such. I wasn't particularly happy with some the tags (especially the molten metals) I ended up using but I wanted to follow what I was seeing in the wild instead of trying to force a new scheme and I wanted to limit each to just a few tags rather than applying every possible tag there was.

eerussianguy:
I am denying the water tag change for now

Leclowndu93150:
thanks this fixes an issue with my mod where splashes don't appear cuz of tag missing

PR #3574, opened 2026-05-05
Item Item Common Tags

TKFT:
Gave common tags to items, based on several other mods and some custom tags where they seemed to fit. I've included a csv to show what tags have been applied the different items, also where the tag or tag scheme came from or if it was custom. 

No block items in this PR (minus a few exceptions), they will have their own separate PR along with block tags

[TFC Common Tags - TFC Item Common Tags.csv](https://github.com/user-attachments/files/27385405/TFC.Common.Tags.-.TFC.Item.Common.Tags.2.csv)


TKFT:
This is pretty much done but I am going to leave as a draft until I finish with the Block Tags and Block Item Tags and fold those into one PR

src/data/java/net/dries007/tfc/data/providers/BuiltinItemTags.java
```diff
@@ -101,65 +104,953 @@ protected void addTags(HolderLookup.Provider provider)
...
         for (Metal metal : Metal.values())
         {
             metalTag(metal, Metal.ItemType.INGOT, Tags.Items.INGOTS);
             if (metal.defaultParts())
             {
                 metalTag(metal, Metal.ItemType.DOUBLE_INGOT, DOUBLE_INGOTS);
-                metalTag(metal, Metal.ItemType.SHEET, SHEETS);
-                metalTag(metal, Metal.ItemType.DOUBLE_SHEET, DOUBLE_SHEETS);
+                metalTag(metal, Metal.ItemType.SHEET, PLATES); // changed to plates to match other mods usage
+                metalTag(metal, Metal.ItemType.DOUBLE_SHEET, DOUBLE_PLATES);  //Changed to Double plates to match other mods usage
```
eerussianguy: TFC specifically does not tag sheets as plates, this is a known thing and I don't think we should be changing it, its not a bug.

eerussianguy (changes requested):

eerussianguy:
The github app isn't behaving for some reason, so I'm leaving comments this way

There are piles of repeated lines and tags for crops and other types. What happens if there's a spelling error? What happens when there's crops added or removed? This is a recipe for a maintenance disaster. There has to be a more repeatable and programmatic way to do this rather than listing out dozens of strings. Datagen is for hard references, not just rewriting what was in python. Let's use programming here.

```java
tag(commonTagOf(Registries.ITEM, "dusts/kaolinite"))
            .add(TFCItems.POWDERS.get(Powder.KAOLINITE).key());
```

^ This can be a loop. Nobody wants to go and check for usages in 10 places when they add to an enum.

Some of these seem like they are of tenuous usefulness - like 'raw materials' - is this a tag that is used? especially with all these sub tags of it? wouldn't a modder use c:gems/sapphire or c:sapphire before they did c:raw_materials/sapphire ? Who is the use case here? I feel like some of this was done for completeness rather than usefulness. Remember, packmakers are also free to add tags.

I appreciate the work that went into this, but lets clean this up a little to get it ready to go.

TKFT:
I agree, I can clean it up some and see where loops can be used. Unfortunately some can not be fixed and will need to be listed out but I can use more enums for those. 

As for the usefulness of tags, I'll definitely admit I over tagged with the expectation that they could be approved or not. Going through it all at once is easier then going back and forth so any that are not needed can be deleted. 

When it comes to the raw_materials tag how it is used or at least how it is supposed to be used is for items that are precursors for the final item, in the case of gems raw_materials/<GEM> is not for the final cut gem but the gem item. 

I am also working on blocks and should have those done soon.

TKFT:
Made changes, still some manually listed sections but there is a lot less now. I am pretty close to finishing the block tags if you want to wait for that I can resubmit as one PR

PR #3581, opened 2026-05-17
Made Tables in Field Guide Translatable & Updated ru_ru

KardyN:


eerussianguy:
@KardyN  why does this have changes to the english translation?
image

KardyN:
I changed them in a few places to match with crop table entries, iirc. Could be done other way around, if you say so

PR #3586, opened 2026-05-22
1.21.1: More Than Just Shelf Rotate/Mirror

Allmoz:
ShelfBlock is missing those methods
Had a look into all rotate/mirror implementations in the codebase

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blocks/devices/AnvilBlock.java
+++ b/src/main/java/net/dries007/tfc/common/blocks/devices/AnvilBlock.java
@@ -159,15 +159,14 @@ public class AnvilBlock extends DeviceBlock implements Tiered
     }
 
     @Override
-    @SuppressWarnings("deprecation")
-    protected BlockState rotate(BlockState state, Rotation rot)
+    public BlockState rotate(BlockState state, Rotation rot)
     {
         return state.setValue(FACING, rot.rotate(state.getValue(FACING)));
     }
 
     @Override
     @SuppressWarnings("deprecation")
-    protected BlockState mirror(BlockState state, Mirror mirror)
+    public BlockState mirror(BlockState state, Mirror mirror)
     {
         return state.rotate(mirror.getRotation(state.getValue(FACING)));
     }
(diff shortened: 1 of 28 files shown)


src/main/java/net/dries007/tfc/common/blocks/devices/AnvilBlock.java
```diff
@@ -159,15 +160,14 @@ protected void createBlockStateDefinition(StateDefinition.Builder<Block, BlockSt
     }
 
     @Override
-    @SuppressWarnings("deprecation")
-    protected BlockState rotate(BlockState state, Rotation rot)
+    public BlockState rotate(BlockState state, Rotation rot)
```
danielpetkau: Why make all these methods public? The super implementation is protected

Allmoz: Is true that i didn't had any particular trouble with those (had not yet gotten to that part of the testing),  seemed wierd to me that only those were protected, since I know those method are called form outside the class and package by other mods. I was tempted to revise all mirror to use the proper (state, mirror) method and get rid of all the deprecation suppressions, but for this PR I just tried to make it consistent with itself

Allmoz: i'm probably in the wrong, should i make them all protected? and work the mirror ones too?

danielpetkau: Yeah I would make both rotate and mirror protected

Allmoz: finished,  im testing it, but mirror in super implementation is public not sure why, still want it as protected?

danielpetkau: Mirror is protected in BlockBehavior in 1.21, maybe you're looking at some other TFC block in an inheritance chain?

Allmoz: you are absolutely right... its mirror(Mirror mirror) the one is public, my eyes are the issue

Allmoz: done >.< thankiuu for keeping me in check

Allmoz:
rebased because of the merge conflict

PR #3590, opened 2026-05-24
Vivecraft compatibility

UrBoiTom:
Fixes the firestarter being unusable in VR.


eerussianguy:
Please apply TFC's code style. Also is there any reason this can't be a `ModList#get` call to check if Vivecraft is installed?

UrBoiTom:
> Also is there any reason this can't be a `ModList#get` call to check if Vivecraft is installed?

That would also work but would disable the animation with vivecraft installed when not in VR

PR #3592, opened 2026-05-24
Tree visuals updated

Therighthon:
Use different textures for leaves in autumn/winter/spring instead of just tinting


Therighthon:
Marking this ready for review, would be nice if someone took a look at the mixin to see if it seems reasonable to them.

The last thing I want to do is make sure that I am happy with all of the new textures and colormaps. Right now, I think that conifers are a little bit too brightly colored in some cases, for instance, and some of the flowering leaves might be a little too vibrant as well.

PR #3613, opened 2026-06-04
Fix isEyeInWaterLikeFluid (1.20)

Pyritie:
SALT_WATER is referenced twice. I assume one is supposed to be SPRING_WATER? This affects things like the underwater ambient sound effects


Allmoz:
Maybe 🤔 ANY_FRESH_WATER

PR #3617, opened 2026-06-10
Horse Parthenogenesis

Allmoz:
Give the genetics of the mother to baby horses and donkeys born from an spawn egg (used on a female specimen), and who i'm to say that females mules can't do parthenogenesis too. No more 53 health babies


Allmoz:
added the fix for baby chested horses

PR #3624, opened 2026-06-16
No Vertical Shelves

Allmoz:
fix crash when placing shelves on invalid surface


danielpetkau:
Turns out there was already a helper to do this properly, I'm using that instead

PR #3628, opened 2026-06-25
Fish Sounds Adjustments 

Allmoz:
Changes fishes, jellyfish and manatee ambiental sounds from splashing to swimming and restricted them to happen only in water


BlackAures1:
That would be very helpful. Fish can be quite noisy when you live near the water.

PR #3631, opened 2026-06-27
Minor fixes

Ankmaniac:
- Fix Clock modes and redstone output
- Fix Powderbowl texture rendering being slightly too big
- Added Powderkeg block entity renderer displaying the fullness level, hitbox gets updated too

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
+++ b/src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
@@ -174,35 +174,6 @@ public class PowderkegBlock extends SealableDeviceBlock
     @Override
     public VoxelShape getShape(BlockState state, BlockGetter level, BlockPos pos, CollisionContext context)
     {
-        if (!state.getValue(SEALED))
-        {
-            @Nullable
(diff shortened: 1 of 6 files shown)


src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
```diff
@@ -171,6 +174,35 @@ protected void neighborChanged(BlockState state, Level level, BlockPos pos, Bloc
...
+                    final double y = Mth.clampedMap(count, 0, maxCount, 2D, 14D);
+                    final VoxelShape gunpowder = box(2D, 2D, 2D, 14D, y, 14D);
+
+                    return Shapes.or(SHAPE_UNSEALED, gunpowder);
+                }
+            }
+
+        }
         return state.getValue(SEALED) ? SHAPE : SHAPE_UNSEALED;
     }
```
eerussianguy, line 175: No, this is just a needless performance sink, there is no reason to have this logic and it adds nothing to the game.

eerussianguy (changes requested):

src/main/java/net/dries007/tfc/common/blocks/devices/PowderkegBlock.java
Ankmaniac: removed the hitbox

PR #3632, opened 2026-06-27
Fresh water in ocean fixes

TKFT:
Fixed fresh water generating around coasts and into oceans.

Original 

Fixed

Note: The lava is salt water, texture swapped  for testing


Therighthon:
Nice, this seems to work great in canyons!

PR #3639, opened 2026-07-02
Enclosed ocean fix

TKFT:
Changes how far inland oceanic types of mountains will generate


TKFT:
There is another version that is less extreme of a change

Therighthon:
Not planned, oceanic/coastal mountains are intentionally shorter than other mountains, not just a single layer shore biome

PR #3652, opened 2026-07-06
New Feature: TFC Camels

strangelyng:
Adds camels with basic TFC mechanics (familiarity, gender, TFC breeding mechanics, etc.)

A few main issues to work out:
- [x] What counts as camel food? (currently using horse food tag)
- [x] Fauna definitions (currently rainfall < 100; temperature > -5C)
- [x] Breeding numbers (currently copied from TFCHorse numbers)
- [x] Loot table balancing
- [x] Merge Bactrian Camels


src/main/java/net/dries007/tfc/common/entities/livestock/camel/TFCCamel.java
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    @Override
+    protected @Nullable SoundEvent getEatingSound()
+    {
+        return super.getEatingSound();
+    }
+
+    @Override
+    protected float getBlockSpeedFactor()
```
Therighthon: For camels, do we want to have a modest speed boost when they are on sand? I know that's been discussed before, and I am not sure if there was a consensus.

strangelyng:
I left a comment in the development channel about some breeding behavior quirks, but overall everything seems to be working. I would still appreciate someone else testing things out to see if I overlooked anything, but I think I was fairly thorough.

src/main/java/net/dries007/tfc/client/ClientEventHandler.java
```diff
@@ -213,30 +214,7 @@
...
-import net.dries007.tfc.client.render.entity.TFCChestedHorseRenderer;
-import net.dries007.tfc.client.render.entity.TFCCodRenderer;
-import net.dries007.tfc.client.render.entity.TFCFishingHookRenderer;
-import net.dries007.tfc.client.render.entity.TFCHorseRenderer;
-import net.dries007.tfc.client.render.entity.TFCPufferfishRenderer;
-import net.dries007.tfc.client.render.entity.TFCSalmonRenderer;
-import net.dries007.tfc.client.render.entity.TFCSquidRenderer;
-import net.dries007.tfc.client.render.entity.TFCTropicalFishRenderer;
-import net.dries007.tfc.client.render.entity.ThrownJavelinRenderer;
+import net.dries007.tfc.client.render.entity.*;
```
Therighthon: Code style thing, we avoid * imports

Therighthon (commented):
Had some time on a bus to look at this more closely. Looks good if you just revert that * import. If you want to add Bactrian camels after this, Dodo made an entity model you can use.

Therighthon:
And, as I think about it, should they have an immunity to cactus damage to make them better for desert riding?

src/main/java/net/dries007/tfc/common/entities/livestock/camel/TFCCamel.java
strangelyng: That seems reasonable to me, something in the range of 1.2-1.3 doesn't feel super dramatic just walking between sand and non-sand, but you can definitely feel it when sprinting.

I could go in the middle and tweak if needed?

strangelyng:
I can add cactus damage immunity too. It would be good to add additional incentives to use them.

strangelyng:
Actually, should the player be made immune to cactus damage too while riding a camel? I should be able to do that with an event.

Therighthon:
> Actually, should the player be made immune to cactus damage too while riding a camel? I should be able to do that with an event.

I don't think so, my thought with the damage immunity is mostly so that you don't have to worry about riding over shorter cactuses. But if you ride under an arm of a saguaro while riding a camel, which is I think the only way you could conceivably touch a cactus from the back of a camel, you deserve to take damage.

strangelyng:
> I don't think so, my thought with the damage immunity is mostly so that you don't have to worry about riding over shorter cactuses. But if you ride under an arm of a saguaro while riding a camel, which is I think the only way you could conceivably touch a cactus from the back of a camel, you deserve to take damage.

I do agree with the idea, but it seems that the player hitbox actually doesn't even shrink while riding mounts? So you'll still take damage riding through prickly pears and can take damage from the barrel cactuses if you hit them in the wrong way.

PR #3653, opened 2026-07-06
Make prey animals avoid predators

strangelyng:
The `hunts_land_prey` entity tag is not currently assigned to any entities, meaning that prey animals have no issue walking up to predators until they aggro and get attacked.

Changing these usages of HUNTS_LAND_PREY to LAND_PREDATORS makes prey animals avoid predators.


Therighthon:
Good catch. We should remove the HUNTS_LAND_PREY tag entirely if we are removing all uses of it. I'd also like to check if something like this is why animals are not avoiding ramming animals

PR #3654, opened 2026-07-06
Block and Item Common Tags

TKFT:
Completed block and item common tags.


TKFT:
fixed the failing test, added new items and blocks and made safe to merge

TKFT:
added new tools

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+
+        tag(c.apply("anvils/stone"))
+            .add(TFCBlocks.ROCK_ANVILS);
+
+        tag(anvilsTag)
+            .addTag(c.apply("anvils/stone"));
+
+        //Barrels
+
+        tag(Tags.Blocks.BARRELS_WOODEN).add(TFCBlocks.WOODS, Wood.BlockType.BARREL);
```
Therighthon, line 367: Shouldn't we not use this tag since our barrels are pretty different from what the vanilla idea of a barrel is now?

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+                    .add(blocks.get(Metal.BlockType.BARS));
+
+                tag(barsTag)
+                    .addTag(metalBarsTag);
+            }
+        });
+
+        //Bookshelves
+
+        tag(c.apply("bookshelves/wooden"))
```
Therighthon, line 388: Our bookshelves are more like the fancy storage bookshelves of vanilla, is that a separate tag?

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+            .add(TFCBlocks.ROCK_BLOCKS, Rock.BlockType.MOSSY_BRICKS)
+            .add(pivot(TFCBlocks.ROCK_DECORATIONS, Rock.BlockType.MOSSY_BRICKS)
+                .values().stream().map(DecorationBlockHolder::slab))
+            .add(pivot(TFCBlocks.ROCK_DECORATIONS, Rock.BlockType.MOSSY_BRICKS)
+                .values().stream().map(DecorationBlockHolder::stair))
+            .add(pivot(TFCBlocks.ROCK_DECORATIONS, Rock.BlockType.MOSSY_BRICKS)
+                .values().stream().map(DecorationBlockHolder::wall));
+
+        tag(c.apply("bricks/mud"))
+            .add(TFCBlocks.SMOOTH_MUD_BRICKS);
```
Therighthon, line 453: add to line 402?

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+            TFCBlocks.KAOLIN_CLAY_GRASS
+        );
+        tag(c.apply("clays/fire")).add(TFCBlocks.FIRE_CLAY_BLOCK);
+        tag(c.apply("clays/hardened")).add(TFCBlocks.HARDENED_CLAY);
+
+        tag(c.apply("clays")).addTags(
+            c.apply("clays/normal"),
+            c.apply("clays/kaolin"),
+            c.apply("clays/fire"),
+            c.apply("clays/hardened")
```
Therighthon, line 552: So, the main benefit of us using common tags is more out of the box mod compatibility for recipes. I don't think another mod's recipe that accepts clay blocks would expect hardened clay blocks.

Also, I doubt a lot of mods are using tags for clay blocks anyways, this seems unnecessary to me in general.

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+
+        //Cobblestones
+
+        final TagKey<Block> cobblestonesTag = c.apply("cobblestones");
+
+        TFCBlocks.ROCK_BLOCKS.forEach((rock, blocks) -> {
+            final TagKey<Block> rockCobblestonesTag = c.apply("cobblestones/" + rock.getSerializedName());
+            final Map<Rock.BlockType, DecorationBlockHolder> decorations = TFCBlocks.ROCK_DECORATIONS.get(rock);
+
+            tag(rockCobblestonesTag).add(
```
Therighthon, line 563: Is it consistent for stairs/slabs of blocks to use the parent tab? I am not familiar with common tag conventions, but consider a mod recipe: if slabs count the same as full blocks, that's half price.

Keep in mind that even though these are block tags, some mods use block tags in recipes, for instance ArborFirmaCraft and Roads and Roofs. Ofc both of those are built to be compatible with AFC anyways, but other mods do similar things

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+
+        tag(c.apply("corals/block")).add(
+            Blocks.BRAIN_CORAL_BLOCK, Blocks.DEAD_BRAIN_CORAL_BLOCK,
+            Blocks.BUBBLE_CORAL_BLOCK, Blocks.DEAD_BUBBLE_CORAL_BLOCK,
+            Blocks.FIRE_CORAL_BLOCK, Blocks.DEAD_FIRE_CORAL_BLOCK,
+            Blocks.HORN_CORAL_BLOCK, Blocks.DEAD_HORN_CORAL_BLOCK,
+            Blocks.TUBE_CORAL_BLOCK, Blocks.DEAD_TUBE_CORAL_BLOCK
+        );
+
+        tag(c.apply("corals/living")).add(
```
Therighthon, line 608: Are there vanilla coral tags?

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+            c.apply("corals/plant"),
+            c.apply("corals/fan"),
+            c.apply("corals/block"),
+            c.apply("corals/living"),
+            c.apply("corals/dead")
+        );
+
+        //Crates
+
+        tag(c.apply("crates/wooden"))
```
Therighthon, line 638: Is this changing an existing tfc:crate tag to be a common tag? I don't really see a benefit of that. And it will break addons to do things like that.

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+            .addTag(c.apply("crops/" + Crop.BEET.getSerializedName()));
 
-        // Ore Tags
+        tag(c.apply("crops/pumpkin"))
+            .add(TFCBlocks.PUMPKIN);
+
+        tag(c.apply("crops/melon"))
+            .add(TFCBlocks.MELON);
+
+        tag(cropsTag).addTags(
```
Therighthon, line 674: Seems like we are missing some crops here...

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+            TFCBlocks.PLANTS.get(Plant.PURPLE_WATER_LILY), TFCBlocks.PLANTS.get(Plant.QANTU), TFCBlocks.PLANTS.get(Plant.RAMIREZELLA),
+            TFCBlocks.PLANTS.get(Plant.RAMUNDA), TFCBlocks.PLANTS.get(Plant.SACRED_DATURA), TFCBlocks.PLANTS.get(Plant.SILVER_SPURFLOWER),
+            TFCBlocks.PLANTS.get(Plant.SNAPDRAGON_PINK), TFCBlocks.PLANTS.get(Plant.SNAPDRAGON_RED), TFCBlocks.PLANTS.get(Plant.SNAPDRAGON_WHITE),
+            TFCBlocks.PLANTS.get(Plant.SNAPDRAGON_YELLOW), TFCBlocks.PLANTS.get(Plant.TANK_BROMELIAD), TFCBlocks.PLANTS.get(Plant.TRILLIUM),
+            TFCBlocks.PLANTS.get(Plant.TROPICAL_MILKWEED), TFCBlocks.PLANTS.get(Plant.TULIP_ORANGE), TFCBlocks.PLANTS.get(Plant.TULIP_PINK),
+            TFCBlocks.PLANTS.get(Plant.TULIP_RED), TFCBlocks.PLANTS.get(Plant.TULIP_WHITE), TFCBlocks.PLANTS.get(Plant.WHITE_WATER_LILY),
+            TFCBlocks.PLANTS.get(Plant.YELLOW_SAXIFRAGE), TFCBlocks.PLANTS.get(Plant.YELLOW_WATER_LILY), TFCBlocks.PLANTS.get(Plant.YUCCA)
+        );
+
+        tag(c.apply("flowers")).addTags(
```
Therighthon, line 817: Am I reading these wrong when I assume that new stuff you've added is going to the common namespace, and not to vanilla tags? 

Because we already add flowers to the vanilla flowers tag, I don't think we do anything for tall flowers tag

But my point is, if a vanilla tag exists for something, we should be using that, not a forge tag

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+
+        //Icicle
+
+        tag(c.apply("icicle"))
+            .add(TFCBlocks.ICICLE);
+
+        //Item Piles
+        //c:item_piles, c:item_piles/TYPE
+
+        tag(c.apply("item_piles/double_ingot"))
```
Therighthon, line 907: Why would we want tags for something that A. no other mod adds and B. we don't use a tag for internally C. is only a single block

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+
+                tag(lampsTag)
+                    .addTag(metalLampTag);
+            }
+        });
+
+        //Leaves
+        //c:leaves
+
+        tag(c.apply("leaves"))
```
Therighthon, line 1103: We already put leaves in the vanilla leaves tags, which is what every reasonable mod uses...

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+
+        tag(c.apply("pipes"))
+            .addTag(c.apply("pipes/fluid"));
+
+        //Planks
+        //c:planks, c:planks/WOOD
+
+        final TagKey<Block> planksTag = c.apply("planks");
+
+        TFCBlocks.WOODS.forEach((wood, blocks) -> {
```
Therighthon, line 1140: Surely there are vanilla tags, and I think we already use them?

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+        tag(c.apply("player_workstations/sewing_table"))
+            .add(TFCBlocks.WOODS, Wood.BlockType.SEWING_TABLE);
+
+        tag(c.apply("player_workstations/stove"))
+            .add(TFCBlocks.STOVE, TFCBlocks.STOVE_POT);
+
+        tag(Tags.Blocks.PLAYER_WORKSTATIONS_CRAFTING_TABLES)
+            .add(TFCBlocks.WOODS, Wood.BlockType.WORKBENCH);
+
+        tag(playerWorkstationsTag).addTags(
```
Therighthon, line 1253: I don't totally understand the point of this, but I'm willing to blame that on my own ignorance

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+
+        tag(c.apply("potted_plants"))
+            .add(TFCBlocks.POTTED_PLANTS)
+            .add(TFCBlocks.FRUIT_TREE_POTTED_SAPLINGS)
+            .add(TFCBlocks.WOODS, Wood.BlockType.POTTED_SAPLING);
+
+        //Pressure Plates
+        //c:pressure_plates, c:pressure_plates/wooden, c:pressure_plates/stone
+
+        tag(c.apply("pressure_plates/wooden"))
```
Therighthon, line 1285: Vanilla

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+            .add(TFCBlocks.FIRE_BRICK_SHELF);
+
+        tag(c.apply("shelves")).addTags(
+            c.apply("shelves/wooden"),
+            c.apply("shelves/brick")
+        );
+
+        //Signs
+
+        final TagKey<Block> signTag = c.apply("signs");
```
Therighthon, line 1521: Vanilla?

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+        );
+
+        //Slabs
 
-        tag(Tags.Blocks.STORAGE_BLOCKS_WHEAT).remove(Blocks.HAY_BLOCK); // We repurpose this as storing straw
-        pivot(TFCBlocks.METALS, Metal.BlockType.BLOCK).forEach((metal, block) -> tag(storageBlockTagOf(Registries.BLOCK, metal)).add(block));
+
+        //Sluices
+
+        tag(c.apply("sluices/wooden"))
```
Therighthon, line 1559: This is a great example of something that should just be a tfc tag

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+        tag(c.apply("sluices"))
+            .addTags(c.apply("sluices/wooden"));
+
+        //Stairs
+
+
+        //Storage Blocks
+
+        tag(Tags.Blocks.STORAGE_BLOCKS_WHEAT)
+            .remove(Blocks.HAY_BLOCK); // We repurpose this as storing straw
```
Therighthon, line 1571: Why do this?

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+
+        tag(Tags.Blocks.STORAGE_BLOCKS_SLIME)
```
Therighthon, line 1573: No, this surely makes mod compat worse not better

src/data/java/net/dries007/tfc/data/providers/BuiltinBlockTags.java
```diff
@@ -329,65 +337,1330 @@ protected void addTags(HolderLookup.Provider provider)
...
+        //Stripped Wood
+        //c:stripped_woods, c:stripped_woods/WOOD
+
+        TFCBlocks.WOODS.forEach((wood, blocks) -> {
+            final TagKey<Block> strippedWoodTag = c.apply("stripped_woods/" + wood.getSerializedName());
+
+            tag(strippedWoodTag)
+                .add(blocks.get(Wood.BlockType.STRIPPED_WOOD));
+
+            tag(Tags.Blocks.STRIPPED_WOODS)
```
Therighthon, line 1604: This seems like something where we could conceivably migrate to just using the vanilla (if it exists, otherwise common) stripped woods tag

Therighthon (commented):
Just doing a first round review of BuiltinBlockTags, and honestly, I am not clear on what your goal is with this. There is some stuff that just seems to run contrary to mod compatibility. A LOT of what I am seeing appears to be adding common tags for things that either only exist in TFC, or are different enough when they are implemented by other mods that it doesn't make sense for them to share a tag, which in my mind is not the point of common tags, that is what precisely what tags in the TFC namespace are for. Also, unless I am missing some code due to my skimming, you seem to be adding things to common tags when there are vanilla tags available.

I appreciate that you've put a substantial amount of effort into this PR, but based on this first file that I've finally had time to skim through, most of it doesn't actually seem productive to me.

PR #3657, opened 2026-07-10
TFC Armadillos

strangelyng:
## Initial Implementation Notes
Armadillos are currently implemented as basic wild animals. They share most of the functionality from the WildAnimal class, and their brain is currently a slightly modified verison of the vanilla ArmadilloAI.

They spawn anywhere with more than 250mm of rainfall and a temperature greater than 12C.

They drop armadillo scutes and raw armadillo, which currently shares its nutritional values with rabbit.

They remain brushable until it's decided whether they should have familiarity and breeding behaviors implemented.
They don't currently implement Temptable, though the AI is set up for it.

This initial implementation uses a custom SensorType that uses the DANGER_DETECTED_RECENTLY MemoryModuleType used by the vanilla Armadillo but modified to detect TFC mobs. This may be made generic if that seems useful, alternatively, I may be able to rewrite their unique behaviors to work with AVOID_TARGET instead like other Prey animals.

Any of these details are subject to change based on feedback, but I wanted to get the PR up and visible.

## Implementation Notes (Final?)
- Armadillos are implemented similarly to Prey animals and members of the WildAnimal class.
- They spawn in temperatures greater than 12C in areas with between 100 and 400mm of rainfall.
- They drop armadillo scutes and raw armadillo, which shares its nutritional values with rabbit.
- Armadillo scute drops are linked to the TFC Calendar, and drop every 30 hours.
- Armadillos can be tempted with spider eyes, but not bred or tamed.
- Includes a custom sensor called ScareSensor, which pairs with the new Scareable interface to handle unique distance-based prey behaviors, such as the armadillo balling up when predators get close


Therighthon:
Thoughts:
- Scute shedding should probably be based on the calendar for consistency
- At some point, it might make sense to rework our turtles to be based on the armadillo behavior instead of our play dead behavior
- I don't really see a need for them to use avoid target, this seems to work well

Therighthon:
image

Looks like they still have an outdated scute timer in Jade

Therighthon:
Otherwise, looks good to me. My initial thought was that they should be a dry biome animal, but as I actually look them up, most armadillos live in pretty wet regions, so that's just my American armadillo biases I guess

PR #3668, opened 2026-07-23
Rock Settings Registry

Notenoughmail:
Allows the rock settings to be [individually modified](https://discord.com/channels/432522930610765835/646085141847998484/1529363568779071498) without needing to rewrite the entire world preset file. Inline definitions are still possible, meaning this should be backwards compatible with any datapacks that were made before this change

There is one minor break, `RockLayerSettings$Layer#rock` now returns a `Holder<RockSettings>` instead of a raw `RockSettings`

Also, while doing this I had to fix the EMI dependency stuff that has been causing other PRs to fail checks. I'm not certain if it's the *proper* fix, but would y'all be interested in having that included?


src/data/java/net/dries007/tfc/data/providers/BuiltinRockSettings.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public BuiltinRockSettings(BootstrapContext<RockSettings> context)
+    {
+        for (Rock rock : Rock.VALUES)
+        {
+            context.register(BuiltinWorldPreset.rockKey(rock), makeSetting(rock));
+        }
+    }
+
+    public static RockSettings makeSetting(Rock rock) {
```
eerussianguy: braces on newlines

src/data/java/net/dries007/tfc/data/providers/BuiltinRockSettings.java
```diff
@@ -0,0 +1,143 @@
...
+        .put(SCHIST, Boolean.FALSE)
+        .put(GNEISS, Boolean.FALSE)
+        .put(MARBLE, Boolean.TRUE)
+        .build();
+
+    // Used by badlands to determine whether they should have black sand
+    private static final Map<Rock, Boolean> ROCK_SET_MAFIC = ImmutableMap.<Rock, Boolean>builder()
+        .put(GRANITE, Boolean.FALSE)
+        .put(DIORITE, Boolean.FALSE)
+        .put(GABBRO, Boolean.TRUE)
```
eerussianguy: would prefer you use autoboxing (`true`) if it allows it.

src/data/java/net/dries007/tfc/data/providers/BuiltinWorldPreset.java
```diff
@@ -246,24 +199,7 @@ private static LayerData layerOf(String layerId, Map<Rock, String> layers)
...
-            TFCBlocks.SANDSTONE.get(color).get(SandstoneBlockType.RAW).get(),
-            Optional.of(blocks.get(BlockType.SPIKE).get()),
-            Optional.of(blocks.get(BlockType.LOOSE).get()),
-            Optional.of(blocks.get(BlockType.MOSSY_LOOSE).get()),
-            Optional.of(karst),
-            Optional.of(mafic)
-        );
+    public static ResourceKey<RockSettings> rockKey(Rock rock) {
+        return ResourceKey.create(RockSettings.KEY, Helpers.identifier(rock.getSerializedName()));
     }
```
eerussianguy: This is a public method but it forces TFC's identifier. Is that what we want here?

eerussianguy (changes requested):
Dont see anything wrong with this, but my datapack registries knowledge is limited. Lets clean up the code style a bit.

I assume this breaks existing worlds?

Would you also mind updating the API docs repo?

src/data/java/net/dries007/tfc/data/providers/BuiltinWorldPreset.java
Notenoughmail: Seeing as the method is in the data source set and only accepts TFC's rock types, I don't believe anyone would actually be able to use it without copying it anyway... at least until that gets properly published somewhere

Notenoughmail:
> I assume this breaks existing worlds?

No, `RegistryFileCodec.create` allows inline definitions unless `false` is explicitly passed as the 3rd parameter, so existing worlds/datapacks with inline definitions load and work without issue

> Would you also mind updating the API docs repo?

Certainly!

PR #3689, opened 2026-08-18
Make wild crop maturity months configurable

rjdunlap:
## What changed

- Adds `wildCropMaturityStartMonth` and `wildCropMaturityEndMonth` to the synced per-world server config.
- Uses those values when determining whether wild crops are mature.
- Keeps `JUNE` through `OCTOBER` as the defaults, preserving existing behavior.
- Supports maturity periods that wrap across the end of the year, such as October through March.

## Why

Wild crop maturity is currently fixed to June–October in `WildCropBlock`. Exposing the boundaries lets modpack authors and server operators tune seasonal food availability without patching TFC.

## Impact

Existing worlds and packs retain the current June–October behavior unless they opt into different values. The settings are stored in the per-world server config and synchronized to clients through the existing config system.

## Validation

- `./gradlew build` — passed
- `./gradlew updateLicenses` — passed
- `git diff --check` — passed


eerussianguy:
Let this slide by accident, but this needed to target 1.21 as well.

PR #3695, opened 2026-08-25
Give datagen welding recipes a minimum tier

strangelyng:
Applies Math.max() to the tier of datagen'd welding recipes to prevent recipes with a tier of -1, which caused a broken tooltip:

BEFORE

AFTER

### DIFF
--- a/src/data/java/net/dries007/tfc/data/recipes/WeldingRecipes.java
+++ b/src/data/java/net/dries007/tfc/data/recipes/WeldingRecipes.java
@@ -50,7 +50,7 @@ public interface WeldingRecipes extends Recipes
         add(new WeldingRecipe(
             ingredientOf(ingot1, ItemType.INGOT),
             ingredientOf(ingot2, ItemType.INGOT),
-            ingotOut.tier() - 1,
+            Math.max(0, ingotOut.tier() - 1),
             ItemStackProvider.of(TFCItems.METAL_ITEMS.get(ingotOut).get(ItemType.INGOT)),
             IGNORE
         ));
@@ -61,7 +61,7 @@ public interface WeldingRecipes extends Recipes
         add(new WeldingRecipe(
             ingredientOf(metal, input1),
             ingredientOf(metal, input2),
-            metal.tier() - 1,
+            Math.max(0, metal.tier() - 1),
             ItemStackProvider.of(TFCItems.METAL_ITEMS.get(metal).get(output)),
             behavior
         ));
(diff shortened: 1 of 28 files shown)


Notenoughmail:
imo the tooltip should just not be preset when the tier is negative (or ≤ zero?) since welding recipe `tier`s are [optional and ~~default to -1~~](https://terrafirmacraft.github.io/Documentation/1.21.x/recipes/#anvil-welding) *default to [0](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.21.x/src/main/java/net/dries007/tfc/common/recipes/WeldingRecipe.java#L36)*

strangelyng:
> imo the tooltip should just not be preset when the tier is negative (or ≤ zero?) since welding recipe `tier`s are [optional and ~default to -1~](https://terrafirmacraft.github.io/Documentation/1.21.x/recipes/#anvil-welding) _default to [0](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.21.x/src/main/java/net/dries007/tfc/common/recipes/WeldingRecipe.java#L36)_

I can do that too, if that's desired. I didn't notice that docs mismatch, I just went off of what the recipe classes showed.

EDIT: Looks like the change happened here: 5934ab2 with the addition of EMI support

Therighthon:
> imo the tooltip should just not be preset when the tier is negative (or ≤ zero?) since welding recipe `tier`s are [optional and ~default to -1~](https://terrafirmacraft.github.io/Documentation/1.21.x/recipes/#anvil-welding) _default to [0](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.21.x/src/main/java/net/dries007/tfc/common/recipes/WeldingRecipe.java#L36)_

I agree, would be better to just not show a tooltip for tier<=0

strangelyng:
> > imo the tooltip should just not be preset when the tier is negative (or ≤ zero?) since welding recipe `tier`s are [optional and ~default to -1~](https://terrafirmacraft.github.io/Documentation/1.21.x/recipes/#anvil-welding) _default to [0](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.21.x/src/main/java/net/dries007/tfc/common/recipes/WeldingRecipe.java#L36)_
> 
> I agree, would be better to just not show a tooltip for tier<=0

I suppose I should clarify, do you want this to apply to Anvil recipes as well? That's what I did, because it seems inconsistent otherwise, but I should confirm. Also, do you want the change to the datagen recipes reverted, then?

PR #3701, opened 2026-08-29
fix: prevent rotten eggs from starting incubation

### LINKED ISSUE #3699: Rotten eggs can still hatch (1.20)
1. Obtain a fertilised egg as normal
2. Put the egg in a chest
3. Wait until the egg rots and says it's ready to hatch
4. Put the egg back in a nest box
5. The egg hatches

If you wait even longer, the egg can hatch into an adult.

mvanhorn:
Add per-slot incubation eligibility state to `NestBoxBlockEntity`, setting a slot eligible when a fertilized, non-rotten egg first enters it and clearing or recomputing that state whenever the slot is removed or replaced. A fertilized egg can age in an ordinary chest until its food component is rotten, then be moved into a nest box and still hatch because `NestBoxBlockEntity.serverTick` checks only `EggComponent.canHatch()`.

Insert a fresh fertilized egg into an empty nest slot, serialize and reload the nest, advance beyond its hatch and rotten dates, and verify the slot remains eligible for the existing catch-up hatch path; Mark a fertilized egg rotten before inserting it and verify the slot is ineligible and the hatch path does not consume the egg or spawn its saved entity after the hatch day.

Fixes #3699

### DIFF
--- a/src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
+++ b/src/main/java/net/dries007/tfc/common/blockentities/NestBoxBlockEntity.java
@@ -8,6 +8,8 @@ package net.dries007.tfc.common.blockentities;
 
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
+import net.minecraft.core.HolderLookup;
+import net.minecraft.nbt.CompoundTag;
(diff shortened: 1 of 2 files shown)


eerussianguy:
Please punish your agent, this was actually a one line fix and they were targeting the wrong branch.

Issue #3706, opened 2026-09-05
[BUG] [1.20.1] [1.21.1] Hoe tooltips are not effected by config.


Glitchcode2447:
Using `climateTooltipStyle = "FAHRENHEIT"` changes the temperature tab of the Survival UI to Fahrenheit, but not the Hoe Tootips.

As [mentioned](https://github.com/TerraFirmaCraft/TerraFirmaCraft/issues/3693#issuecomment-5421246758) by [Notenoughmail](https://github.com/Notenoughmail) the code for this seems to have [completely ignored](https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.20.x/src%2Fmain%2Fjava%2Fnet%2Fdries007%2Ftfc%2Fcommon%2Fblocks%2Fsoil%2FFarmlandBlock.java#L80-L82) the config and just used the default; Celsius.

This was tested on the latest of both releases before posting.

---

This issue was recreated from issue #3693 due to having only one of two changes being made, before it was closed.

PR #3709, opened 2026-09-06
Add style to the hoe tooltip (1.20)

ospis12:
This PR adds temperature style to the FarmlandBlock  hoe tooltip.
Fixes: #3706


Therighthon:
Haven't checked this one out, but before anyone approves confirm that there isn't a crash for fruit trees like in the 1.21 version of this PR

